PR #2, opened 2018-01-23
Adding sha1 to generation

### LINKED ISSUE #1: Add sha1s to gmaven.bzl
This would improve security. Also I hypothesize that this will make things work better with `--experimental_repository_cache` flag. (not 100% sure how that works, but appears to work with other types of builds once a hash is added)

mikelikespie:
Fixes #1


kamalmarhubi:
As I just mentioned in https://github.com/bazelbuild/gmaven_rules/issues/1#issuecomment-458758950, I don't think sha1 will help with the repo cache. These rules `java_import_external` and `aar_import_external` in `import_external.bzl`, which doesn't have a sha1 attribute, and can't add one as [`repository_ctx.download()`](https://docs.bazel.build/versions/master/skylark/lib/repository_ctx.html#download) doesn't support it.

jin:
we're deprecating the old gmaven_rules, please take a look at the new repository rule. Thanks for the PR!

PR #3, opened 2018-01-24
Override maven settings file

mikelikespie:
This is a change that lets one override the settings file. This is useful for us since we have a corp mirror of the repos.


mikelikespie:
Thanks all for looking at this. I don't work at Square anymore, nor did wee end up using this anyways (we ended up writing our own script + skylark files). Going to close this issue. Thanks!

PR #5, opened 2018-03-27
Add a macro to convert FQNs to generated target names

jin:
This adds `gmaven_artifact`, a convenience macro for users to directly specify the artifact's FQN. This saves users from grepping `gmaven.bzl` for target names, and mentally replace dots with underscores.

### DIFF
--- a/defs.bzl
+++ b/defs.bzl
@@ -1,0 +1,23 @@
+def gmaven_artifact(fqn):
+  parts = fqn.split(":")
+  packaging = "aar"
+
+  if len(parts) < 2 or len(parts) > 5:
+    fail("Invalid qualified name for artifact: %s" % fqn)
+  elif len(parts) == 3:
+    group_id, artifact_id, version = parts
+  elif len(parts) == 4:
+    group_id, artifact_id, version, packaging = parts
+  elif len(parts) == 5:
+    _, _, _, _, classifier = parts
+    fail("Classifiers are currently not supported. Please remove it from the coordinate: %s" % classifier)
+
+  return "@%s_%s_%s//%s" % (
+        escape(group_id),
+        escape(artifact_id),
+        escape(version),
+        packaging
+      )
+
+def escape(string):
+  return string.replace(".", "_")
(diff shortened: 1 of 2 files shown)


defs.bzl
```diff
@@ -0,0 +1,23 @@
+def gmaven_artifact(fqn):
+  parts = fqn.split(":")
+  packaging = "aar"
```
aj-michael: The default should probably be jar. Every other tool that parses maven coordinates seems to think so.

defs.bzl
```diff
@@ -0,0 +1,23 @@
...
+
+  if len(parts) == 3:
+    group_id, artifact_id, version = parts
+  elif len(parts) == 4:
+    group_id, artifact_id, version, packaging = parts
```
aj-michael: I believe the orders for length 4 and 5 are
groupId:artifactId:packaging:version
groupId:artifactId:packaging:classifier:version
https://maven.apache.org/pom.html#Maven_Coordinates

defs.bzl
aj-michael: I just realized why you did this: Because most of them are aar's and a user doesn't necessarily know that.

Perhaps a good solution here would be: In addition to generating the maven_aar/maven_jar (which in turn generate a package named "jar" or "aar"), also generate an alias rule somewhere* with a name that does not depend on "aar" or "jar", so then the BUILD file writer does not need to know which it is.

* I'm not sure where that somewhere would be. Maybe the maven_jar/maven_aar rule could create the alias in it's repository? Otherwise we'd need to create a new repository with these aliases, which is ugly because it adds yet another layer of abstraction. We could use bind, but that I think that might be deprecated.

defs.bzl
jin: I remembered following the order in that page, but it was switched out about a year ago: https://github.com/bazelbuild/bazel/commit/7d5b31f9cb33dd86542cd2cd0fea63954010fc00

jin: Done. Switched to follow `groupId:artifactId:packaging:version` and `groupId:artifactId:packaging:classifier:version`

defs.bzl
jin: I've switched it to use jar by default. I like the alias idea, but I think that's out of scope of this PR.

Issue #6, opened 2018-04-10
Initialize transfer to github.com/bazelbuild


jin:
- [x] Add LICENSE file (Apache 2.0)
- [x] Add license headers to files
- [x] Add CONTRIBUTING file, noting CLA policy
- [x] Add AUTHORS file
- [x] Update README with support policy

jin:
https://github.com/bazelbuild/gmaven_rules

PR #11, opened 2018-04-10
Add support policy and update README

jin:
#6

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,14 @@
-A bzl file that contains maven_jar and maven_aar rules for all artifacts in
-https://maven.google.com. Not guaranteed to be correct or up to date. Some of
-the artifacts depend on artifacts that are not present on
-https://maven.google.com. We ignore these and hope not to fail.
+# gmaven_rules
+
+This project generates `gmaven.bzl`, a file containing `maven_jar` and `maven_aar`
+rules for all artifacts in [https://maven.google.com](https://maven.google.com).
+
+# Support Policy
+
+This project is an **interim solution** during which Google Maven and AAR
+support is added to [bazel-deps](https://github.com/johnynek/bazel-deps).
+
+# Usage instructions
 
 To use this from your project, in your `WORKSPACE` file add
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,7 +1,14 @@
...
-the artifacts depend on artifacts that are not present on
-https://maven.google.com. We ignore these and hope not to fail.
+# gmaven_rules
+
+This project generates `gmaven.bzl`, a file containing `maven_jar` and `maven_aar`
+rules for all artifacts in [https://maven.google.com](https://maven.google.com).
+
+# Support Policy
+
+This project is an **interim solution** during which Google Maven and AAR
```
aj-michael: I think the phrasing of "solution during which" doesn't make sense. Suggestion:

This project is an interim solution that is supported until Google Maven and AAR support is added to bazel-deps

README.md
```diff
@@ -29,10 +36,19 @@ android_library(
 )
 ```
 
-You can see the full list of generated targets in [gmaven.bzl](https://raw.githubusercontent.com/aj-michael/gmaven_rules/master/gmaven.bzl).
+You can see the full list of generated targets in [`gmaven.bzl`](https://raw.githubusercontent.com/aj-michael/gmaven_rules/master/gmaven.bzl).
 
-To regenerate gmaven.bzl, run the following command. It will take about 5 minutes.
+# Regenerating targets
```
aj-michael: nit: Perhaps change the title to "Regenerating gmaven.bzl" or "Updating gmaven.bzl"? This doesn't actually generate targets, it generates repository rules

README.md
```diff
@@ -29,10 +36,19 @@ android_library(
...
 
 ```
 rm gmaven.bzl && javac GMavenToBazel.java && java GMavenToBazel
 ```
+
+
+# Known issues
+
+Some of the artifacts depend on other artifacts that are not present on Google
+Maven. These targets do not work as the cross-repository resolution is not
```
aj-michael: Perhaps mention something like "These dependencies are silently ignored and may cause failures at runtime."

README.md
```diff
@@ -1,7 +1,14 @@
...
+
+This project generates `gmaven.bzl`, a file containing `maven_jar` and `maven_aar`
+rules for all artifacts in [https://maven.google.com](https://maven.google.com).
+
+# Support Policy
+
+This project is an **interim solution** during which Google Maven and AAR
+support is added to [bazel-deps](https://github.com/johnynek/bazel-deps).
+
+# Usage instructions
```
aj-michael: While you're here, could you add a sentence describing what to use for the <FILL IN A COMMIT HERE>? Suggestion, something like:

"We suggest you use the most recent commit from this repository."

jin: I was thinking of creating releases after migrating to bazelbuild. We can keep track of versions that way, instead of relying on unreadable hashes.

README.md
jin: done

README.md
jin: done

README.md
jin: done

README.md
aj-michael: SGTM

Issue #36, opened 2019-03-11
Failed build on Buildkite Windows pipeline


jin:
```
Error while trying to read dep-tree.json: java.io.IOException: ERROR: src/main/native/windows/processes-jni.cc(454): CreateProcessW("C:\msys64\usr\bin\cat" D:/b/wpicawde/external/maven/dep-tree.json): The system cannot find the file specified.
--
  | and referenced by '//ui/espresso/IntentsBasicSample:IntentsBasicSampleTestLib'
```

We'll need to replace https://github.com/bazelbuild/rules_jvm_external/blob/3af48fbb018beb8eb3f08ea8d838054510465d57/coursier.bzl#L247 with another approach. 

Related: https://github.com/bazelbuild/bazel/issues/3766 and https://github.com/bazelbuild/bazel/pull/7309

PR #37, opened 2019-03-11
Fix cat on Windows

jin:
We shouldn't hardcode the path to `cat` on Windows. Derive it from `BAZEL_SH` instead.

Fixes https://github.com/bazelbuild/rules_jvm_external/issues/36

### DIFF
--- a/coursier.bzl
+++ b/coursier.bzl
@@ -88,17 +88,18 @@ def generate_imports(repository_ctx, dep_tree, srcs_dep_tree = None):
             if artifact_path != None and artifact_path not in seen_imports:
                 seen_imports[artifact_path] = True
                 if repository_ctx.attr.use_unsafe_shared_cache:
-                  # If using unsafe shared cache, the path is absolute to the artifact in $COURSIER_CACHE
-                  artifact_relative_path = _relativize_and_symlink_file(repository_ctx, artifact_path)
+                    # If using unsafe shared cache, the path is absolute to the artifact in $COURSIER_CACHE
+                    artifact_relative_path = _relativize_and_symlink_file(repository_ctx, artifact_path)
                 else:
-                  # If not, it's a relative path to the one in output_base/external/$maven/v1/...
-                  artifact_relative_path = artifact_path
+                    # If not, it's a relative path to the one in output_base/external/$maven/v1/...
+                    artifact_relative_path = artifact_path
                 target_label = _escape(_strip_packaging_and_classifier(artifact["coord"]))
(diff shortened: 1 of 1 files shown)


jin:
Almost there: https://buildkite.com/bazel/android-testing/builds/232#6ec9473c-8ef8-4e80-8e00-a7dee4947060

jin:
To verify this fix works, we should merge https://github.com/bazelbuild/rules_jvm_external/pull/42 first.

meteorcloudy:
`type` is a builtin command in cmd.exe, when using CreateProcessW, we are not running under cmd.exe so `type` is reported as Not Found. 
Can you try using `repository_ctx.execute(["cmd.exe",  "/c", "type", "<file name>"])`  on Windows?

jin:
@meteorcloudy I've tried using that, but the call execution doesn't return, so it hangs. I'm not sure why. `cat` works fine though.

jin:
@meteorcloudy my bad, I just learned from László that `\k`\ deadlocks but `\c` doesn't. I was using `\k`. Let me try it with `\c`. Thanks!

jin:
Hm, it still doesn't work. It basically stalls at the repository_ctx.execute() call when using `type`, but it's fine if we use `cat`. Could we merge this first to get a working build on Buildkite? I've added a TODO to remove bash when Bazel is going bashless.

meteorcloudy:
@jin Sounds good, I'm ok with using `cat` temporarily, but Windows presubmit failed with a different error, can you take another look? Thanks!

jin:
We've removed the `windows` job from the .bazelci presubmit. The failure is due to a singlejar long paths issue and is unrelated to rules_jvm_external. It was also a test for the legacy gmaven_rules implementation, not the new one.

I think this is ready for a review.

coursier.bzl
```diff
@@ -278,8 +295,8 @@ def _coursier_fetch_impl(repository_ctx):
         artifact_coordinates.append(utils.artifact_coordinate(a))
         if "exclusions" in a:
             for e in a["exclusions"]:
-                exclusion_lines.append(":".join([a["group"], a["artifact"]]) + \
-                                       "--" + \
+                exclusion_lines.append(":".join([a["group"], a["artifact"]]) +
+                                       "--" +
```
dkelmer: Is the removal of the "\" intentional?

jin: ugh, didn't realize I included the formatting changes. Yes, this is buildifier's output.

dkelmer: There weren't too many, I'm ok merging with the formatting changes :)

PR #44, opened 2019-03-12
Create a migration script from gmaven_rules

jin:
From https://github.com/jin/rules_maven/pull/14, adapted for `rules_jvm_external`'s 1.0 release.

### DIFF
--- a/scripts/migrate_from_gmaven.sh
+++ b/scripts/migrate_from_gmaven.sh
@@ -1,0 +1,68 @@
+#!/bin/bash
+
+set -euo pipefail
+
+# Replace the load statements
+find ./ -type f -name BUILD* -exec sed -i 's/gmaven_rules/rules_jvm_external/g' {} \;
+
+# Replace the loaded gmaven_artifact symbol with maven_artifact
+find ./ -type f -name BUILD* -exec sed -i 's/"gmaven_artifact"/maven_artifact = "artifact"/g' {} \;
+
+# Replace the gmaven_artifact macro with maven_artifact
+find ./ -type f -name BUILD* -exec sed -i 's/gmaven_artifact(/maven_artifact(/g' {} \;
+
+# Remove the explicit packaging type from the coordinates
+find ./ -type f -name BUILD* -exec sed -i '/maven_artifact(/s/:aar//g; /maven_artifact(/s/:jar//g;' {} \;
+
+echo ""
+echo "Paste this into your WORKSPACE file:"
+echo "-------------------------"
+echo ""
+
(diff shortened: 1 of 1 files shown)


scripts/migrate_from_gmaven.sh
```diff
@@ -0,0 +1,68 @@
+#!/bin/bash
+
+set -euo pipefail
+
+# Replace the load statements
+find ./ -type f -name BUILD* -exec sed -i 's/gmaven_rules/rules_jvm_external/g' {} \;
+
+# Replace the loaded gmaven_artifact symbol with maven_artifact
+find ./ -type f -name BUILD* -exec sed -i 's/"gmaven_artifact"/maven_artifact = "artifact"/g' {} \;
```
dkelmer: Why not leave it at artifact?

scripts/migrate_from_gmaven.sh
```diff
@@ -0,0 +1,68 @@
+#!/bin/bash
+
+set -euo pipefail
+
+# Replace the load statements
+find ./ -type f -name BUILD* -exec sed -i 's/gmaven_rules/rules_jvm_external/g' {} \;
```
dkelmer: For all of these commands, is it possible people have these rules in bzl files?

scripts/migrate_from_gmaven.sh
```diff
@@ -0,0 +1,68 @@
...
+  | cut -d')' -f1 \
+  | cut -d'(' -f2 \
+  | sort \
+  | uniq \
+  | sed 's/^/        /; s/$/,/;'
+
+cat <<-EOF
+    ],
+    repositories = [
+        "https://maven.google.com",
```
dkelmer: gmaven couldn't handle repositories other than these but some artifacts did need to rely on additional repositories, right? Maybe add a comment after the copy-paste section to say that they may need to add more repositories?

jin:
Hm, on further thought I think it's worth taking a stab at using buildifier to do this instead.

scripts/migrate_from_gmaven.sh
jin: done.

scripts/migrate_from_gmaven.sh
jin: Yes, but I don't think this rule should cover them. It's quite basic, but we can add more flexible features in follow up PRs.

scripts/migrate_from_gmaven.sh
jin: Good point. Added a note.

PR #46, opened 2019-03-12
Add an example for android_local_test

jin:


examples/android_local_test/bazel/robolectric/gen_deps.py
```diff
@@ -0,0 +1,23 @@
+import sys
```
dkelmer: oh god i forgot about this

examples/android_local_test/bazel/robolectric/BUILD
```diff
@@ -0,0 +1,32 @@
+load("@rules_jvm_external//:defs.bzl", "artifact")
```
dkelmer: Pending thinking about my other comment, I wonder if this file should go somewhere more discoverable as an official recommendation of how to use Bazel and robolectric. Or maybe this is a sufficiently official location and should just be pointed to from the docs for discoverability

examples/android_local_test/WORKSPACE
```diff
@@ -0,0 +1,25 @@
...
+
+load("@rules_jvm_external//:defs.bzl", "maven_install")
+
+maven_install(
+    artifacts = [
+        "androidx.appcompat:appcompat:1.0.2",
+        "androidx.annotation:annotation:1.0.2",
+        "androidx.test.ext:junit:1.1.0",
+        "org.robolectric:robolectric:4.1",
+        "org.robolectric:android-all:8.1.0-robolectric-4611349",
```
dkelmer: This is a use case I probably should have thought of earlier: while it is uncommon to test against all apis, it is not uncommon to test against a few apis with your robolectric tests (you can use annotations in your test code to indicate which apis to test against). I definitely don't think this needs to be addressed in this CL, but note to self: the robolectric use case needs to be examined. I wonder how painful it would be to create a separate "maven_install" rule for each api you want to test against. The genrule/py binary that creates robolectric-deps.properties would need to be updated but other than that I think it would work

examples/android_local_test/src/test/BUILD
```diff
@@ -0,0 +1,14 @@
+load("@rules_jvm_external//:defs.bzl", "artifact")
+
+android_local_test(
+    name = "main_activity_test",
```
dkelmer: General file comment: why isn't the BUILD file next to MainActivityTest? If it was, and you called the target "MainActivityTest", you wouldn't need the test_class attribute. I don't have a strong preference, but I do think we recommend this practice at least for java tests

examples/android_local_test/src/test/java/com/example/bazel/MainActivityTest.java
```diff
@@ -0,0 +1,25 @@
...
+import org.robolectric.android.controller.ActivityController;
+
+import androidx.test.ext.junit.runners.AndroidJUnit4;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Junit Test using Robolectric with AssertJ matchers.
+ */
+@RunWith(AndroidJUnit4.class)
```
dkelmer: This should be RobolectricTestRunner.class

dkelmer:
Also, this is great :)

examples/android_local_test/bazel/robolectric/gen_deps.py
jin: 😄

examples/android_local_test/bazel/robolectric/BUILD
jin: 100%. Will think about this when we're refreshing the docs to include this project.

examples/android_local_test/WORKSPACE
jin: Yeah, this is one case where the artifact versioning pinning wouldn't work. I was thinking about a macro / attribute to "force" coursier to download additional artifacts.

examples/android_local_test/src/test/java/com/example/bazel/MainActivityTest.java
jin: 4.0 docs recommend AndroidJUnit4. RobolectricTestRunner was for 3.0. http://robolectric.org/blog/2018/05/09/robolectric-4-0-alpha/

examples/android_local_test/src/test/BUILD
jin: Hmm, I wanted to mirror the structure in src/main. I'll leave it as it is for now.

Issue #47, opened 2019-03-13
Add `maven_coordinates=` tag to targets


keith:
If you want to build a jar for upload that uses these rules, you need to generate a pom file. One way to do that is with this rule https://github.com/google/bazel-common/blob/f1115e0f777f08c3cdb115526c4e663005bec69b/tools/maven/pom_file.bzl

The way this rule works is it checks your dependencies for a tag in the format `maven_coordinates={actual maven coordinates}`. It would be useful if these rules propagated this tag (or some alternative solution to this) so that pom file generation could use the same coordinate values specified only once.

keith:
Alternatively maybe pom file generation is a rule that makes sense to have in this repository?

keith:
If we want to go down the tag path, I've submitted a PR for it https://github.com/bazelbuild/rules_jvm_external/pull/48

jin:
We should definitely expose the original Maven coordinate in the generated target, even if it's not used for POM file generation. Other than converting the target labels back to the Maven coordinates form, using tags SGTM, since we can record arbitrary strings in it and it doesn't interfere with analysis and execution.

PR #48, opened 2019-03-13
Add maven_coordinates tags to generated java_imports

keith:
This tag can be used by rules to generate a pom file that includes the
coordinates of all of a target's dependencies

Fixes https://github.com/bazelbuild/rules_jvm_external/issues/47

### DIFF
--- a/coursier.bzl
+++ b/coursier.bzl
@@ -181,6 +181,7 @@ def generate_imports(repository_ctx, dep_tree, srcs_dep_tree = None):
             target_import_string.append("".join(target_import_labels) + "\t],")
 
             # 5. Add a tag with the original maven coordinates for use generating pom files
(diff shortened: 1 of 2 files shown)


coursier.bzl
```diff
@@ -180,7 +180,19 @@ def generate_imports(repository_ctx, dep_tree, srcs_dep_tree = None):
 
             target_import_string.append("".join(target_import_labels) + "\t],")
 
-            # 5. Finish the java_import rule.
+            # 5. Add a tag with the original maven coordinates for use generating pom files
```
jin: Please add a link to bazel_common to show how this is used, or https://github.com/google/bazel-common/blob/f1115e0f777f08c3cdb115526c4e663005bec69b/tools/maven/pom_file.bzl#L177

jin:
CI looks good now.

coursier.bzl
keith: Done

keith:
Updated!

jin:
lgtm, @dkelmer?

keith:
Thanks!

PR #51, opened 2019-03-13
Unixify all paths and fix Windows builds

jin:
And add the integration tests targets to macOS and Windows `build_targets` for Buildkite.


jin:
It doesn't. This comes after reading the file.

PR #60, opened 2019-03-14
Split presubmit.yml into test targets and examples

jin:
The `tasks` and `platforms` blocks need to be in separate files, and we'll have to create a new Buildkite pipeline for the examples.


dkelmer:
Does this mean only the tasks in presubmit.yml will get run on presubmit? When will the examples.yml tasks get built/tested?

jin:
The filenames don't actually matter. I'll need to make a new Buildkite pipeline for examples.yml.

jin:
We don't really have a postsubmit, since presubmit covers the same set of targets.

dkelmer:
So the idea is that both of the buildkite pipelines will be triggered on
presubmit?

jin:
Yup!

PR #62, opened 2019-03-14
Add an example for generating POM <dependency> nodes using the generated targets

jin:


jin:
cc @keith

jin:
Also FYI @ronshapiro

examples/pom_file_generation/BUILD
```diff
@@ -0,0 +1,28 @@
...
+    srcs = ["java/com/example/bazel/Main.java"],
+    main_class = "com.example.bazel",
+    deps = [
+        artifact("com.google.inject:guice"),
+        artifact("com.google.guava:guava"),
+    ],
+)
+
+pom_file(
+    name = "my_binary_pom",
```
ronshapiro: I think it's confusing that you're creating a pom file for a binary. pom files are usually shipped with libraries

examples/pom_file_generation/README.md
```diff
@@ -0,0 +1,70 @@
+# Generating `<dependency>` nodes for a POM file
+
+This example shows how you can generate the `<dependencies>` section of a POM
+file using `rules_jvm_external`'s generated targets and the `pom_file` rule from
+`bazel-common`.
+
+For reference,
```
ronshapiro: I don't get why you're showing how to generate a pom file for a library that already exists, and I found it distracting. As a user, I don't particularly care about the pom files of my dependencies - I care about my own pom file. Once I see aopalliance I start wondering what that is and why I care

jin: Good point, removed.

examples/pom_file_generation/BUILD
jin: Switched to java_library. Thanks!

keith:
Do you have any thoughts on incorporating that rule in this repo? It seems like it aligns well with the other rules here and would make it a bit more mainstream

jin:
@keith that is up to @ronshapiro decide, and additionally, we don't yet have a plan for this repo on whether it should contain rules other than the ones for resolving and fetching artifacts (to keep its purpose focused).

ronshapiro:
I don't particularly care where that rule lives. I put it in bazel-common for my ease of development, but if it makes more sense elsewhere happy to entertain that. Maybe it should be it's own standalone repo? Or rules_pom? I don't know what the right organizational structure is

jin:
@ronshapiro Let's keep in bazel-common for now; I don't see an urgent need to affect downstream callers for this.

keith:
I guess my feeling from the outside looking at Bazel common was that it wasn't "as official" as other rules repos and therefore may be less stable. Because of that we copied that rule internally instead of referencing that repo as we do for other ones. I think in general that functionality would be useful to more folks so putting it somewhere more mainstream could be beneficial for others too

keith:
Sounds good, if we did choose to move it would could still leave a shim there that called through to this repo if you wanted

examples/pom_file_generation/README.md
```diff
@@ -0,0 +1,39 @@
...
+This example shows how you can generate the `<dependencies>` section of a POM
+file using `rules_jvm_external`'s generated targets and the `pom_file` rule from
+`bazel-common`.
+
+To generate the XML for the example Java library, build the `//:my_library_pom`
+target:
+
+```
+$ bazel build //:my_library_pom
+INFO: Analysed target //:my_library_pom (13 packages loaded, 444 targets configured).
```
dkelmer: I think just the command is helpful, the rest of the bazel output is mostly noise

PR #66, opened 2019-03-16
Add Stardoc API reference

### LINKED ISSUE #67: Generate an API reference for documentation

jin:
Use Stardoc to generate an API reference.

Added a script `generate_docs.sh` to:

1. Generate Stardoc markdown files for specified symbols
2. Concatenate all the files and header includes into `api.md`
3. Copy `api.md` into the `docs` folder and set the appropriate permissions on the file (rm executable bit, add rw for user)
4. Use `doctoc` to add a ToC.

Example API reference: https://github.com/bazelbuild/rules_jvm_external/blob/c75a1ac348ea084078615b462885068fe649e638/docs/api.md

Fixes #67


specs.bzl
```diff
@@ -4,26 +4,27 @@
...
-                    "password": String
-                }
-        }
-    """
+        url: A string containing the repository url (ex: "https://maven.google.com/").
+        user: A username for this Maven repository, if it requires authentication (ex: "johndoe").
+        password: A password for this Maven repository, if it requires authentication (ex: "example-password").
+    """
+
+    # Output Schema:
```
dkelmer: Why is this now commented out? Here and below

tests/unit/specs_test.bzl
```diff
@@ -1,4 +1,4 @@
-load("@bazel_skylib//lib:unittest.bzl", "asserts", "unittest")
+load("@bazel_skylib_local//lib:unittest.bzl", "asserts", "unittest")
```
dkelmer: Why the change to bazel_skylib_local? In the workspace you imported skylib as "io_bazel_skylib"

defs.bzl
```diff
@@ -12,8 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-load("@rules_jvm_external//:coursier.bzl", "coursier_fetch")
```
dkelmer: Did you need to remove this because stardoc couldn't handle it? What was the error?

tests/unit/specs_test.bzl
jin: That's io_bazel_skydoc, not io_bazel_skylib. skylib_repositories imports its own bazel_skylib at an another version: https://github.com/bazelbuild/skydoc/blob/f57c9880b9033241ad341ffc17a88d804cdfde3a/setup.bzl#L67

defs.bzl
jin: Yep: https://github.com/bazelbuild/skydoc/issues/174

specs.bzl
jin: I wasn't sure if this output schema is useful to end users to be included in Stardoc. Also, `Args:` needs to be in the final section of the macro documentation.

jin:
doctoc is an executable from npm: https://github.com/thlorenz/doctoc

It's the most hassle free tool I've found: `npm i -g doctoc && doctoc docs/api.md`

specs.bzl
dkelmer: Oh I didn't notice it was outside of the docs section. Yeah, good call. I think it's fine not in the official docs

defs.bzl
dkelmer: That's a bummer, I'll look into the bug. In the meantime, I'm not sure it's a good idea to remove the repo name. There are scenarios in which if the repository is used as an external repo (which this one always will be), if you don't fully qualify the load statements with the repository's name, then things break. I'm blanking on when that is, but it is definitely a recommendation to keep the fully qualified names. 

I might suggest checking in the generated docs.md as a separate PR and keep this PR open to add the actual stardoc target later (so that the @rules_jvm_external doesn't need to be removed). It's theoretically a little messy because it's easier for the code to get out of sync, but I don't think there will be too much churn to the api in the very immediate future

dkelmer:
So it takes as input the stardoc output?

jin:
Yup, and adds the ToC in-place.

defs.bzl
jin: Sounds good to me. I'll make a new PR with the changes to the documentation.

tests/unit/specs_test.bzl
dkelmer: Ah, yes, my mistake on the skylib/skydoc swap.

But I still have a comment: are the two versions of skylib incompatible (the version that's pulled in at the top level and the version that skydoc_repositories pulls in)? Because the skydoc_repositories one is inside of an "if_not_defined" block, so if the versions are compatible, the rules_jvm_external declaration of bazel_skylib can be moved on top of the call to skydoc_repositories and then there will only be one bazel_skylib

jin: Right, yeah but that means that I will make Stardoc depend on my version of bazel_skylib, which may or may not work. Should diamond splitting be used here instead?

dkelmer: The ideal scenario is that that will work :) 

The way it is now is effectively diamond splitting so it's fine to leave it as is if you don't merge the two bazel_skylib versions

jin: This is now fixed.

jin:
Updated to master, ready for another review.

jin:
@dkelmer ping, this is ready for another review

WORKSPACE
```diff
@@ -4,6 +4,35 @@ android_sdk_repository(name = "androidsdk")
...
+bazel_skylib_workspace()
+
+# End Skylib dependencies
+
+load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")
+
+git_repository(
+    name = "io_bazel_skydoc",
+    remote = "https://github.com/bazelbuild/skydoc.git",
+    commit = "e235d7d6dec0241261bdb13d7415f3373920e6fd",
```
c-parsons: This is fine, but, out of curiosity, was there something you needed since the last release?

specs.bzl
```diff
@@ -32,32 +33,30 @@ def _maven_repository(url, user = None, password = None):
...
+        artifact: The Maven artifact coordinate artifact name (ex: "guava").
+        version: The Maven artifact coordinate version name (ex: "27.0-jre").
+        packaging: The Maven packaging specifier (ex: "jar").
+        classifier: The Maven artifact classifier (ex: "javadoc").
+        override_license_types: An array of Bazel license type strings to use for this artifact's rules (overrides autodetection) (ex: ["notify"]).
+        exclusions: An array of exclusion objects to create exclusion specifiers for this artifact (ex: maven.exclusion("junit", "junit")).
+        neverlink: Determines if this artifact should be part of the runtime classpath.
     """
+
+    # Output Schema:
```
c-parsons: Want to add this to the Returns section of the docstring? (Your call)

WORKSPACE
jin: No, good catch. Switching it back to 0.3.0.

specs.bzl
jin: No, I would prefer not to.

jin:
Thanks for the review!

Issue #54, opened 2019-03-13
macOS, Windows CI: Ongoing concurrent download for https://oss.sonatype.org/...


jin:
https://buildkite.com/bazel/rules-jvm-external/builds/109#82cb6eb7-5f52-4bd0-af58-9d69bf8dd0f1

```
ERROR: /Users/buildkite/builds/buildkite-imacpro-7/bazel/rules-jvm-external/tests/integration/BUILD:3:1: no such package '@other_maven//': Unable to run coursier: Error while downloading https://oss.sonatype.org/content/repositories/releases/io/get-coursier/coursier-cli_2.12/1.1.0-M9/coursier-cli_2.12-1.1.0-M9-standalone.jar: Ongoing concurrent download for https://oss.sonatype.org/content/repositories/releases/io/get-coursier/coursier-cli_2.12/1.1.0-M9/coursier-cli_2.12-1.1.0-M9-standalone.jar, ignoring it
```

jin:
This doesn't seem to be happening on CI anymore.

jin:
Seeing this problem again on the Kotlin PR (https://github.com/bazelbuild/rules_jvm_external/pull/69) with macOS and Windows workers.

```
ERROR: D:/b/bk-windows-java8-dn2r/bazel/rules-jvm-external/tests/unit/kotlin/BUILD:3:1: no such package '@kotlin_tests//': Unable to run coursier: Error while downloading https://oss.sonatype.org/content/repositories/releases/io/get-coursier/coursier-cli_2.12/1.1.0-M9/coursier-cli_2.12-1.1.0-M9-standalone.jar: Ongoing concurrent download for https://oss.sonatype.org/content/repositories/releases/io/get-coursier/coursier-cli_2.12/1.1.0-M9/coursier-cli_2.12-1.1.0-M9-standalone.jar, ignoring it
--
  | and referenced by '//tests/unit/kotlin:inline_function_test'
[...]

PR #69, opened 2019-03-18
Add support for Kotlin JARs with inlined functions, and an Android Kotlin example app

### LINKED ISSUE #59: java_import invokes ijar, which clobbers inline functions in kotlin
Given bazelbuild/bazel#4549, the maven repository workspace shouldn't use java_import, but should use something similar to kt_jvm_import which does the same thing, but offers the compile_jar as the same as the regular jar.  This will affect things like kotlin.test whose `assertFailsWith<ExceptionType>() { throw ExceptionType() }` inlines, and therefore needs the method bodies at compile-time.

jin:
This PR adds support for Kotlin-built jars, tests, and an Android Kotlin example app.

I added the tests and app into the macOS and Linux CI configuration. rules_kotlin does not yet work on Windows (https://github.com/bazelbuild/rules_kotlin/issues/179), so the Windows failure in the CI is expected. We exclude the Kotlin unit tests for Windows in this PR so postsubmit is expected to be green.

Kotlin-sourced jars can contain metadata of inlined functions (`inline fun ...`), and regular java_import will strip these away with ijar, causing builds to fail. See https://github.com/bazelbuild/rules_jvm_external/pull/69/files#diff-90e077b9f9533f17dcef14f92cbae44c for the error message. To get around that, we create a simple `no_ijar_java_import` rule for internal use during the BUILD generation. 

To detect whether a JAR is Kotlin-sourced, we run `jar tf` on jars during BUILD file generation, and check for the existence of `kotlin_module` files. If it exists, we use the `no_ijar_java_import` rule class. If it does, we use the regular `java_import`, which has support for interface jars for more parallelizable downstream Java compilation.

Fixes #59


coursier.bzl
```diff
@@ -38,6 +41,12 @@ def _is_linux(repository_ctx):
 def _is_macos(repository_ctx):
     return repository_ctx.os.name.find("mac") != -1
 
+def _contains_kotlin_module(repository_ctx, artifact_relative_path):
+    exec_result = repository_ctx.execute(["jar", "tf", artifact_relative_path])
```
jin: TODO: try to use `JAVA_HOME` to infer the location of `jar`

jin:
Excellent idea from @davidsantiago: perhaps we should generate, by default, `no_ijar_java_import` targets instead, and let users opt into `java_import` if they want to use ijar. This way, there is no ambiguity on whether there was ijar processing on the imported jars.

jin:
As discussed offline, let's go with completely replacing `java_import` with its no-ijar counterpart.

coursier.bzl
```diff
@@ -382,7 +389,7 @@ def _coursier_fetch_impl(repository_ctx):
     cmd = _generate_coursier_command(repository_ctx)
     cmd.extend(["fetch"])
     cmd.extend(artifact_coordinates)
-    cmd.extend(["--artifact-type", ",".join(_COURSIER_PACKAGING_TYPES)])
+    cmd.extend(["--artifact-type", ",".join(_COURSIER_PACKAGING_TYPES + ["src"])])
```
jin: drive-by fix to add `src` back into the fetch_sources branch.

coursier.bzl
```diff
@@ -73,16 +76,17 @@ def _relativize_and_symlink_file(repository_ctx, absolute_path):
...
-  # For all potential reverse dep artifacts,
-  for maybe_rdep in dep_tree["dependencies"]:
-      # For all dependencies of this artifact,
-      for maybe_rdep_coord in maybe_rdep["dependencies"]:
-          # If this artifact depends on the missing artifact,
-          if maybe_rdep_coord == coord:
-              # Then this artifact is an rdep :-)
-              reverse_deps.append(maybe_rdep)
-  return reverse_deps
+    reverse_deps = []
```
jin: drive by formatting fix.

.bazelci/presubmit.yml
```diff
@@ -44,4 +44,9 @@ platforms:
     # TODO(jin): fix long paths issue on Windows
     # - "//examples:app"
     test_targets:
+    - "--"
```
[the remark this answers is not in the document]
jin: allows using the target subtraction syntax `-//:target`

.bazelci/presubmit.yml
```diff
@@ -35,7 +35,11 @@ platforms:
     # - "//tests/integration:all"
     - "//examples:app"
     test_targets:
+    - "--"
     - "//tests/unit/..."
+    # TODO(jin): Concurrent download problems on macOS
+    # https://github.com/bazelbuild/rules_jvm_external/issues/54
+    - "-//tests/unit/kotlin/..."
```
jin: FYI: had to disable the test on the Mac too, due to network / race problems in https://github.com/bazelbuild/rules_jvm_external/issues/54

jin: Hmm maybe the CI can't resolve more than one maven_install repository concurrently?

jin: Yup, that's exactly it.

PR #81, opened 2019-03-26
Add .bazelignore on the thirdparty files and examples so we can run bazel test //...

jin:
Also stop building //examples:app for gmaven on presubmit

Failures are expected, since the `examples/` directory is ignored with `//...`


jin:
Nvm, this isn't working as I thought it would.

PR #82, opened 2019-03-26
Add proxy args to coursier command

Zetten:
Convert standard *nix proxy environment variables into Java-style
property arguments, and add them to the generated coursier command.

### DIFF
--- a/coursier.bzl
+++ b/coursier.bzl
@@ -308,15 +308,48 @@ def _generate_coursier_command(repository_ctx):
         # The -noverify option seems to be required after the proguarding step
         # of the main JAR of coursier.
         java = repository_ctx.path(java_home + "/bin/java")
-        cmd = [java, "-noverify", "-jar", coursier]
+        cmd = [java, "-noverify", "-jar"] + _get_java_proxy_args(repository_ctx) + [coursier]
     elif repository_ctx.which("java") != None:
         # Use 'java' from $PATH
-        cmd = [repository_ctx.which("java"), "-noverify", "-jar", coursier]
+        cmd = [repository_ctx.which("java"), "-noverify", "-jar"] + _get_java_proxy_args(repository_ctx) + [coursier]
     else:
         # Try to execute coursier directly
-        cmd = [coursier]
+        cmd = [coursier] + ["-J%s" % arg for arg in _get_java_proxy_args(repository_ctx)]
 
     return cmd
+
+# Extract the well-known environment variables http_proxy, https_proxy and
+# no_proxy and convert them to java.net-compatible property arguments.
+def _get_java_proxy_args(repository_ctx):
+    # Check both lower- and upper-case versions of the environment variables, preferring the former
+    http_proxy = repository_ctx.os.environ.get("http_proxy", repository_ctx.os.environ.get("HTTP_PROXY"))
+    https_proxy = repository_ctx.os.environ.get("https_proxy", repository_ctx.os.environ.get("HTTPS_PROXY"))
+    no_proxy = repository_ctx.os.environ.get("no_proxy", repository_ctx.os.environ.get("NO_PROXY"))
+
+    proxy_args = []
+
+    # Extract the host and port from a standard proxy URL:
+    # http://proxy.example.com:3128 -> ["proxy.example.com", "3128"]
+    if http_proxy != None:
+        proxy = http_proxy.split("://", 1)[1].split(":", 1)
+        proxy_args.extend([
+            "-Dhttp.proxyHost=%s" % proxy[0],
+            "-Dhttp.proxyPort=%s" % proxy[1],
+        ])
+
+    if https_proxy != None:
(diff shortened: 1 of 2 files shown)


coursier.bzl
```diff
@@ -308,16 +308,44 @@ def _generate_coursier_command(repository_ctx):
...
-        cmd = [java, "-noverify", "-jar", coursier]
+        cmd = [java, "-noverify", "-jar"] + _get_java_proxy_args(repository_ctx) + [coursier]
     elif repository_ctx.which("java") != None:
         # Use 'java' from $PATH
-        cmd = [repository_ctx.which("java"), "-noverify", "-jar", coursier]
+        cmd = [repository_ctx.which("java"), "-noverify", "-jar"] + _get_java_proxy_args(repository_ctx) + [coursier]
     else:
         # Try to execute coursier directly
-        cmd = [coursier]
+        cmd = [coursier] + _get_java_proxy_args(repository_ctx)
```
jin: Hm, this this work? 

```
→ third_party/coursier/coursier -Dhttp.proxyHost=127.0.0.1
Unrecognized argument: -Dhttp.proxyHost=127.0.0.1
```

coursier.bzl
```diff
@@ -308,16 +308,44 @@ def _generate_coursier_command(repository_ctx):
...
 
     return cmd
 
+# Extract the well-known environment variables http_proxy, https_proxy and
+# no_proxy and convert them to java.net-compatible property arguments.
+def _get_java_proxy_args(repository_ctx):
+    http_proxy = repository_ctx.os.environ.get("http_proxy") if repository_ctx.os.environ.get("http_proxy") else repository_ctx.os.environ.get("HTTP_PROXY")
```
jin: Does this also work with the capitalized versions of the environment variables? Is that common?

coursier.bzl
```diff
@@ -308,16 +308,44 @@ def _generate_coursier_command(repository_ctx):
...
+    https_proxy = repository_ctx.os.environ.get("https_proxy") if repository_ctx.os.environ.get("https_proxy") else repository_ctx.os.environ.get("HTTPS_PROXY")
+    no_proxy = repository_ctx.os.environ.get("no_proxy") if repository_ctx.os.environ.get("no_proxy") else repository_ctx.os.environ.get("NO_PROXY")
+
+    proxy_args = []
+
+    if http_proxy != None:
```
jin: Please add some comments on how these string manipulations would look like on an example.

jin:
Also, please add the environment variables to:

https://github.com/bazelbuild/rules_jvm_external/blob/85196d956db2efd0edbf4f2014ee5e2fce67188a/coursier.bzl#L467

so that the rule can track env var changes and re-run accordingly.

coursier.bzl
Zetten: Good point, I misread https://github.com/coursier/coursier/blob/master/doc/docs/other-proxy.md

coursier.bzl
Zetten: I've never seen cases other than all lower- or all upper-case, although it's possible. Bazel only seems to implicitly work with the two forms I've checked here though: https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/bazel/repository/downloader/ProxyHelper.java#L46

coursier.bzl
Zetten: Ack

Zetten:
> Also, please add the environment variables to:
> 
> [rules_jvm_external/coursier.bzl](https://github.com/bazelbuild/rules_jvm_external/blob/85196d956db2efd0edbf4f2014ee5e2fce67188a/coursier.bzl#L467)
> 
> Line 467 in [85196d9](/bazelbuild/rules_jvm_external/commit/85196d956db2efd0edbf4f2014ee5e2fce67188a)
>  environ = ["JAVA_HOME"], 
> 
> so that the rule can track env var changes and re-run accordingly.

The proxy variables are implicitly included by Bazel in repository_ctx, and should be automatically used when determining re-runs, so I'm not sure they need to be explicitly added here. I'm happy to add them if you definitely want it though.

https://docs.bazel.build/versions/master/external.html#using-proxies

jin:
@Zetten I see, didn't know that :) Feel free to skip it then.

coursier.bzl
jin: I don't think there's special handling within a repository_ctx.os.env.get, is there? env.get is just a dictionary get. Feel free to skip this, we can add capitalized env var support when users request for it.

Zetten:
@jin Yeah, for better or worse I spend a large amount of my day wrestling with proxies so I've had to get well acquainted :grin:

Having just tested it, I think the default behaviour is actually to ignore changes once the repository rule is resolved... I've added the vars to the list now anyway, just for safety.

coursier.bzl
Zetten: There's at least a `default=` so we can shorten the assignment.

PR #87, opened 2019-03-28
Turn the shell integration tests into Java tests

jin:
This PR switches the shell integration tests into Java tests. I converted the tests for artifact exclusion and unsafe shared cache. The simple integration test is testing more on the Java rules' deploy jar logic than this project, so I skipped that.

The general idea of these tests is to depend on a jar provided by rules_jvm_external, and look into the classpath for assertions. This is identical to how an end user would depend on the jars, so I think it's a pretty realistic integration test setup.

Also removed a bunch of maven_installs from the WORKSPACE file.


jin:
Tests are passing, this is now ready for review.

tests/integration/UnsafeSharedCacheTest.java
```diff
@@ -0,0 +1,51 @@
...
+import static org.junit.Assert.assertTrue;
+
+public class UnsafeSharedCacheTest {
+
+  private static String OS = System.getProperty("os.name").toLowerCase();
+
+  @Test
+  public void test_jarsOnClassPath_areInTheSharedCache() throws IOException, URISyntaxException {
+    // No support for symlinks on Windows. Let's skip that.
+    if (OS.indexOf("win") == -1) {
```
dkelmer: To reduce nesting a nice pattern is to check the opposite condition first and exit early. So in this case it would be something like
```
if (OS.indexOf("win") != -1) { System.exit(); } 
/* rest of code un-nested one level */
```
or something like that

PR #88, opened 2019-03-28
Replace third_party/coursier/coursier bootstrap jar with the standalone jar

jin:
This fixes https://github.com/bazelbuild/rules_jvm_external/issues/54

Previously, third_party/coursier/coursier is a bootstrap jar that downloads the standalone jar at runtime. When we have multiple maven_install calls, this causes a race condition as seen in #54. By switching to the standalone jar, there's no need to download any classes at runtime, so users can use as many maven_installs as they want. 

Also use repository_ctx.download to get the standalone jar at build time.

Also remove all the shell scripts pretend-tests in preparation for them to be replaced by Java tests in #87


jin:
Hold on one second, maybe we can use repository_ctx.download isntead so we don't need to vendor this.

jin:
K, should be good to go. The failures from the Linux tests are expected because third_party/coursier/coursier is gone.

jin:
I just realized that with this change, you can't really use rules_jvm_external in a network without internet access. Maybe we should keep the bootstrap jar around as an alternative since it's so small?

dkelmer:
You can't use these rules without an internet connection anyway. Coursier needs to download the jars somehow

dkelmer:
Oh you could point to a local repository

.bazelci/presubmit.yml
```diff
@@ -2,37 +2,21 @@
 platforms:
   ubuntu1404:
     # TODO(jingwen): switch these to sh_test and port them to macos and windows
-    shell_commands:
-    - ./tests/integration/simple_integration_test.sh
```
dkelmer: Were these setup shell scripts? Can they be deleted if they're not being used?

jin: Nope, these were tests from the early days of this project. I'm converting them into Java tests in #87.

dkelmer: so shouldn't they stay around until #87 is in? :)

jin: oops. I'll get #87 ready asap.

PR #96, opened 2019-04-02
Add a example for Spring Boot

jin:
This adds an example web application built with Spring Boot and rules_jvm_external, as requested by a user.

https://spring.io/guides/gs/spring-boot/

### DIFF
--- a/examples/spring_boot/src/main/java/hello/BUILD.bazel
+++ b/examples/spring_boot/src/main/java/hello/BUILD.bazel
@@ -1,0 +1,20 @@
+java_library(
+    name = "lib",
+    srcs = glob(["*.java"]),
+    deps = [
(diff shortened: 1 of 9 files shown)


examples/spring_boot/src/main/java/hello/BUILD.bazel
```diff
@@ -0,0 +1,20 @@
+java_library(
+    name = "lib",
+    srcs = glob(["*.java"]),
+    deps = [
+        "@maven//:org_springframework_boot_spring_boot_starter_web",
```
dkelmer: You only specified this one. I think it's a bad practice to not specify all direct deps used and relying on rules_jvm_external to pull direct deps in from a different direct dep. Could you add the other deps to maven_install?

jin: Done.

PR #97, opened 2019-04-03
Neverlink support

jin:
Add neverlink support in the artifact spec. 

Some artifacts, like com.squareup.javapoet, are meant to be used only at compile time, and should not be linked into the runtime classpath / APK. This allows users to set the `neverlink` attribute in the artifact spec.


coursier.bzl
```diff
@@ -231,7 +231,24 @@ def generate_imports(repository_ctx, dep_tree, srcs_dep_tree = None):
...
+            # 	],
+            #   tags = ["maven_coordinates=org.hamcrest:hamcrest.library:1.3"],
+            #   neverlink = True,
+            if (neverlink_artifacts.get(_strip_packaging_and_classifier_and_version(artifact["coord"]))):
+                target_import_string.append("\tneverlink = True,")
+
+
+            # 7. Finish the java_import rule.
             #
             # java_import(
```
dkelmer: Each of these numbered steps is cumulative, right? So #7 should have neverlink too

jin: Good catch, done.

jin:
Using a java_test integration test to test the neverlink functionality instead. I also rebased this on top of https://github.com/bazelbuild/rules_jvm_external/pull/87, so that should go in first.

Issue #98, opened 2019-04-05
Special handling needed for pom-only dependencies


mpassell:
I was really excited to try this out (v1.2) and went with a single library I'm familiar with that brings in a whole big tree of dependencies.

```python
maven_install(
    artifacts = [
        "nz.ac.waikato.cms.weka:weka-stable:3.8.1",
    ],
    repositories = [
        "https://repo1.maven.org/maven2",
    ],
)
```

When I ran `bazel query @maven//:all` (thanks @jin), it blew up with the following, because `com.github.fommil.netlib:all` (transitive dep) is a pom-only dependency and exists solely to aggregate other dependencies:
```
The artifact for com.github.fommil.netlib:all:pom:1.1.2 was not downloaded. Perhaps its packaging type is
not one of: jar,aar,bundle,eclipse-plugin?
```

jin:
https://github.com/bazelbuild/rules_jvm_external/pull/99 should fix this.

jin:
```
maven_install(
    name = "issue98",
    artifacts = [
        "nz.ac.waikato.cms.weka:weka-stable:3.8.1",
    ],
    repositories = [
        "https://repo1.maven.org/maven2",
    ],
)
```

```
$ bazel query @issue98//:all 
@issue98//:nz_ac_waikato_cms_weka_weka_stable
@issue98//:nz_ac_waikato_cms_weka_weka_stable_3_8_1
@issue98//:nz_ac_waikato_cms_weka_thirdparty_java_cup_11b_runtime
@issue98//:nz_ac_waikato_cms_weka_thirdparty_java_cup_11b_runtime_2015_03_26
@issue98//:nz_ac_waikato_cms_weka_thirdparty_java_cup_11b
@issue98//:nz_ac_waikato_cms_weka_thirdparty_java_cup_11b_2015_03_26
@issue98//:nz_ac_waikato_cms_weka_thirdparty_bounce
@issue98//:nz_ac_waikato_cms_weka_thirdparty_bounce_0_18
@issue98//:net_sourceforge_f2j_arpack_combined_all
@issue98//:net_sourceforge_f2j_arpack_combined_all_0_1
@issue98//:com_googlecode_netlib_java_netlib_java
@issue98//:com_googlecode_netlib_java_netlib_java_1_1
@issue98//:com_googlecode_matrix_toolkits_java_mtj
@issue98//:com_googlecode_matrix_toolkits_java_mtj_1_0_4
[...]

PR #99, opened 2019-04-05
Support pom-only dependencies

jin:
Fixes https://github.com/bazelbuild/rules_jvm_external/issues/98

Add a whitelist of artifacts that are "aggregators", that is, they only have a POM file and no jar. 

If Coursier tells us that a particular artifact doesn't have a file, we generate a `java_library` instead, and re-export its deps.


coursier.bzl
```diff
@@ -101,9 +108,6 @@ def generate_imports(repository_ctx, dep_tree, srcs_dep_tree = None):
     # The list of java_import/aar_import declaration strings to be joined at the end
     all_imports = []
 
-    # A mapping of FQN to the artifact's sha256 checksum
```
jin: drive-by removal of the checksumming code

coursier.bzl
```diff
@@ -125,29 +129,17 @@ def generate_imports(repository_ctx, dep_tree, srcs_dep_tree = None):
...
                     # If not, it's a relative path to the one in output_base/external/$maven/v1/...
                     artifact_relative_path = _normalize_to_unix_path(artifact_path)
-                target_label = _escape(_strip_packaging_and_classifier(artifact["coord"]))
+                target_label = _escape(_strip_packaging_classifier(artifact["coord"]))
                 srcjar_paths[target_label] = artifact_relative_path
 
     # Iterate through the list of artifacts, and generate the target declaration strings.
     for artifact in dep_tree["dependencies"]:
         artifact_path = artifact["file"]
+        target_label = _escape(_strip_packaging_classifier(artifact["coord"]))
```
jin: instead of using the artifact path as the key to dedupe target generation, we use the target label as the key, since it's canonical and unique anyway

coursier.bzl
```diff
@@ -31,12 +34,16 @@ load("@{repository_name}//:jvm_import.bzl", "jvm_import")
...
+        coord = coord.replace(":%s:" % packaging_type, ":")
+    for classifier_type in ["sources", "natives"]:
+        coord = coord.replace(":%s:" % classifier_type, ":")
+
+    return coord
 
-def _strip_packaging_and_classifier_and_version(coord):
-    return ":".join(_strip_packaging_and_classifier(coord).split(":")[:-1])
+def _strip_packaging_classifier_version(coord):
+    return ":".join(_strip_packaging_classifier(coord).split(":")[:-1])
```
dkelmer: Why the change to the name? _strip_packaging_and_classifier is clearer (without the and it sounds like you're stripping a single thing called "packaging classifier"). Also with the version method

coursier.bzl
```diff
@@ -31,12 +34,16 @@ load("@{repository_name}//:jvm_import.bzl", "jvm_import")
 # download them if it's not asked to to resolve "eclipse-plugin".
 _COURSIER_PACKAGING_TYPES = ["jar", "aar", "bundle", "eclipse-plugin"]
 
-# Super hacky :(
-def _strip_packaging_and_classifier(coord):
-    return coord.replace(":jar:", ":").replace(":aar:", ":").replace(":sources:", ":")
+def _strip_packaging_classifier(coord):
+    for packaging_type in _COURSIER_PACKAGING_TYPES + ["pom"]:
```
dkelmer: Add "pom" to _COURSIER_PACKAGING_TYPES instead of adding it here

private/special_artifacts.bzl
```diff
@@ -0,0 +1,3 @@
+POM_ONLY_ARTIFACTS = {
+    "com.github.fommil.netlib:all": True,
```
dkelmer: Does this mean that we need to hardcode all pom only artifacts?

coursier.bzl
```diff
@@ -263,7 +278,7 @@ def generate_imports(repository_ctx, dep_tree, srcs_dep_tree = None):
             reverse_dep_coords = [reverse_dep["coord"] for reverse_dep in reverse_deps]
             reverse_dep_pom_paths = [
                 repository_ctx.path(reverse_dep["file"].replace(".jar", ".pom").replace(".aar", ".pom"))
-                for reverse_dep in reverse_deps
+                for reverse_dep in reverse_deps if reverse_dep["file"]
```
dkelmer: I think the "if reverse_dep["file"] part isn't necessary because the previous elif clause will catch the pom only dependencies which are the ones that don't have a file

private/special_artifacts.bzl
jin: Yes. My estimate is that this is a rare occurrence, so we probably shouldn't bake huge custom logic into the rule.

coursier.bzl
jin: Reverted.

coursier.bzl
jin: _COURSIER_PACKAGING_TYPES is fed into the coursier CLI, which doesn't recognize `pom` as a packaging type. I'll add a comment about it.

coursier.bzl
jin: Great catch. Done.

PR #100, opened 2019-04-05
(example): add scala akka http application.

dgarzon:
This PR adds and example of a fairly complex Scala application, and how to use the rules to successfully build / run / test it. In comparison to `bazel-deps`, we now need to define extra `deps` in order for `rules_scala` to pick them up. No big deal! @jin awesome job on this project!

### DIFF
--- a/.bazelci/examples.yml
+++ b/.bazelci/examples.yml
@@ -99,8 +99,33 @@ tasks:
     - "//..."
   spring-boot-windows:
     name: "Spring boot example"
-    platform: windows 
+    platform: windows
     working_directory: examples/spring_boot
+    build_targets:
+    - "//..."
+    test_targets:
+    - "//..."
+
+  scala-akka-linux:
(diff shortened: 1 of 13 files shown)


dgarzon:
I signed it!

jin:
Nice, thank you! Do you mind adding some entries to the CI config for the build and test targets, like the ones for the Spring Boot application?

https://github.com/bazelbuild/rules_jvm_external/blob/40d2e50d7d140eea39b353709854a26b7a870cd4/.bazelci/examples.yml#L84

Feel free to add it for all three platforms; we can delete them if they fail on postsubmit later.

jin:
also, can you rename the example directory from `scala` to `scala_akka`? So it's clearer that it's an Akka application from browsing the file tree.

dgarzon:
Will do @jin . Will push once I address all your comments! Thank you for the review ❤️

Issue #101, opened 2019-04-05
Artifact with classifier looking at wrong path


dgarzon:
Hi,

I am trying to fetch some artifacts that have a classifier. Unfortunately, it seems as if this feature is not properly implemented.

For example:

```
maven.artifact(
    group = "com.digitalasset",
    artifact = "damlc",
    version = "100.12.1",
    classifier = "osx",
),
maven.artifact(
    group = "com.digitalasset",
    artifact = "damlc",
    version = "100.12.1",
    classifier = "linux",
)
```

Yields:

```
https://digitalassetsdk.bintray.com/DigitalAssetSDK/com/digitalasset/damlc/osx/damlc-osx.pom
https://digitalassetsdk.bintray.com/DigitalAssetSDK/com/digitalasset/damlc/linux/damlc-linux.pom
```

When according to: https://stackoverflow.com/a/47733421. It should yield:

```
https://digitalassetsdk.bintray.com/DigitalAssetSDK/com/digitalasset/damlc/100.12.1/damlc-100.12.1.pom
https://digitalassetsdk.bintray.com/DigitalAssetSDK/com/digitalasset/damlc/100.12.1/damlc-100.12.1-osx.jar
https://digitalassetsdk.bintray.com/DigitalAssetSDK/com/digitalasset/damlc/100.12.1/damlc-100.12.1-linux.jar
```

jin:
I think this is now fixed by #102. Thanks!

PR #102, opened 2019-04-08
(hotfix): proper classifier resolution.

dgarzon:
According to: https://github.com/coursier/coursier/pull/735, the proper way to specify fine-grained `classifier` is by doing: `group:artifact:packaging:version,classifier=xxx`. This PR aims to align the implementation with what `coursier` expects and address https://github.com/bazelbuild/rules_jvm_external/issues/101. I would like to add an integration test of some sorts, can someone provide me with guidance on how to add one and run them locally? Thanks!


jin:
For the integration test, you can add the artifact with the classifier in:

https://github.com/bazelbuild/rules_jvm_external/blob/40d2e50d7d140eea39b353709854a26b7a870cd4/WORKSPACE#L96

and then, add the generated target in:

https://github.com/bazelbuild/rules_jvm_external/blob/40d2e50d7d140eea39b353709854a26b7a870cd4/tests/unit/build_tests/BUILD#L8

specs.bzl
```diff
@@ -233,11 +233,9 @@ json = struct(
...
 
 def _artifact_to_coord(artifact):
-    return artifact["group"] + ":" + \
-        artifact["artifact"] + ":" + \
-        ((artifact["packaging"] + ":") if artifact.get("packaging") != None else "") + \
-        ((artifact["classifier"] + ":") if artifact.get("classifier") != None else "") + \
-        artifact["version"]
+    packaging = (artifact["packaging"] + ":") if artifact.get("packaging") != None else ""
+    classifier = (",classifier=" + artifact["classifier"]) if artifact.get("classifier") != None else ""
+    return artifact["group"] + ":" + artifact["artifact"] + ":" + packaging + artifact["version"] + classifier
```
jin: do you mind adding a comment here about the Coursier-specific "classifier=" syntax?

dgarzon:
Hi @jin after doing some more research, I found out that the `packaging` field is also not supported on a granular level by `coursier`. Right now when we call `_artifact_to_coord(artifact` the `packaging gets added to the resulting coordinate, which is wrong. I will proceed and push with the fix for this, as well as a comment explaining the `classifier` syntax.

dgarzon:
Sorry for bothering you so much @jin , but I am stuck now and have no idea why. My changes partially fix the issue with `classifier` and `packaging`, but not entirely. See, the jar's with `classifier` are downloaded successfully, but they are not found in the generated `BUILD` file with all the generated targets. Any ideas why this can be happening? Can you point me in the right direction?

dgarzon:
@jin After some debugging, I was able to get it working 👍 Here are some of the things I found:

1. Coursier does not consider `javadoc` or `sources` as classifiers. Instead, it expects the cli to be invoked with `--javadoc` or `--sources`. Therefore, we will need to modify the algorithm in `generate_imports` to account for when the user wants both `sources` and `javadoc` (out of scope for this PR).
2. Our implementation of `_coursier_fetch_impl` was wrong for when `fetch_sources == True` as it was fetching both `sources` and `default` artifacts. The implementation would choke in `generate_imports` when a given artifact had no sources, as it was marked as `seen_import` when analysing sources. I have fixed the coursier command to only fetch sources now, and we should do something similar for `javadoc` in the future.

Hope this makes sense!

jin:
Thanks! I was trying to debug too, but looks like you got there first.

> Coursier does not consider javadoc or sources as classifiers. Instead, it expects the cli to be invoked with --javadoc or --sources. Therefore, we will need to modify the algorithm in generate_imports to account for when the user wants both sources and javadoc (out of scope for this PR).

I see, when implementing source jar support, I was slightly confused by this too. This makes sense. Leaving out javadoc support is fine.

>Our implementation of _coursier_fetch_impl was wrong for when fetch_sources == True as it was fetching both sources and default artifacts. The implementation would choke in generate_imports when a given artifact had no sources, as it was marked as seen_import when analysing sources. I have fixed the coursier command to only fetch sources now, and we should do something similar for javadoc in the future.

"as it was marked as seen_import when analysing sources." -> I had this hunch too :) I think we should add `else` branch in `generate_imports` to catch artifacts that were fetched but nothing was generated for them. That's out of scope of this PR too.

dgarzon:
Thanks @jin for the review and merge! Can we please get a release soon? I would like to use this in my project 👍

jin:
I'll make a release now.

jin:
@dgarzon https://github.com/bazelbuild/rules_jvm_external/releases/tag/1.3

PR #108, opened 2019-04-15
Use bazel_tools/jdk from upper level

aefimov:
This will force using bundled jdk from Bazel instead of separate JDK installed. For example, before that change this cause forcing setup default global JDK on Mac and will show up Dialog "Please install JDK".


jin:
I don't think we want to have a dependency on Bazel's bundled JDK, since these rules inherently depend on system or the downloaded full JDKs.

Issue #119, opened 2019-04-17
POM-only dependencies still don't work


mandrean:
While `v2.0.1` introduced some level of support for POM-only dependencies, I'm still having issues with them.

### How to reproduce:
1. Use `rules_jvm_external` `v2.0.1`
2. Try to `maven_install` the dependency `io.vavr:vavr-parent:0.10.0`
3. Watch the world explode

Ref: http://central.maven.org/maven2/io/vavr/vavr-parent/0.10.0/vavr-parent-0.10.0.pom

enriched:
I believe that I might be running into this issue as well, with a workspace file like:

```py
RULES_JVM_EXTERNAL_TAG = "2.0.1"

RULES_JVM_EXTERNAL_SHA = "55e8d3951647ae3dffde22b4f7f8dee11b3f70f3f89424713debd7076197eaca"

http_archive(
    name = "rules_jvm_external",
    sha256 = RULES_JVM_EXTERNAL_SHA,
    strip_prefix = "rules_jvm_external-%s" % RULES_JVM_EXTERNAL_TAG,
    url = "https://github.com/bazelbuild/rules_jvm_external/archive/%s.zip" % RULES_JVM_EXTERNAL_TAG,
)

load("@rules_jvm_external//:defs.bzl", "maven_install")

maven_install(
    artifacts = ["org.apache.flink:flink-test-utils_2.12:1.8.0"],
    repositories = [
        "https://repo1.maven.org/maven2",
    ]
)
```

I get the error:

```
The artifact for org.apache.flink:flink-runtime_2.12:test-jar:tests:1.8.0 was not downloaded. Perhaps its packaging type is
not one of: jar,aar,bundle,eclipse-plugin?

It is also possible that the packaging type of org.apache.flink:flink-runtime_2.12:test-jar:tests:1.8.0 is specified
incorrectly in the POM file of an artifact that depends on it. For example,
org.apache.flink:flink-runtime_2.12:test-jar:tests:1.8.0 may be an AAR, but the dependent's POM file specified its `<type>`
value to be a JAR.

The artifact(s) depending on org.apache.flink:flink-runtime_2.12:test-jar:tests:1.8.0 are:



and their POM files are located at:



---

[...]

PR #120, opened 2019-04-17
Add support for test-jar packaging type

jin:
See https://github.com/bazelbuild/rules_jvm_external/issues/119#issuecomment-484278260

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -35,6 +35,24 @@ maven_install(
     use_unsafe_shared_cache = True,
 )
 
+maven_install(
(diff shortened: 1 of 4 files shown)


WORKSPACE
```diff
@@ -35,6 +35,24 @@ maven_install(
     use_unsafe_shared_cache = True,
 )
 
+maven_install(
+    name = "exclusion_testing",
```
jin: I moved this out of `@regression_testing` because other artifacts were bringing animal-sniffer-annotations back in.

PR #131, opened 2019-04-26
Fix passing credentials that are not URL-safe

TheMarex:
Passing the credentials as part of the repository URL is deprecated in coursier and has some unfortunate side-effects, such as the escaping of non-URL-safe usernames/password does not work correctly. (if you pass it in URL-escaped it does not work, if you pass it in plain it does not work either)

This PR does two things:

1. Use `--credentials` for `coursier` if the user passes in the credentials explicitly. If the user supplies an string for the repository URL it is passed-through as-is to coursier (which is also the current behavior). This means the having the basic-auth in the URL still works.

### DIFF
--- a/coursier.bzl
+++ b/coursier.bzl
@@ -702,7 +702,9 @@ def _coursier_fetch_impl(repository_ctx):
         repository_ctx.file("exclusion-file.txt", "\n".join(exclusion_lines), False)
         cmd.extend(["--local-exclude-file", "exclusion-file.txt"])
     for repository in repositories:
-        cmd.extend(["--repository", utils.repo_url(repository)])
+        cmd.extend(["--repository", repository["repo_url"]])
+        if "credentials" in repository:
+            cmd.extend(["--credentials", utils.repo_credentials(repository)])
     for a in excluded_artifacts:
         cmd.extend(["--exclude", ":".join([a["group"], a["artifact"]])])
     if not repository_ctx.attr.use_unsafe_shared_cache:
--- a/specs.bzl
+++ b/specs.bzl
@@ -251,14 +251,22 @@ def _artifact_to_coord(artifact):
     classifier = (",classifier=" + artifact["classifier"]) if artifact.get("classifier") != None else ""
     return artifact["group"] + ":" + artifact["artifact"] + ":" + artifact["version"] + classifier
 
-def _repository_url(repository_spec):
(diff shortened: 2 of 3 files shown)


TheMarex:
@googlebot I signed it!

coursier.bzl
```diff
@@ -510,7 +510,10 @@ def _coursier_fetch_impl(repository_ctx):
         repository_ctx.file("exclusion-file.txt", "\n".join(exclusion_lines), False)
         cmd.extend(["--local-exclude-file", "exclusion-file.txt"])
     for repository in repositories:
-        cmd.extend(["--repository", utils.repo_url(repository)])
+        cmd.extend(["--repository", repository["repo_url"]])
+    for repository in repositories:
```
jin: collapse into the same loop above

specs.bzl
```diff
@@ -246,14 +246,21 @@ def _artifact_to_coord(artifact):
...
+
+    (_, remainder) = repository_spec["repo_url"].split("//")
+    hostname = remainder.split("/")[0]
+
+    credentials = repository_spec["credentials"]
 
-    credentials = repository_spec.get("credentials")
-    auth_string = (credentials["user"] + ":" + credentials["password"] + "@") if credentials != None else ""
-    return protocol + "//" + auth_string + remainder
+    return hostname + " " + credentials["user"] + ":" + credentials["password"]
```
jin: Please add a comment linking to https://get-coursier.io/docs/other-credentials.html#inline to have the context for this format

coursier.bzl
```diff
@@ -510,7 +510,10 @@ def _coursier_fetch_impl(repository_ctx):
...
+        if "credentials" in repository:
+            cmd.extend(["--credentials", utils.repo_credentials(repository)])
```
jin: Do you need quotes around `util.repo_credentials(repository)`?  https://get-coursier.io/docs/other-credentials.html#inline suggests so.

TheMarex: The linked article shows usage through bash, we are passing it to an execution context directly here. A string here should be passed as a single argument to `coursier`, so there is no need to escape it.

TheMarex:
Can someone please take a look again, I addressed the comments. Thanks!

jin:
Thanks! Please address the failing tests.

jin:
LGTM, thanks. Verified manually with a local private artifactory instance.

Issue #136, opened 2019-05-02
Use versionless name in jvm_import and don't use alias


borkaehw:
I am using [rules_scala](https://github.com/higherkindness/rules_scala) and I have a target using alias for external dependencies. 
```
scala_library(
    name = "my_project",
    srcs = glob(["src/main/**/*.scala"]),
    visibility = ["//visibility:public"],
    deps = [
        "@maven//:com_box_box_java_sdk",
        "@maven//:com_eclipsesource_minimal_json_minimal_json",
        "@maven//:com_amazonaws_amazon_kinesis_connectors",
    ],
)
```
In [rules_scala](https://github.com/higherkindness/rules_scala), there is a deps checker which will log a buildozer command to fix the deps. In my case for example, `@maven//:com_amazonaws_amazon_kinesis_connectors` is not used, so a command like
```
buildozer 'remove deps @mvn//:com_amazonaws_amazon_kinesis_connectors_1_3_0' //my_project:my_project
```
will be logged. However, the resolved deps name is used rather than the alias name, so buildozer fails to fix `BUILD.bazel`.

I investigated how deps checker works but I realized there is no easy fix in [rules_scala](https://github.com/higherkindness/rules_scala). I am wondering why do we need alias in `rules_jvm_external` in first place?

Is it possible to have an option to use versionless name in `jvm_import` directly and don't use `alias`? It will solve the problem above easily.

Thanks.

jin:
I've ran into the same issue with buildozer too. Perhaps the ergonomics will be better if we generate the versioned targets as `alias`, and versionless targets as `jvm_imports`, and nothing should break.

@dkelmer wdyt?

borkaehw:
I like this idea and it seems to be a small change.

dkelmer:
Oh interesting. Yeah, sgtm

borkaehw:
@jin would you have time to work on this recently?

dkelmer:
@borkaehw, We probably won't be able to get to this in the immediate future. Would you like to take a stab at it? 

Take a look [here](https://github.com/bazelbuild/rules_jvm_external/blob/master/coursier.bzl#L278) to see where the alias is created

jin:
@borkaehw I have to work on other areas for the moment, please feel free to take this up if you'd like to!

borkaehw:
No problem. I am happy to work on this.

PR #140, opened 2019-05-07
Use versionless jvm_import name

borkaehw:
Attempt to use versionless name in `jvm_import` and use versioned name in `alias`. 

Fix: #136


borkaehw:
I think it could be as simple as it. Let me know if I miss something. Thanks.

PR #141, opened 2019-05-08
Update README to simplify initial WORKSPACE snippet

jin:
Break fetch_sources and fail_on_missing_checksum attributes out into their own
sections.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,11 +51,6 @@ maven_install(
         "https://maven.google.com",
         "https://repo1.maven.org/maven2",
     ],
-    # Fail if a checksum file for the artifact is missing in the repository.
-    # Falls through "SHA-1" and "MD5". Defaults to True.
-    fail_on_missing_checksum = False,
-    # Fetch srcjars. Defaults to False.
-    fetch_sources = True,
 )
 ```
 
@@ -81,7 +76,7 @@ alias(
   actual = "@maven//:junit_junit_4_12",
 )
 
-java_import(
+jvm_import(
   name = "junit_junit_4_12",
   jars = ["@maven//:https/repo1.maven.org/maven2/junit/junit/4.12/junit-4.12.jar"],
   srcjar = "@maven//:https/repo1.maven.org/maven2/junit/junit/4.12/junit-4.12-sources.jar",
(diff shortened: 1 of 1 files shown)


jin:
failure is https://github.com/bazelbuild/continuous-integration/issues/644, not related to this change

README.md
```diff
@@ -110,14 +105,48 @@ example](examples/pom_file_generation/) for more information.
...
+
+```python
+maven_install(
+    artifacts = [
+        # ...
+    ],
+    repositories = [
+        # ...
+    ],
+    fail_on_missing_checksum = False,
```
dkelmer: Isn't the default false?

README.md
```diff
@@ -81,15 +76,15 @@ alias(
   actual = "@maven//:junit_junit_4_12",
 )
 
-java_import(
+jvm_import(
```
dkelmer: While you're here, if you want, these are actually the versionless targets now and the alias has the version on line 74. But if not, lgtm and I can do it in a follow up

README.md
dkelmer: It is not. Shouldn't it be? Oh wait it's maybe coming back to me. Does it not do anything if there isn't actually a checksum?

jin: It fails the build if there isn't a checksum. Why do you think it shouldn't fail if the checksum is missing?

README.md
jin: Done

README.md
dkelmer: Err, maybe I'm misremembering how this works. If `fail_on_missing_checksum` is true, then the build should indeed fail with missing checksums :) 

but I think people should be able to write: 
```
maven_install(
    artifacts = [...],
    repositories = [...]
)
``` 

and that's it. That requires the default of `fail_on_missing_checksum` to be False. If it's not, then in order to not specify checksums for artifacts, users must write:

```
maven_install(
    artifacts = [...],
    repositories = [...],
    fail_on_missing_checksum = False,
)
``` 
which if the second snippet I wrote is the minimum, then the README should actually keep the `fail_on_misisng_checksum` attribute in this "default" snippet

jin: Do you mean you'd like to change the default for fail_on_missing_checksum from False to True in this PR?

dkelmer: No :)
But I'm saying that 
1. maybe we should consider the default and 
2. since the default is True right now, have the example in the README on line 54ish set `fail_on_missing_checksum=False` because otherwise it's not a working example

PR #149, opened 2019-05-20
Add tests for maven_install.excluded_artifacts

jin:
This PR adds an JDK-11 compatible integration test for globally excluded artifacts using `maven_install.excluded_artifacts`.


jin:
Can I get a review on this please?

PR #150, opened 2019-05-20
Add an environment variable, COURSIER_CACHE, to set the location of the unsafe shared cache

jin:
This is really a convenience thing: sometimes users don't want to use ~/.cache/coursier (for Linux), but the path to the shared cache should not be a WORKSPACE-level configuration. The paths tend to differ between users and environment, so we use an environment variable instead.

The path needs to be absolute because Bazel parses relative paths rooted at output_base/external/@repo, and we don't want users to be able to change the structure of that directory.

### DIFF
--- a/coursier.bzl
+++ b/coursier.bzl
@@ -517,8 +517,37 @@ def _coursier_fetch_impl(repository_ctx):
         cmd.extend(["--repository", utils.repo_url(repository)])
     for a in excluded_artifacts:
         cmd.extend(["--exclude", ":".join([a["group"], a["artifact"]])])
-    if not repository_ctx.attr.use_unsafe_shared_cache:
-        cmd.extend(["--cache", "v1"])  # Download into $output_base/external/$maven_repo_name/v1
+
+    # Coursier supports downloading the artifact into a custom directory on disk. This improves
+    # resolution performance since artifacts can be aggressively cached.
+    #
+    # However, it's not a safe thing to do as Bazel has no knowledge about modifications made
+    # to this directory.
+    #
+    # We support custom absolute cache locations using the COURSIER_CACHE environment variable.
+    # This should not be a WORKSPACE configuration because it varies from user to user. Locations
+    # tend to differ between CI and dev machines, for example.
+    #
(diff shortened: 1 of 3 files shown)


coursier.bzl
```diff
@@ -517,8 +517,37 @@ def _coursier_fetch_impl(repository_ctx):
...
+    # This should not be a WORKSPACE configuration because it varies from user to user. Locations
+    # tend to differ between CI and dev machines, for example.
+    #
+    # We do not support relative paths because they will be relative to Bazel's output base and
+    # not the user's WORKSPACE directory.
+    if repository_ctx.attr.use_unsafe_shared_cache:
+        shared_cache_path = repository_ctx.os.environ.get("COURSIER_CACHE")
+        if shared_cache_path:
+            absolute_shared_cache_path = repository_ctx.path(shared_cache_path)
+            if str(absolute_shared_cache_path).find(str(repository_ctx.path(""))) != -1:
```
aehlig: I don't think this is a valid test to see if the provided path was absolute. An absolute path might well contain the absolute path of the working directory as a substring, e.g., if some form prefix pattern is used; consider an invocation like `env COURSIER_CACHE=/var/cache/home/ci/work/external/jvm_external bazel build --output_base=/home/ci/work`

jin: We want to check if the path is relative, not absolute. A relative path (COURSIER_CACHE=a/b/c bazel build //...) that's resolved using repository_ctx.path() will always contain the same path prefix with the empty repository_ctx.path(""). 

The purpose is to prevent users from using output_base as a shared cache location. Do you have other ideas on how we can achieve this?

aehlig: >  [...] will always contain the same path prefix with the empty repository_ctx.path("").

The important word is "prefix"; the `.find() != -1` check checks for substring.

> The purpose is to prevent users from using output_base as a shared cache location.

That's also something different than what your error message says: an absolute path to `output_base` is still an absolute path; but according to that explanation you want to fail in this case as well. So the error message has to be changed as well.

jin:
Thanks @aehlig - I'll go back to the drawing board on this PR, and maybe we can use the new managed directories feature with this.

PR #153, opened 2019-05-24
Add .bazelignore so that we can run bazel test //...

jin:


.bazelci/presubmit.yml
```diff
@@ -2,22 +2,22 @@
 platforms:
   ubuntu1604:
     test_targets:
-      - "//tests/..."
+      - "//..."
```
dkelmer: Is this clearer than //tests/...? 

When it's //tests/... then you can tell exactly what's getting tested just from the contents of the presubmit.yml file but with this change you need to know to reference another file to find out the actual set of tests being run

jin: IMO presubmit.yml is just for CI consumption, not users. We can see the tests pretty easily from the Buildkite UI or with `query`.

PR #156, opened 2019-05-30
Buildifier and CI config cleanup

jin:
https://github.com/bazelbuild/continuous-integration/blob/master/buildkite/README.md#legacy-format

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -43,7 +43,10 @@ maven_install(
             artifact = "guava",
             version = "27.0-jre",
             exclusions = [
-                maven.exclusion(group = "org.codehaus.mojo", artifact = "animal-sniffer-annotations"),
+                maven.exclusion(
+                    group = "org.codehaus.mojo",
+                    artifact = "animal-sniffer-annotations",
+                ),
                 "com.google.j2objc:j2objc-annotations",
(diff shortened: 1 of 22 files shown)


WORKSPACE
```diff
@@ -2,11 +2,6 @@ workspace(name = "rules_jvm_external")
 
 android_sdk_repository(name = "androidsdk")
 
-local_repository(
```
dkelmer: was this unused?

defs.bzl
```diff
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 load("@rules_jvm_external//:coursier.bzl", "coursier_fetch")
-load("@rules_jvm_external//:specs.bzl", "maven", "parse", "json")
```
dkelmer: was maven unused? I'm very impressed buildifier knows how to check for that

WORKSPACE
jin: Yes, unused

jin: Buildifier says this name was defined earlier by the workspace() declaration

defs.bzl
jin: Yep! Caught by buildifier

Issue #157, opened 2019-05-31
Breaks with Python toolchains (Bazel 0.27)


brandjon:
Bazel 0.27 flips [`--incompatible_use_python_toolchains`](https://github.com/bazelbuild/bazel/issues/7899), which [breaks](https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/1008#57a9d0b1-81bd-4134-902f-3b1862e69a74) this repo. Note that this pipeline was [already failing](https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/1009#0bf6703c-00da-436e-b4f2-5e2ff4b90ba5) due to a different error.

This flag causes Bazel to use the version of Python requested at analysis time, rather than whatever the system `python` command happens to be. You likely need to add `python_version = "PY2"` to your Python targets, and/or enable `--force_host_python=PY2` in your bazelrc (if any of these are used in the host configuration).

PR #158, opened 2019-05-31
Fix incompatible change with python toolchains

jin:
Passes with `$ bazel test //... --host_force_python=PY2 --incompatible_use_python_toolchains`

Fixes https://github.com/bazelbuild/rules_jvm_external/issues/157

### DIFF
--- a/.bazelci/examples.yml
+++ b/.bazelci/examples.yml
@@ -23,6 +23,8 @@ tasks:
     platform: ubuntu1804
     working_directory: examples/android_instrumentation_test
     test_flags:
+      # Compatibility with --incompatible_use_python_toolchains
+      # https://github.com/bazelbuild/bazel/issues/7899
       - "--host_force_python=PY2"
     test_targets:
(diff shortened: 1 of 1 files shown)


.bazelci/examples.yml
```diff
@@ -22,6 +22,10 @@ tasks:
     name: "Android instrumentation test example"
     platform: ubuntu1804
     working_directory: examples/android_instrumentation_test
+    test_flags:
+      # Compatibility with --incompatible_use_python_toolchains
+      # https://github.com/bazelbuild/bazel/issues/7899
+      - "--host_force_python=PY2"
```
laurentlb: Shall we put it in a .bazelrc file?

laurentlb:
(fyi, @brandjon)

.bazelci/examples.yml
jin: I'm upstreaming a change to the broken dependency.

PR #160, opened 2019-06-04
Add generate_compat_repositories attribute to maven_install for top level repository aliases and cross-workspace dependencies

### LINKED ISSUE #85: Top-level repository aliases and cross-workspace dependencies
Quite a few projects in the Bazel ecosystem expect their java dependencies to be individually imported as repository rules in their own right, for example [grpc-java](https://github.com/grpc/grpc-java/blob/master/repositories.bzl) expects to find the top-level `@com_google_guava_guava//jar`, which is created via the built-in rule `jvm_maven_import_external` (from `jvm.bzl`).

In my current projects, I tend to have a (generated) bunch of `jvm_maven_import_external`/`java_import_external` targets which supply both my own targets as well as third-party ones like grpc-java. That is, I don't use the `grpc_java_repositories` function to import those dependencies; I just ensure that I provide a compatible version in the expected place.

With `rules_jvm_external` the paradigm is slightly different. Resolution is now to many targets within a single repository rule, i.e. `@com_google_guava_guava//jar` becomes `@maven//:com_google_guava_guava`. This of course means that the loaded grpc-java rules can't find their dependencies when I switch to these new rules. If the `grpc_java_repositories` function is called, it will create a second repository rule for things which may already be available in the `@maven` workspace, and may cause classpath collisions downstream later.

I can see clear advantages to the new rules, particularly that different versions of the same artifact can now be resolved in different repositories (`@maven//:foo_bar` != `@othermaven//:foo_bar`). Downstream classpath collisions are clearly still possible there, so to some extent this grpc-java issue is just the natural outcome of an assumption of top-level repository safety/unity.

But, it might be useful to generate these `@foo_bar` top-level repository aliases for consistency with the old-style patterns. At least until there is a more obvious solution for reuse of Java deps across a full workspace including external dependency workspaces. Maybe this exists already! I'd be happy to find a solution to this problem :)

The built-in Bazel options such as [shadowing](https://docs.bazel.build/versions/master/external.html#shadowing-dependencies) and binding are only apparently functional for an entire workspace, so can't be used to generically alias individual targets in either the source or the target workspaces, i.e. it's not possible to redirect the `@foo_bar//jar` generated by `java_import_external`, nor target the `@maven//:foo_bar` for a single artifact.

A couple of options:

1. Add a parameter to the `maven_install` rule to allow generation of an additional `repositories.bzl` file, containing a function which can be loaded to create new repository rules for some/all artifacts in the closure, mirroring the format generated by `java_import_external`.
2. Make it easier for downstream workspaces to do the above, by exposing the generated `@maven//:dep-tree.json` file.

I have a hacked-up version of the first one, but it feels a bit ugly so far. Is there any plan for a migration path between `java_import_external` and `maven_install`? If not, would a PR for either option above be useful?

jin:
This PR adds a new attribute, `generate_compat_repositories`, to `maven_install`. The default is `False`. The motivation for this change is https://github.com/bazelbuild/rules_jvm_external/issues/85. (Fixes #85)

With this, we can refer to `@maven//:junit_junit` as `@junit_junit//jar`, where the latter is the most commonly used naming convention that is popularized by rules like `maven_jar` and `java_import_external`. This enables project maintainers to incrementall migrate from `maven_jar` to `rules_jvm_external` without a single change in their BUILD files. It also allows cross-external repository sharing of artifacts.

If enabled, `maven_install` generates an additional `compat.bzl` file in the `@maven` (or your custom) external repository. This file exposes one symbol, `compat_repositories`, to be loaded after `maven_install`, like this:

```python
maven_install(
    # ...
    generate_compat_repositories = True,
)

load("@maven//:compat.bzl", "compat_repositories")
compat_repositories()
```

`compat_repositories` expands into a list of `compat_repository` declarations, one for each `JAR` artifact in the transitive closure of the resolved dependency tree. For example:

```
$ cat compat.bzl 
load("@maven//:compat_repository.bzl", "compat_repository")
def compat_repositories():
    compat_repository(name = "android_arch_core_common")
    compat_repository(name = "android_arch_lifecycle_common")
    compat_repository(name = "androidx_annotation_annotation")
    compat_repository(name = "com_android_support_support_annotations")
    compat_repository(name = "com_google_code_findbugs_jsr305")
    compat_repository(name = "com_squareup_javawriter")
    compat_repository(name = "javax_inject_javax_inject")
    compat_repository(name = "junit_junit")
    compat_repository(name = "net_sf_kxml_kxml2")
    compat_repository(name = "org_ccil_cowan_tagsoup_tagsoup")
    compat_repository(name = "org_hamcrest_hamcrest_core")
    compat_repository(name = "org_hamcrest_hamcrest_integration")
    compat_repository(name = "org_hamcrest_hamcrest_library")
```

Each `compat_repository` creates a new external repository in `output_base/external`. In the case of junit, the directory is `output_base/external/junit_junit`, and the contents are as follows:

```
jingwen@jingwen:~/.cache/bazel/_bazel_jingwen/f1b46d82f4f66241b4c332b551d2226c/external/junit_junit
$ tree
.
├── jar
│   └── BUILD
└── WORKSPACE
```

The `WORKSPACE` is an empty file. The `BUILD` file contains an alias to the `@maven//:junit_junit` target:

```
$ cat jar/BUILD 
alias(name = "jar", actual = "@maven//:junit_junit")
```

Using this, we can refer to the target as `@junit_junit//jar:jar`, or `@junit_junit//jar` for short.

TODO 

- [x] Tests
- [x] Documentation

### DIFF
--- a/coursier.bzl
+++ b/coursier.bzl
@@ -584,7 +584,7 @@ def _coursier_fetch_impl(repository_ctx):
             compat_repositories_bzl.append("    )")
 
         repository_ctx.file(
-            "repositories.bzl",
+            "compat.bzl",
             "\n".join(compat_repositories_bzl),
             False,  # not executable
         )
--- a/coursier.bzl
+++ b/coursier.bzl
@@ -128,7 +128,7 @@ def _generate_imports(repository_ctx, dep_tree, neverlink_artifacts = {}):
     # A list of versionless target labels for jar artifacts. This is used for
     # generating a compatibility layer for repositories. For example, if we generate
     # @maven//:junit_junit, we also generate @junit_junit//jar as an alias to it.
-    jar_imports = []
+    jar_versionless_target_labels = []
 
     # First collect a map of target_label to their srcjar relative paths, and symlink the srcjars if needed.
     # We will use this map later while generating target declaration strings with the "srcjar" attr.
@@ -180,7 +180,7 @@ def _generate_imports(repository_ctx, dep_tree, neverlink_artifacts = {}):
                 # Kotlin compile interface JARs to be incorrect. We replace java_import
                 # with a simple jvm_import Starlark rule that skips ijar.
                 target_import_string = ["jvm_import("]
-                jar_imports.append(target_label)
+                jar_versionless_target_labels.append(target_label)
             elif packaging == "aar":
                 target_import_string = ["aar_import("]
             else:
@@ -367,7 +367,7 @@ Parsed data: {parsed_artifact}""".format(
             )
             fail(error_message)
 
-    return ("\n".join(all_imports), jar_imports)
+    return ("\n".join(all_imports), jar_versionless_target_labels)
 
 def _deduplicate_list(items):
     seen_items = {}
@@ -546,7 +546,7 @@ def _coursier_fetch_impl(repository_ctx):
 
     neverlink_artifacts = {a["group"] + ":" + a["artifact"]: True for a in artifacts if a.get("neverlink", False)}
     repository_ctx.report_progress("Generating BUILD targets..")
-    (generated_imports, jar_imports) = _generate_imports(
+    (generated_imports, jar_versionless_target_labels) = _generate_imports(
         repository_ctx = repository_ctx,
         dep_tree = dep_tree,
         neverlink_artifacts = neverlink_artifacts,
@@ -555,6 +555,13 @@ def _coursier_fetch_impl(repository_ctx):
     repository_ctx.template(
         "jvm_import.bzl",
         repository_ctx.attr._jvm_import,
+        substitutions = {},
+        executable = False,  # not executable
+    )
+
+    repository_ctx.template(
+        "compat_repository.bzl",
+        repository_ctx.attr._compat_repository,
         substitutions = {},
         executable = False,  # not executable
     )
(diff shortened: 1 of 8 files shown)


jin:
(fyi @dslomov, this can help with maven_jar deprecation)

Globegitter:
@jin that is a little bit off-topic but I wonder if it makes sense to write some "best practices" on how to expose external dependencies to not always having to think how to import an external dep when switching language. I feel nearly every rule does it differently, e.g. rules_nodejs and rules_pip also expose a single repository like the default here but they expose it without the `:`, so `@npm//package` or `@pip//package` respectively. Or rules_go also does it per-repository and exposes things as `@com_github_spf13_cobra//:go_default_library`. Though with gazelle one rarely has to write things per hand there. And I have seen other rules that expose a function, so one imports it in each `BUILD` file and then can write something like `pip('package')` or similar.

I personally like the function exposing way the least and I do not really have a preference on one repository or multiple repositories but I do prefer not having to write the `:`, it does save one character and somehow does feel more "natural".

I am also happy to move this conversation to an issue on bazel itself or on Slack to not sidetrack things any further here.

coursier.bzl
```diff
@@ -562,6 +568,31 @@ def _coursier_fetch_impl(repository_ctx):
...
+                executable = False,
+            )
+            repository_ctx.file(
+                versionless_target_label + "/jar/BUILD",
+                "alias(name = \"jar\", actual = \"@maven//:%s\", visibility = [\"//visibility:public\"])" % versionless_target_label,
+                executable = False,
+            )
+            compat_repositories_bzl.append("    native.local_repository(")
+            compat_repositories_bzl.append("        name = \"%s\"," % versionless_target_label)
+            compat_repositories_bzl.append("        path = \"%s\"," % repository_ctx.path(versionless_target_label))
```
dslomov: Instead of using local_repository in this hacky way, please add a new repository rule implementation of which is just generating the right BUILD file

jin:
@Globegitter thanks for raising this discussion. I personally haven't thought much about setting some kind of convention across these rules, and I think it's a good idea to do that in order to reduce cognitive load. 

This PR set ups the framework for `@foo_bar//jar`, and it's trivial to extend that by also creating `@maven//foo_bar:foo_bar`/`@maven//foo_bar`, on top of `@maven//:foo_bar`. However, I'm on the fence about adding yet another alias that's very similar to `@maven//:foo_bar`. `@foo_bar//jar` has the additional functionality of better interop with the ecosystem. 

Let's fork this discussion onto a new GitHub issue?

PR #167, opened 2019-06-14
Expose the Starlark source of truth for the dependency tree

jin:
This PR exposes `@maven//:dependency_tree.bzl`, a file containing the Starlark source of truth dictionary containing the full dependency tree information generated by Coursier. This is also the source of truth for generating the `@maven//:BUILD` file.

Since this is a full dependency tree, it can be used for further analysis in `.bzl` or `BUILD` files, like the golden tests in `examples/artifact_resolution_analysis`.


jin:
Extending / consolidating this work in #169.

PR #169, opened 2019-06-17
Support pinning artifacts with checksums and downloading artifacts with Bazel

### LINKED ISSUE #125: Support artifact checksum pinning
From Bazel slack: https://bazelbuild.slack.com/archives/CE6AZNQP7/p1556024364000500

~Coursier supports it: https://github.com/coursier/coursier/pull/625~

### LINKED ISSUE #146: Make bazel download files
Coursier optimistically downloads every single jar file required when starting a build. When the number of dependencies is large, this can cause a significant delay before anything starts being built, particularly since `jar` files are significantly larger than `pom` files (on the whole)

If, instead, `maven_install` output an `http_file` for each jar or source jar required, builds could start a lot sooner and we could ensure that all downloads happen through Bazel's own mechanisms.

There doesn't appear to be a command line flag to make this happen automatically, but once we have resolved all the dependencies, the URLs to download from tend to be a simple transformation of the maven coordinates.

### LINKED ISSUE #170: Errors with "+" character in artifact coordinates
e.g. `"ch.epfl.scala:compiler-interface:1.3.0-M4+20-c8a2f9bd"` with https://github.com/bazelbuild/rules_jvm_external/pull/169

```
ERROR: /usr/local/google/home/jingwen/.cache/bazel/_bazel_jingwen/d8d6ce4c4882b95e8ff0a54790fbf3bc/external/annex/BUILD:50:1: @annex//:ch_epfl_scala_compiler_interface_jar_sources_1_1_7+62_0f4ad9d5_extension: invalid label '@ch_epfl_scala_compiler_interface_jar_sources_1_1_7+62_0f4ad9d5//file' in element 0 of attribute 'srcs' in 'genrule' rule: invalid repository name '@ch_epfl_scala_compiler_interface_jar_sources_1_1_7+62_0f4ad9d5': workspace names may contain only A-Z, a-z, 0-9, '-', '_' and '.'
```

jin:
This PR adds a new `maven_install_json` attribute that accepts a label to a JSON file, `maven_install.json`. This file contains the [pinned](https://before-you-ship.18f.gov/infrastructure/pinning-dependencies/) / locked state of a `maven_install` fetch. 

This feature improves artifact resolution and fetch speed, since it uses Bazel's download mechanism which caches files on their sha256 checksums. It also improves resiliency and security by encoding the sha256 checksums and original artifact urls into the JSON file.

Fixes #125 
Fixes #146 
Fixes #170 

To get started with pinning artifacts, run the following command:

```
$ bazel run @maven//:pin
```

This generates a `maven_install.json` (or `your_maven_install_name_install.json`) in the root of your Bazel workspace.

Then, specify `maven_install_json` in `maven_install` and load `pinned_maven_install` from `@maven//:defs.bzl`:

```python
maven_install(
    # artifacts, repositories, ...
    maven_install_json = "//:maven_install.json",
)

load("@maven//:defs.bzl", "pinned_maven_install")
pinned_maven_install()
```

Whenever you make a change to the list of `artifacts` or `repositories` and want to update `maven_install.json`, run this command to re-pin the unpinned `@maven` repository:

```
$ bazel run @unpinned_maven//:pin
```

By specifying `maven_install_json`, an additional `@unpinned_maven` (or `unpinned_<your_maven_install_name>`) repo will be created. For example, if your `maven_install` is named `@foo`, `@unpinned_foo` will be created.

The contents of `maven_install.json` will look something like this:

```json
$ cat maven_install.json
{
    "dependency_tree": {
        "conflict_resolution": {},
        "dependencies": [
            {
                "coord": "com.google.code.findbugs:jsr305:3.0.2",
                "dependencies": [],
                "file": "v1/https/repo1.maven.org/maven2/com/google/code/findbugs/jsr305/3.0.2/jsr305-3.0.2.jar",
                "sha256": "766ad2a0783f2687962c8ad74ceecc38a28b9f72a2d085ee438b7813e928d0c7",
                "url": "https://repo1.maven.org/maven2/com/google/code/findbugs/jsr305/3.0.2/jsr305-3.0.2.jar"
            },
            {
                "coord": "com.google.errorprone:error_prone_annotations:2.2.0",
                "dependencies": [],
                "file": "v1/https/repo1.maven.org/maven2/com/google/errorprone/error_prone_annotations/2.2.0/error_prone_annotations-2.2.0.jar",
                "sha256": "6ebd22ca1b9d8ec06d41de8d64e0596981d9607b42035f9ed374f9de271a481a",
                "url": "https://repo1.maven.org/maven2/com/google/errorprone/error_prone_annotations/2.2.0/error_prone_annotations-2.2.0.jar"
            },
            {
                "coord": "com.google.guava:failureaccess:1.0",
                "dependencies": [],
                "file": "v1/https/repo1.maven.org/maven2/com/google/guava/failureaccess/1.0/failureaccess-1.0.jar",
                "sha256": "d084bef9cd07a8537a1753e4850a69b7e8bab1d1e22e9f3a1e4826309a7a2336",
                "url": "https://repo1.maven.org/maven2/com/google/guava/failureaccess/1.0/failureaccess-1.0.jar"
            },
            {
                "coord": "com.google.guava:guava:27.0-android",
                "dependencies": [
                    "com.google.guava:listenablefuture:9999.0-empty-to-avoid-conflict-with-guava",
                    "com.google.code.findbugs:jsr305:3.0.2",
                    "com.google.guava:failureaccess:1.0",
                    "org.checkerframework:checker-compat-qual:2.5.2",
                    "org.codehaus.mojo:animal-sniffer-annotations:1.17",
                    "com.google.j2objc:j2objc-annotations:1.1",
                    "com.google.errorprone:error_prone_annotations:2.2.0"
                ],
                "file": "v1/https/repo1.maven.org/maven2/com/google/guava/guava/27.0-android/guava-27.0-android.jar",
                "sha256": "9fcd95a6d27ce9309b68cfb92369192ae13498d5104a3367e6c64e3879215324",
                "url": "https://repo1.maven.org/maven2/com/google/guava/guava/27.0-android/guava-27.0-android.jar"
            },
            {
                "coord": "com.google.guava:listenablefuture:9999.0-empty-to-avoid-conflict-with-guava",
                "dependencies": [],
                "file": "v1/https/repo1.maven.org/maven2/com/google/guava/listenablefuture/9999.0-empty-to-avoid-conflict-with-guava/listenablefuture-9999.0-empty-to-avoid-conflict-with-guava.jar",
                "sha256": "b372a037d4230aa57fbeffdef30fd6123f9c0c2db85d0aced00c91b974f33f99",
                "url": "https://repo1.maven.org/maven2/com/google/guava/listenablefuture/9999.0-empty-to-avoid-conflict-with-guava/listenablefuture-9999.0-empty-to-avoid-conflict-with-guava.jar"
            },
            {
                "coord": "com.google.j2objc:j2objc-annotations:1.1",
                "dependencies": [],
                "file": "v1/https/repo1.maven.org/maven2/com/google/j2objc/j2objc-annotations/1.1/j2objc-annotations-1.1.jar",
                "sha256": "2994a7eb78f2710bd3d3bfb639b2c94e219cedac0d4d084d516e78c16dddecf6",
                "url": "https://repo1.maven.org/maven2/com/google/j2objc/j2objc-annotations/1.1/j2objc-annotations-1.1.jar"
            },
            {
                "coord": "org.checkerframework:checker-compat-qual:2.5.2",
                "dependencies": [],
                "file": "v1/https/repo1.maven.org/maven2/org/checkerframework/checker-compat-qual/2.5.2/checker-compat-qual-2.5.2.jar",
                "sha256": "d7291cebf5e158d169807ae49d4b16ff672986f0c6d803e5f207c40cb61ef982",
                "url": "https://repo1.maven.org/maven2/org/checkerframework/checker-compat-qual/2.5.2/checker-compat-qual-2.5.2.jar"
            },
            {
                "coord": "org.codehaus.mojo:animal-sniffer-annotations:1.17",
                "dependencies": [],
                "file": "v1/https/repo1.maven.org/maven2/org/codehaus/mojo/animal-sniffer-annotations/1.17/animal-sniffer-annotations-1.17.jar",
                "sha256": "92654f493ecfec52082e76354f0ebf87648dc3d5cec2e3c3cdb947c016747a53",
                "url": "https://repo1.maven.org/maven2/org/codehaus/mojo/animal-sniffer-annotations/1.17/animal-sniffer-annotations-1.17.jar"
            }
        ],
        "version": "0.1.0"
    }
}
```

Since all artifacts are checksummed with sha256, it means that **fully offline builds are possible even with `bazel clean --expunge`** after performing one `bazel fetch @maven//...`.

Whenever a change is made to the list of `artifacts` or `repositories`, you will need to rerun 

```
$ bazel run @unpinned_maven//:pin
```

to update the pinned artifacts.

This change adds a dependency on system Python for two things: platform independent sha256 tool from `@bazel_tools` and pretty printing of `maven_install.json`.


Globegitter:
Awesome, this is a really good step into the right direction imo.

Would it be also possible to expose a new `.bzl` file with independent external repositories as discussed in #146 rather than calling `ctx.download`?

Also I might try and look at adding the json file capability to `coursier resolve` as it would be a nice optimisation.

jin:
> Would it be also possible to expose a new .bzl file with independent external repositories as discussed in #146 rather than calling ctx.download?

@Globegitter is this use case covered by https://github.com/bazelbuild/rules_jvm_external#repository-aliases already?

README.md
```diff
@@ -393,6 +393,95 @@ load("@maven//:compat.bzl", "compat_repositories")
...
+    "file": # Relative path to the artifact file.
+    "dependencies": [
+        # A list of coordinates for the dependencies of this artifact.
+    ]
+  },
+  # Other artifacts..
+]
+```
+
+To get the absolute path to the artifact, prefix the `file` value with the value of `bazel info $(output_base)/external/maven`.
```
borkaehw: Sorry I know it's WIP, just a reminder here.
I suppose it should be `$(bazel info output_base)/external/maven`

jin: Oo, good catch.

Globegitter:
@jin that only changes the way one can import deps but because they still rely on the same external repository all deps will be downloaded the first time. The use case discussed in #146 is just possible by having independent repository rules that call ctx.download.

borkaehw:
I have tried it out and it's heading to the direction we hope for, thanks for taking care of these two issues. Few concerns I have so far:

1. Coursier seems to resolve colons in URLs to encoding form in `dep-tree.json`. For instance, `http://repo.foobar.com:8000` will become `v1/http/repo.foobar.com%3A8000`, which causes downloading error when it tries to convert it back to URL.
2. It seems to take twice as long for fetching and resolving. Are both Coursier and Bazel trying to do the downloading? 
3. `pinned_maven_install.json` is currently generated in `external/maven/`. Is there a plan to have it generated in (or copied to) main workspace to be checked in?

Thanks.

jin:
> @jin that only changes the way one can import deps but because they still rely on the same single external repository so all deps will be downloaded the first time. The use case discussed in #146, of just downloading the deps actually needed for the current build, is just possible by having independent repository rules that call ctx.download for one external dependency each.

@Globegitter This is true. Let me play around with this idea instead of ctx.download.

> Coursier seems to resolve colons in URLs to encoding form in dep-tree.json. For instance, http://repo.foobar.com:8000 will become v1/http/repo.foobar.com%3A8000, which causes downloading error when it tries to convert it back to URL.

I suppose we can do a simple string replacement for colons when converting back into URLs.

> It seems to take twice as long for fetching and resolving. Are both Coursier and Bazel trying to do the downloading?

Yes it's doing that, it seems that `ctx.download` downloads eagerly even if the statement is not in an executed branch. Using the `http_file` approach will (probably) bypass this issue.

> pinned_maven_install.json is currently generated in external/maven/. Is there a plan to have it generated in (or copied to) main workspace to be checked in?

This is what `$ bazel run @maven//:pin` is for. The original idea is to have it copied into the main workspace automatically, but I cannot figure out a reliable way to get the path to the source workspace without passing arguments into `$ bazel run @maven//:pin`.

jin:
What are your opinions about the API?

Option 1:

```
maven_install(
     pinned_maven_install = "//:pinned.json",
)
```

Option 2

```
maven_install(
    # usual stuff
)

pinned_maven_install(
     name = "pinned_maven",
     maven_install_json = "//:maven_install.json",
)
```

I like the second one more, because it makes explicit that the workspace depend on `@pinned_maven//:...` artifacts instead of regular `@maven//`. It also allows users who don't rely on pinning to use regular `maven_install`. 

The first option feels like it's cramming too much into `maven_install`.

jin:
@borkaehw @Globegitter I've switched `ctx.download` to `http_file` in the latest commit, and hooked up the Scala example that uses this. PTAL, lmk what you think.

Globegitter:
@jin Awesome! I will take a look at it today. Also about figuring out how to copy things to the source workspace bazel does expose the `BUILD_WORKSPACE_DIRECTORY`  env var on bazel run. Would that help?

jin:
@Globegitter TIL! That makes things a lot simpler. Thanks.

coursier.bzl
```diff
@@ -110,6 +112,17 @@ def _get_reverse_deps(coord, dep_tree):
                 reverse_deps.append(maybe_rdep)
     return reverse_deps
 
+def _genrule_copy_artifact_from_http_file(artifact):
+    http_file_repository = _escape(artifact["coord"])
+    return "\n".join([
+        "genrule(",
+        "     name = \"%s_extension\"," % http_file_repository,
+        "     srcs = [\"@%s//file:downloaded\"]," % http_file_repository,
```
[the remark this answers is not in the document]
jin: You're right, done.

coursier.bzl
```diff
@@ -575,6 +667,22 @@ def _coursier_fetch_impl(repository_ctx):
         False,  # not executable
     )
 
+    # If this is an unpinned repository, expose the script to let users pin the
+    # state of the fetch in `<workspace_root>/maven_install.json`.
+    #
+    # $ bazel run @unpinned_maven//:pin
+    #
+    if not repository_ctx.attr.maven_install_json:
```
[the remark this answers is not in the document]
jin: Yes. `maven_install.json` must only be generated from `@unpinned_maven`, since `@unpinned_maven` doesn't depend on `maven_install` to prevent cyclic dependencies. 

`@unpinned_maven//:pin` -> generates `maven_install.json` -> generates `@maven//...` targets.

defs.bzl
```diff
@@ -49,6 +54,18 @@ def maven_install(
         generate_compat_repositories = generate_compat_repositories,
     )
 
+    if maven_install_json != None:
+        # Create the repository generated from a maven_install.json file.
+        coursier_fetch(
```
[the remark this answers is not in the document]
jin: The second coursier_fetch generates the `@maven` repository generated from `maven_install.json`.

We don't want the two repositories to have edges between them. This allows users to update the `maven_install()` declaration in the WORKSPACE, run `@unpinned_maven//:pin` to update `maven_install.json`, and `bazel build //...` immediately after with the updated artifacts.

borkaehw:
I just saw your updated PR description, but I can't run `bazel run @unpinned_maven//:pin` from clean workspace with an empty `maven_install.json`. What do I do wrong?

coursier.bzl
```diff
@@ -486,70 +487,161 @@ def _coursier_fetch_impl(repository_ctx):
...
+            cmd.append("--sources")
+            cmd.append("--default=true")
+        if _is_windows(repository_ctx):
+            # Unfortunately on Windows, coursier crashes while trying to acquire the
+            # cache's .structure.lock file while running in parallel. This does not
+            # happen on *nix.
+            cmd.extend(["--parallel", "1"])
+
+        repository_ctx.report_progress("Resolving and fetching the transitive closure of %s artifact(s).." % len(artifact_coordinates))
+        exec_result = repository_ctx.execute(cmd)
```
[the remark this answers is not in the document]
jin: Coursier's `resolve` command currently doesn't provide a `dep-tree.json` file. Without decoupling resolve and fetch, there's no point in having Bazel download the artifact the second time whenever the external repository is invalidated.

I believe @Globegitter gave this a shot previously by parsing the console tree output. We can iterate on this in a future pull request.

jin:
What's the error message and what does your WORKSPACE look like?

borkaehw:
Ah sorry I see, I have
```
maven_install(
    name = "mvn",
    ...
)
```
so it should be `bazel run @unpinned_mvn//:pin` for me.

Testing things now.

coursier.bzl
```diff
@@ -486,70 +487,161 @@ def _coursier_fetch_impl(repository_ctx):
...
-
-    if repository_ctx.attr.fail_on_missing_checksum:
-        cmd.extend(["--checksum", "SHA-1,MD5"])
+    if repository_ctx.attr.maven_install_json:
+        # Read Coursier state from maven_install.json.
+        repository_ctx.symlink(
+            repository_ctx.path(repository_ctx.attr.maven_install_json),
+            repository_ctx.path("imported_maven_install.json")
+        )
+        dep_tree = json_parse(
```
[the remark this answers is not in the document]
jin: By invalid json, do you mean general malformed json or json with corrupted maven_install.json schema?

General malformed json will be handled by json_parse.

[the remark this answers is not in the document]
borkaehw: If I point to a completely unrelated file like a `.py`.
```
ERROR: error loading package '': Encountered error while reading extension file 'defs.bzl': no such package '@mvn//': Traceback (most recent call last):
	File "/home/borkaehw/.cache/bazel/_bazel_borkaehw/66240b6c93cac99d85d9f7c0eb226b35/external/rules_jvm_external/coursier.bzl", line 496
		json_parse(repository_ctx.read(repository_c...")))
	File "/home/borkaehw/.cache/bazel/_bazel_borkaehw/66240b6c93cac99d85d9f7c0eb226b35/external/rules_jvm_external/third_party/bazel_json/lib/json_parser.bzl", line 667, in json_parse
		fail(("parsing JSON failed:\n%s\n\npa...)))
parsing JSON failed:
invalid action: -1  on input '#'

parser:
{"rejected": True, "rejected_reason": "invalid action: -1  on input '#'", "max_depth": 20, "mode_stack": [0], "top": 0, "state": 0, "state_chars": None, "reduction_stack": [[]], "reduction_hooks": {"entry_key": <function _reduce_literal from @rules_jvm_external//third_party/bazel_json/lib:json_parser.bzl>, "entry_value": <function _reduce_literal from @rules_jvm_external//third_party/bazel_json/lib:json_parser.bzl>, "object": <function _reduce_object from @rules_jvm_external//third_party/bazel_json/lib:json_parser.bzl>, "array": <function _reduce_array from @rules_jvm_external//third_party/bazel_json/lib:json_parser.bzl>}, "tokenizer_hooks": {"string": <function _tokenize_string from @rules_jvm_external//third_party/bazel_json/lib:json_parser.bzl>, "integer": <function _tokenize_int from @rules_jvm_external//third_party/bazel_json/lib:json_parser.bzl>, "null": <function _tokenize_null from @rules_jvm_external//third_party/bazel_json/lib:json_parser.bzl>, "true": <function _tokenize_true from @rules_jvm_external//third_party/bazel_json/lib:json_parser.bzl>, "false": <function _tokenize_false from @rules_jvm_external//third_party/bazel_json/lib:json_parser.bzl>}, "hacky_treatment_unknown_chars_as_etc": True, "_DEBUG": False}
```
I feel like the error message is too long to see what's wrong easily.

If it's a corrupted `maven_install.json`. It tosses the entire corrupted file content, which in my case has 20k lines. I can't even see the error message.

It would be nice to just let me know it's the json issue.

README.md
```diff
@@ -101,7 +101,72 @@ example](examples/pom_file_generation/) for more information.
...
+
+```python
+maven_install(
+    # artifacts, repositories, ...
+    maven_install_json = "//:maven_install.json",
+)
+```
+
+By specifying `maven_install_json`, an additional `unpinned_maven` (or
+`unpinned_*`) repo will be created. `unpinned_maven` contains targets fetched by
```
borkaehw: It would be better to clarify what does `unpinned_*` mean. Something like `unpinned_<your_maven_install_name>` would help people avoid my mistake. :smile:

README.md
```diff
@@ -101,7 +101,72 @@ example](examples/pom_file_generation/) for more information.
...
+maven_install(
+    # artifacts, repositories, ...
+    maven_install_json = "//:maven_install.json",
+)
+
+load("@maven//:defs.bzl", "pinned_maven_install")
+pinned_maven_install()
+```
+
+Since all artifacts are store locally in Bazel's cache, it means that **fully
```
borkaehw: Typo: artifacts are "stored".

coursier.bzl
```diff
@@ -284,6 +299,19 @@ def _generate_imports(repository_ctx, dep_tree, neverlink_artifacts = {}):
             versioned_target_alias_label = _escape(_strip_packaging_and_classifier(artifact["coord"]))
             all_imports.append("alias(\n\tname = \"%s\",\n\tactual = \"%s\",\n)" % (versioned_target_alias_label, target_label))
 
+            # 9. If using maven_install.json, use a genrule to copy the file from the http_file
+            # repository into this repository.
+            #
+            # genrule(
+            #     name = "org_hamcrest_hamcrest_library_1_3_extension",
+            #     srcs = ["@org_hamcrest_hamcrest_library_1_3//file:downloaded"],
```
borkaehw: Should be just `"@org_hamcrest_hamcrest_library_1_3//file"`.

coursier.bzl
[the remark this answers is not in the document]
jin: Thanks for the feedback! Just pushed a commit that improved the JSON parsing error messages.

If it's an invalid maven_install.json:

```
ERROR: An error occurred during the fetch of repository 'maven':
   Failed to parse /usr/local/google/home/jingwen/code/rules_jvm_external/examples/scala_akka/maven_install.json. It is not a valid maven_install.json file. Has this file been modified manually?
```

If it's not a JSON file:

```
$ bazel test //...
INFO: Call stack for the definition of repository 'maven' which is a coursier_fetch (rule definition at /usr/local/google/home/jingwen/.cache/bazel/_bazel_jingwen/88893d7c3918775765c5e3ae20d931b1/external/rules_jvm_external/coursier.bzl:714:18):
 - /usr/local/google/home/jingwen/.cache/bazel/_bazel_jingwen/88893d7c3918775765c5e3ae20d931b1/external/rules_jvm_external/defs.bzl:79:9
 - /usr/local/google/home/jingwen/code/rules_jvm_external/examples/scala_akka/WORKSPACE:52:1
DEBUG: /usr/local/google/home/jingwen/.cache/bazel/_bazel_jingwen/88893d7c3918775765c5e3ae20d931b1/external/rules_jvm_external/third_party/bazel_json/lib/json_parser.bzl:672:13: JSON parsing failed. Could not parse the input load("@io_bazel at: 
...
load("@io_bazel_rules_scala//scala:scala.bzl", "scala_binary", "scala_library", "scala_test_suite")

lib_dependencies = [
    "@maven//:com_chuusai_shapeless_2_12",
    "@maven//:com_github_pureconfig
...
ERROR: An error occurred during the fetch of repository 'maven':
   Failed to parse /usr/local/google/home/jingwen/code/rules_jvm_external/examples/scala_akka/BUILD.bazel. Is this file valid JSON?
```

README.md
jin: Reworded this section to make it clearer. PTAL.

defs.bzl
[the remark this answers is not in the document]
borkaehw: It makes sense, thanks for the explanation.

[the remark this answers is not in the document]
jin: I added my response as comment in the code.

README.md
borkaehw: It looks great, thanks.

borkaehw:
Question not comment:
In `def maven_install`, how do you prevent the first `coursier_fetch` from being executed every time? There seems no checking condition, but it's only executed when we run `bazel run @maven//:pin`.

jin:
If nothing in your build depends on `@unpinned_maven//...`, it'll not be executed.

borkaehw:
This PR looks great to me. One possible improvement is to have Bazel generate the initial `maven_install.json` and give an option to specify a custom name, but it will only happen once when people start using it. We can address it in another PR if it is not an easy change.

Looking forward to the release. Thanks for the hard work.

Globegitter:
@jin it also seems auth support has landed now: https://github.com/bazelbuild/bazel/commit/479cf339d1e3a2638fa3e8af4e8654a46b5def9a - but do you know if there are any plans to make this accessible through `http_file` also or is that just through the `ctx` methods?

jin:
http_file wraps around ctx.download, so it'll need to expose an attribute or mechanism to read auth data.

README.md
```diff
@@ -101,7 +101,75 @@ example](examples/pom_file_generation/) for more information.
...
+    # artifacts, repositories, ...
+    maven_install_json = "//:maven_install.json",
+)
+```
+
+By specifying `maven_install_json`, an additional `@unpinned_maven` (or
+`unpinned_<your_maven_install_name>`) repo will be created. For example, if your
+`maven_install` is named `@foo`, `@unpinned_foo` will be created.
+
+The `@unpinned_maven` repository accompanies the main `@maven` repositor and
```
borkaehw: Typo: accompanies the main `@maven` "repository"

jin: fixed

jin:
@borkaehw I made some changes to improve the first-time experience. You don't need to create an empty `maven_install.json` file now.

borkaehw:
It works great, thanks.

borkaehw:
I am interested in to know how far are we from merging this PR?

jin:
@borkaehw Waiting on a review from either @laurentlb and @aehlig.

coursier.bzl
```diff
@@ -432,23 +461,132 @@ def _get_java_proxy_args(repository_ctx):
...
+
+    dep_tree = maven_install_json_content["dependency_tree"]
+
+    # Create the list of http_file repositories for each of the artifacts
+    # in maven_install.json. This will be loaded additionally like so:
+    #
+    # load("@maven//:defs.bzl", "pinned_maven_install")
+    # pinned_maven_install()
+    http_files = [
+        "load(\"@bazel_tools//tools/build_defs/repo:http.bzl\", \"http_file\")",
```
borkaehw: Can we have a comment at the very top of this file to remind users that this is a generated file which shouldn't be modified manually?
The message could be like `# Do not edit. rules_jvm_external autogenerates this file from maven_install Please run ... command to update this file.`

jin: It's JSON, so no comments :-) 

I can turn it into a "private" field in the JSON object, through.

borkaehw: Ah, I totally forget json has not comment. Thinking of a better way to remind users.

borkaehw: How about changing `maven_install.json` to something more obvious that implies we shouldn't modify it manually? Like `maven_autogenerated.json`.

jin: I think that still doesn't imply that you shouldn't modify it manually. Do you think users will end up modifying this manually often enough that it becomes a user experience problem?

borkaehw: I am thinking of a scenario that someone may think changing the artifacts version is as simple as modifying `maven_install.json`, but it may break other artifacts which have the dependency on the changed artifact.

Since I am going to introduce `rules_jvm_external` to the entire organization, there is a chance someone doesn't notice the proper way of using it.

jin: Perhaps we can hash the contents of the `dependencies` array, and add that as a field to the generated object? Then `pinned_coursier_maven` would compare the hash before generating the `http_file` targets and see if it was modified manually, and ask the user to not do that.

borkaehw: It sounds like a good idea to me.

borkaehw:
@laurentlb and @aehlig haven't had a chance to look at the PR yet; @jin is there anyone else who might be able to review this?

coursier.bzl
```diff
@@ -548,7 +675,49 @@ def _coursier_fetch_impl(repository_ctx):
...
+            # Coursier saves the artifacts into a subdirectory structure
+            # that mirrors the URL where the artifact's fetched from. Using
+            # this, we can reconstruct the original URL.
+            url = []
+            protocol = None
+            for part in artifact["file"].split("/"):
+                if protocol == None:
+                    if part == "http" or part == "https":
+                        protocol = part
+                        url.extend([protocol, ":/"])
```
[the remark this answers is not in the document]
jin: This actually works because of `protocol = part`. I've switched it to a more robust algorithm anyway.

coursier.bzl
```diff
@@ -438,23 +467,132 @@ def _get_java_proxy_args(repository_ctx):
...
-        ])
-    if (exec_result.return_code != 0):
-        fail("Error while trying to read %s: %s" % (filepath, exec_result.stderr))
-    return exec_result.stdout
+        bash = repository_ctx.os.environ.get("BAZEL_SH")
+        if (bash == None):
+            fail("Please set the BAZEL_SH environment variable to the path of MSYS2 bash. " +
+                 "This is typically `c:\\msys64\\usr\\bin\\bash.exe`. For more information, read " +
+                 "https://docs.bazel.build/versions/master/install-windows.html#getting-bazel")
+        repository_ctx.execute([bash, "-lc", "echo", "works"])
```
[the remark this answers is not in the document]
jin: I'll remove this in a follow up PR. This PR is quite big as-is.

coursier.bzl
jin: I'll add this in a follow up PR.

PR #171, opened 2019-06-20
Add RBE pipeline for CI

jin:


jin:
Ping @laurentlb @aehlig. This change adds a new CI pipeline that uses RBE.

PR #176, opened 2019-07-05
Generate and embed a signature of the dependency tree in maven_install.json

jin:
Prevent users from modifying `maven_install.json` manually by checking a signature at fetch time. This signature takes a sorted list of inputs comprised of elements from the dependency tree.

If manual modification is detected, print instructions to revert the modification and execute the correct pin script instead.

```
ERROR: An error occurred during the fetch of repository 'maven':
   maven_install.json contains an invalid signature and may be corrupted. PLEASE DO NOT MODIFY THIS FILE DIRECTLY! To generate a new maven_install.json and re-pin the artifacts, follow these steps: 

  1) In your WORKSPACE file, comment the 'maven_install_json' attribute in 'maven_install'.
  2) Run 'bazel run @maven//:pin'.
  3) Uncomment the 'maven_install_json' attribute in 'maven_install'.
```

For backwards compatibility, if `maven_install.json` doesn't contain a signature, users will see this message:

```
DEBUG: /usr/local/google/home/jingwen/.cache/bazel/_bazel_jingwen/88893d7c3918775765c5e3ae20d931b1/external/rules_jvm_external/coursier.bzl:546:9: NOTE: maven_install.json does not contain a signature entry of the dependency tree. This feature ensures that the file is not modified manually. To generate this signature, run `bazel run @unpinned_maven//:pin`.
```


borkaehw:
If this is for preventing people from modifying `maven_install.json` manually, I appreciate the work. We have started using `rules_jvm_external` 2.3 and seen this problem.

jin:
@borkaehw PTAL.

jin:
@laurentlb, @aehlig PTAL as well.

coursier.bzl
```diff
@@ -480,6 +480,24 @@ def _windows_check(repository_ctx):
...
+        signature_inputs.append(artifact["coord"])
+        if artifact["file"] != None:
+            signature_inputs.extend([
+                artifact["sha256"],
+                artifact["file"],
+                artifact["url"]
+            ])
+        if len(artifact["dependencies"]) > 0:
+            signature_inputs.append(",".join(sorted(artifact["dependencies"])))
+    return hash(repr(sorted(signature_inputs)))
```
borkaehw: I worry that sorting an entire `signature_inputs` could be slow when dealing with a huge `maven_install.json`, especially it seems like we will compute the hash for every `bazel build //...`. In addition, I found that
```
{
    "coord": "artifact1",
    "dependencies": [
        "deps1"
    ]
},
{
    "coord": "artifact2",
    "dependencies": [
        "deps2"
    ]
}
```
and 
```
{
    "coord": "artifact1",
    "dependencies": [
        "deps2"
    ]
},
{
    "coord": "artifact2",
    "dependencies": [
        "deps1"
    ]
}
```
could yield to the same hash. Even though users are very unlikely to make this kind of change, but the algorithm can be more robust.

I am thinking of something like
```
signature_inputs = []
for artifact in artifacts:
    artifact_group = []
    artifact_group.append(artifact["coord"])
    if artifact["file"] != None:
        artifact_group.extend([
            artifact["sha256"],
            artifact["file"],
            artifact["url"]
        ])
    if len(artifact["dependencies"]) > 0:
        artifact_group.append(",".join(sorted(artifact["dependencies"])))
    signature_inputs.append(artifact_group)
return hash(repr(sorted(signature_inputs, key=lambda arti: arti[0])))
```
It ties the corresponding `sha256`, `file`, `url`, and `dependencies` to the `coord`, and only sorts the entire artifact group.

jin: From my experience developing this so far, performance hasn't been an issue even with 2000+ lines in regression_testing_install.json. 

Thanks for the excellent suggestion to the algorithm, PTAL (there isn't `lambda` in Starlark, and I don't think there's a need to sort on `coord`).

jin:
> It looks good to me in general. One improvement I wish to have is a way to fix maven_install.json by running bazel run @unpinned_mavn//:pin without changing the JSON back. maven_install.json should be overwritten by new run, with or without changes in the artifacts list in WORKSPACE.

@borkaehw If you make a change in `maven_install.json`, `@maven` will need to reevaluate because it detects a change in the input file, and fail accordingly. This failure will cause `bazel run @unpinned_maven//:pin` to also fail, even if you just specify `@unpinned_maven//:pin` as the target to run. To overwrite a corrupted `maven_install.json`, one would need to comment/remove the `maven_install_json` attribute in `maven_install` and re-run the pin command.

coursier.bzl
borkaehw: > From my experience developing this so far, performance hasn't been an issue even with 2000+ lines in regression_testing_install.json.

We currently have up to 15000 lines, but we could worry about the performance once it is really a concern.

> there isn't lambda in Starlark, and I don't think there's a need to sort on coord

Good point, sorry I almost forgot it's not Python.

borkaehw:
It looks great. Thanks for the work!

Issue #178, opened 2019-07-05
external/maven/pin: line 5: syntax error near unexpected token `)'


beasleyr-vmw:
It looks like Maven artifact version ranges create a problem for the "pin" helper because of unescaped parentheses.

Steps to reproduce:

```console
$ cat WORKSPACE
...
load("@rules_jvm_external//:defs.bzl", "maven_install")

maven_install(
    name = "maven",
    artifacts = ["io.kubernetes:client-java:4.0.0-beta1"],
    ...
)
$ bazel run @maven//:pin
.../external/maven/pin: line 5: syntax error near unexpected token `)'
```

Examining the generated script, I see that it's one big `echo` command with escaped characters. Any artifact containing a "ranged" dependency causes the script to contain unescaped closing parentheses characters.

Example json (check out com.nimbusds:nimbus-jose-jwt):

```json
{
  "coord": "io.kubernetes:client-java:4.0.0-beta1",
  "file": "v1/https/build-artifactory.eng.vmware.com/artifactory/maven/io/kubernetes/client-java/4.0.0-beta1/client-java-4.0.0-beta1.jar",
  "dependencies": [
    "javax.activation:activation:1.1",
    "io.sundr:sundr-codegen:0.9.2",
    "io.kubernetes:client-java-api:4.0.0-beta1",
    "joda-time:joda-time:2.9.3",
    "net.minidev:json-smart:1.3.1",
    "io.sundr:sundr-core:0.9.2",
    "com.google.code.findbugs:jsr305:3.0.2",
    "com.squareup.okio:okio:1.6.0",
    "io.sundr:builder-annotations:0.9.2",
    "com.github.stephenc.jcip:jcip-annotations:1.0-1",
    "com.google.code.gson:gson:2.8.0",
    "com.google.guava:guava:25.1-jre",
    "com.nimbusds:nimbus-jose-jwt:[4.29,)",
    "commons-codec:commons-codec:1.11",
    "io.kubernetes:client-java-proto:4.0.0-beta1",
    "com.nimbusds:lang-tag:[1.4.3,)",
    "org.bouncycastle:bcprov-ext-jdk15on:1.59",
    "org.slf4j:slf4j-api:1.7.25",
    "com.microsoft.azure:adal4j:1.6.0",
    "com.nimbusds:oauth2-oidc-sdk:5.24.1",
    "org.apache.commons:commons-compress:1.18",
    "io.swagger:swagger-annotations:1.5.12",
[...]

PR #179, opened 2019-07-05
Sidestep trouble with escaping json in pin.sh

beasleyr-vmw:
pin.sh had trouble serializing an artifact dependency identified by a version
range (ex: "[1.2.3,4.5)") because of unescaped end parentheses.  Work around
this by moving pin.sh's JSON template placeholder into a here-doc statement
and remove the corresponding quote-escape code from coursier.bzl.

Open q: Do we have to worry about command injection?

Resolves https://github.com/bazelbuild/rules_jvm_external/issues/178.

Testing Done:
- Generated a dependency set which named "io.kubernetes:client-java:4.0.0-beta1",
  and the generated pin script no longer chokes.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -134,6 +134,8 @@ maven_install(
         "ch.epfl.scala:compiler-interface:1.3.0-M4+20-c8a2f9bd",
         # https://github.com/bazelbuild/rules_jvm_external/issues/172
         "org.openjfx:javafx-base:11.0.1",
+        # https://github.com/bazelbuild/rules_jvm_external/issues/178
+        "io.kubernetes:client-java:4.0.0-beta1",
     ],
     repositories = [
         "https://repo1.maven.org/maven2",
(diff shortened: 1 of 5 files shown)


jin:
Thanks! Could you add `io.kubernetes:client-java:4.0.0-beta1` to [this list](https://github.com/bazelbuild/rules_jvm_external/blob/40a379d3d50fc755ea4671648d38e78910c7673e/WORKSPACE#L106) and [this target](https://github.com/bazelbuild/rules_jvm_external/blob/40a379d3d50fc755ea4671648d38e78910c7673e/tests/unit/build_tests/BUILD#L6)?

beasleyr-vmw:
> Thanks! Could you add `io.kubernetes:client-java:4.0.0-beta1` to [this list](https://github.com/bazelbuild/rules_jvm_external/blob/40a379d3d50fc755ea4671648d38e78910c7673e/WORKSPACE#L106) and [this target](https://github.com/bazelbuild/rules_jvm_external/blob/40a379d3d50fc755ea4671648d38e78910c7673e/tests/unit/build_tests/BUILD#L6)?

Done! Added as squash commits; lmk if I need to perform squash locally and force-push as a single coherent commit.

BTW, I regenerated regression_testing_install.json on a Linux machine, and `javafx-base-11.0.1-mac.jar` became `javafx-base-11.0.1-linux.jar`. Is this expected, or should I file a new issue?

Issue #190, opened 2019-07-11
rules_jvm_external vs java_import_external


Dig-Doug:
Hi,

I'm trying to updated a library [here](https://github.com/bazelbuild/rules_webtesting/pull/363) to be compatible with rules_jvm_external. However, there's some confusion about what the current recommended to import external jars and how to reference them. So I've got two questions:

- Which method, `rules_jvm_external` or `java_import_external` (or something else?), is the recommended way to reference external java deps?

- Also, when referencing deps in a library (that other bazel projects will import), what's the best way to reference the deps to be compatible with as many as users as possible (who may be using other jar import methods)?

Thanks,

Doug

jin:
> Which method, rules_jvm_external or java_import_external (or something else?), is the recommended way to reference external java deps?

`java_import_external` is not transitive by default. It is (+ `jvm_maven_import_external`) a drop-in replacement for `maven_jar`. `rules_jvm_external` is a more end-to-end solution for managing a collection of Maven artifacts that needs to be resolved transitively.

> Also, when referencing deps in a library (that other bazel projects will import), what's the best way to reference the deps to be compatible with as many as users as possible (who may be using other jar import methods)?

If you're making a library, I think the best approach is to namespace a `maven_install` repository to your library's name and ask downstream users to add a `maven_install` of that same name with the exact list of artifacts that you can expose through a .bzl load. See https://github.com/buchgr/rules_jmh for an example.

I would advise against going with the `@group_artifact//jar` approach as that's only limited to JAR files, and doesn't encode any information about versions. Downstream users may specify a different version of the artifact, that although resolves to the expected `@group_artifact//jar` label, is of a different version and can cause build failures.

jin:
Addendum: that said, I've opened https://github.com/bazelbuild/rules_jvm_external/pull/192 based on feedback in https://github.com/bazelbuild/rules_webtesting/pull/363 so `rules_jvm_external` is more interoperable with `java_import_external`.

Dig-Doug:
Thanks for the quick fix!

One follow up question about the `maven_install` of the library:

If the user code references external java lib in their code, they'll probably want to specify the dep explicitly in their `maven_install`. If the library, e.g. `rules_webtesting`, also specifies the same library (maybe with a different version) in the bzl `maven_install`, what happens? e.g.

```python
# WORKSPACE
maven_install(
  artifacts = ["lib1:v1"],
)

rules_webtesting_java_deps()
```
```python
[...]

PR #192, opened 2019-07-12
Add jar-less alias target for generate_compat_repositories mode

jin:
Make the generated compat repositories also generate the `@group_artifact//:group_artifact` / `@group_artifact` target label on top of `@group_artifact//jar` label to be interoperable with [`java_import_external`](https://github.com/bazelbuild/bazel/blob/master/tools/build_defs/repo/java.bzl#L25
). Some users refer to targets generated by `java_import_external` as `@group_artifact` instead of `@group_artifact//jar`. This was motivated by feedback in https://github.com/bazelbuild/rules_webtesting/pull/363#issuecomment-510121372.

Fixes https://github.com/bazelbuild/rules_jvm_external/pull/190

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,7 @@ maven_install(
     artifacts = [
         "junit:junit:4.12",
         "androidx.test.espresso:espresso-core:3.1.1",
+        "org.hamcrest:hamcrest-library:1.3",
     ],
     repositories = [
         # Private repositories are supported through HTTP Basic auth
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -58,11 +59,19 @@ maven_install(
 and use them directly in the BUILD file by specifying the versionless target alias label:
 
 ```python
+java_library(
```
jin: drive-by addition to improve the first-impression example

PR #202, opened 2019-07-18
Fix incompatibility with --incompatible_string_join_requires_strings

jin:


jin:
Actually, this line will be deleted in #203.

PR #208, opened 2019-07-22
Add Linux CI pipelines for 0.25.3 to 0.28.1 (latest) and update bazel-toolchains pin

jin:
This ensures that `rules_jvm_external` is backwards compatible for these versions.

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -14,7 +14,39 @@ tasks:
       # These tests are currently incompatible with RBE
       - "-//tests/integration:UnsafeSharedCacheTest"
       - "-//tests/integration/override_targets"
-  ubuntu1804:
+  ubuntu1804_0_25_3:
+    platform: ubuntu1804
+    bazel: 0.25.3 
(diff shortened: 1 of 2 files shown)


.bazelci/presubmit.yml
```diff
@@ -14,7 +14,39 @@ tasks:
       # These tests are currently incompatible with RBE
       - "-//tests/integration:UnsafeSharedCacheTest"
       - "-//tests/integration/override_targets"
-  ubuntu1804:
+  ubuntu1804_0_25_3:
+    platform: ubuntu1804
+    bazel: 0.25.3 
+    shell_commands:
+      - bazel run @unpinned_regression_testing//:pin
```
c-parsons: I'm unfamiliar with including both shell_commands and test_targets. What happens? Does shell_commands get run as part of setup?

jin: Yeah: https://buildkite.com/bazel/rules-jvm-external/builds/667#ac7a4379-d6a8-4a2b-b977-068aca09691d/140

PR #209, opened 2019-07-24
Prefix workspace name while referencing json

borkaehw:
When workspace A is calling `maven_install` of an external workspace B, the syntax of `maven_install_json = "//:maven_install.json"` will try to look for `maven_install.json` in workspace A. `maven_install_json = "@B//:maven_install.json"` will solve this problem.

Fix: #212

### DIFF
--- a/private/pin.sh
+++ b/private/pin.sh
@@ -2,6 +2,8 @@
 
 set -euo pipefail
 readonly maven_install_json_loc=$BUILD_WORKSPACE_DIRECTORY/{repository_name}_install.json
+readonly execution_root=$(bazel info execution_root)
+readonly workspace_name=$(basename $execution_root)
 cat <<"RULES_JVM_EXTERNAL_EOF" | python -m json.tool > $maven_install_json_loc
 {dependency_tree_json}
 RULES_JVM_EXTERNAL_EOF
@@ -19,7 +21,7 @@ cat <<EOF
 maven_install(
     artifacts = # ...,
     repositories = # ...,
-    maven_install_json = "//:{repository_name}_install.json",
+    maven_install_json = "@$workspace_name//:{repository_name}_install.json",
 )
 
 load("@{repository_name}//:defs.bzl", "pinned_maven_install")


private/pin.sh
```diff
@@ -2,6 +2,7 @@
 
 set -euo pipefail
 readonly maven_install_json_loc=$BUILD_WORKSPACE_DIRECTORY/{repository_name}_install.json
+readonly workspace_name=$(basename $BUILD_WORKSPACE_DIRECTORY)
```
aehlig: Here we depend on an implementation detail of bazel (the naming of the directory an external repository is put into). While the directory name is unlikely to change in the near future, it is still good paractise to only depend on official interfaces, not implementation details, as the latter is fragile in the long run. Would providing [`repository_ctx.name`](https://docs.bazel.build/versions/master/skylark/lib/repository_ctx.html#name) to the script be an alternative to the `basename` computation?

jin: Yes, the `repository_ctx.name` is not necessarily the same as the directory's basename.

borkaehw:
Thanks for the feedback! I also noticed there are comments and readmes that reference the original format. What would be a good way to express this change?

private/pin.sh
borkaehw: I don't think `repository_ctx.name` is what we want here. `repository_ctx.name` is returning the name we define in `maven_install`, either `maven` or `unpinned_maven`. What we are looking for is the name of workspace like `rules_jvm_external` or `io_bazel`. I have limited knowledge how to get it in `coursier.bzl`.

jin: Hmm.. even if we get the local workspace name when pinning artifacts, the problem is that if you're exporting a `maven_install` to be loaded, the local repository may not name your external repo as the repo used in the `maven_install_json` attribute.

e.g. if you export `maven_install(maven_install_json = "@annex//:maven_install.json")` in a macro to be loaded, downstream workspaces have to name this dependency `annex` and not something else. This is tricky and I think it may require usage of repository remapping.

This also raises an important question on how to connect repositories that use `rules_jvm_external`. The [current guidance in the README](https://github.com/bazelbuild/rules_jvm_external#exporting-and-consuming-artifacts-from-external-repositories) is to load the list of artifacts from the external repo and merge them with your local `maven_install`.

An idea: If you're using your local `maven_install.json`, use `@//:maven_install.json` or just `//:maven_install.json`. If you're using a external repository's `maven_install.json`, use a function from `rules_jvm_external` that accepts the name assigned to the external repository in the local workspace, and loads some set of variables by convention from the external dependency.

```python
load("@rules_jvm_external//:defs.bzl", "external_maven_install")

http_archive(
    name = "annex",
    urls = [...],
    repo_mapping = {"@maven": "@annex_maven"},
)
# annex_foo workspace depends on `@maven//`, 
# but `@maven` may already be used by the local workspace.

# creates `@annex_maven`
external_maven_install(name = "annex_maven", for_repository = "annex")
```

where `external_maven_install` is a macro that creates a repository rule that loads `@<repository_name>//:maven_install.json` by default and names it `<repository_name>_maven`.

jin: Filed https://github.com/bazelbuild/rules_jvm_external/issues/212

jin:
I just discovered a way to reliably get the local workspace name:

```
{
  pushd $BUILD_WORKSPACE_DIRECTORY
  readonly execution_root=$(bazel info execution_root)
  readonly workspace_name=$(basename $execution_root)
  popd
}
```

perhaps we can try that?

borkaehw:
Good idea and it works. Please take a look again.

private/pin.sh
```diff
@@ -19,7 +21,7 @@ cat <<EOF
 maven_install(
     artifacts = # ...,
     repositories = # ...,
-    maven_install_json = "//:{repository_name}_install.json",
+    maven_install_json = "@${workspace_name}//:{repository_name}_install.json",
```
jin: `@$workspace_name` instead of `@${workspace_name}` because `{..}` is used for the `repository_ctx.template` format.

borkaehw: Thanks.

jin:
Thanks!

Issue #218, opened 2019-08-06
Errors while updating from 2.5 to 2.6


natros:
Hi,

```
$ bazelisk clean --expugne
```

```
$ bazelisk info
Starting local Bazel server and connecting to it...
bazel-bin: /private/var/tmp/_bazel_fsousa/5eb111cf39e94839e993e7dc498f4d7c/execroot/demos/bazel-out/darwin-fastbuild/bin
bazel-genfiles: /private/var/tmp/_bazel_fsousa/5eb111cf39e94839e993e7dc498f4d7c/execroot/demos/bazel-out/darwin-fastbuild/bin
bazel-testlogs: /private/var/tmp/_bazel_fsousa/5eb111cf39e94839e993e7dc498f4d7c/execroot/demos/bazel-out/darwin-fastbuild/testlogs
character-encoding: file.encoding = ISO-8859-1, defaultCharset = ISO-8859-1
command_log: /private/var/tmp/_bazel_fsousa/5eb111cf39e94839e993e7dc498f4d7c/command.log
committed-heap-size: 257MB
execution_root: /private/var/tmp/_bazel_fsousa/5eb111cf39e94839e993e7dc498f4d7c/execroot/demos
gc-count: 3
gc-time: 28ms
install_base: /var/tmp/_bazel_fsousa/install/8772b695a4a7b498a4dd3eff54505d9c
java-home: /private/var/tmp/_bazel_fsousa/install/8772b695a4a7b498a4dd3eff54505d9c/_embedded_binaries/embedded_tools/jdk
java-runtime: OpenJDK Runtime Environment (build 11.0.2+7-LTS) by Azul Systems, Inc.
java-vm: OpenJDK 64-Bit Server VM (build 11.0.2+7-LTS, mixed mode) by Azul Systems, Inc.
max-heap-size: 3817MB
output_base: /private/var/tmp/_bazel_fsousa/5eb111cf39e94839e993e7dc498f4d7c
output_path: /private/var/tmp/_bazel_fsousa/5eb111cf39e94839e993e7dc498f4d7c/execroot/demos/bazel-out
package_path: %workspace%
release: release 0.28.1
repository_cache: /var/tmp/_bazel_fsousa/cache/repos/v1
server_log: /private/var/tmp/_bazel_fsousa/5eb111cf39e94839e993e7dc498f4d7c/java.log.nostradamus.fsousa.log.java.20190806-153858.8951
server_pid: 8951
[...]

PR #219, opened 2019-08-06
Fix private repository authentication regression in 2.6

jin:
We didn't take private repository artifacts into account when implementing mirror_urls. Fixes https://github.com/bazelbuild/rules_jvm_external/issues/218


jin:
It turns out that maven_install.json does not work with private repository urls at all, but the regular unpinned case should work.

Merging this as it's blocking users from upgrading. @aehlig / @laurentlb / @c-parsons, please review and leave comments if you'd like.

PR #221, opened 2019-08-12
WIP: mvn_package rule

jin:


artem-zinnatullin:
What use case does it address? Seems like a way to represent Maven projects as Bazel targets from project sources but usually those are already published to some Maven repository

mvn.bzl
```diff
@@ -0,0 +1,41 @@
...
+    target_dir = ctx.actions.declare_directory("target")
+    outputs = [output_jar, target_dir]
+
+    # -Djar.finalName=custom-jar-name
+    ctx.actions.run_shell(
+        inputs = inputs,
+        outputs = outputs,
+        mnemonic = "MvnPackage",
+        progress_message = "Running 'mvn package' for %s" % output_jar.short_path,
+        command = "mvn package -DskipTests -DbazelOutputDir=%s -Pbazel" % (output_jar.dirname),
```
artem-zinnatullin: We probably shouldn't expect `mvn` be in the system and on PATH, should it be lazily downloaded from https://maven.apache.org/download.cgi?

jin: it can be - this is just a quick prototype.

jin:
> What use case does it address? Seems like a way to represent Maven projects as Bazel targets from project sources but usually those are already published to some Maven repository

Incremental migration. If you have a monorepo that's built with Maven, you can wrap existing Maven modules and use their outputs in java_* rules.

artem-zinnatullin:
Got it, sounds good 👍

mvn.bzl
bnczk: Trying this rule when having the pom.xml not in the root package results in the mvn command failing as it does not find the pom.xml.
`mvn package -f %s -DskipTests -DbazelOutputDir=%s -Pbazel" % (ctx.file.pom_xml.dirname,output_jar.dirname)` worked for me.

walles:
@jin this is still a Draft.

Do you think you will do more work to get this merged?

jin:
Unlikely, I don't have the bandwidth to work on this anytime soon.

PR #234, opened 2019-08-29
Write documentation on generated label syntax

jin:


jin:
This is a documentation update, PTAL.

Issue #94, opened 2019-04-02
Transitive bazel dependencies have public visibility


realityforge:
As a reasonably new user I found it extremely surprising that a transitive dependency that was not declared in the `artifacts` attribute had public visibility. I would have expected that the way to make a transitive artifact public was to supply a 2-part spec in the `artifacts` attribute.

This would make it much easier for relative beginners to read the rules as the public targets would be explicitly declared.

Was this ever considered?

Other than that, it has been reasonably easy to get started. Thanks!

jin:
Yes, we should definitely restrict the visibility of unspecified artifacts. I'll mark this as a good first issue for anyone to take up if they want to give it a shot.

mpassell:
This seems like a good idea in principle, but there are unfortunately some dependencies that seem to be specifically set up to bring in dependencies for you to directly reference. The [Spring Boot starters](https://www.baeldung.com/spring-boot-starters) would be a good example. It seems like a bad practice to me, but they get used pretty heavily. :(  So, even if this gets done, I think there should be an option to turn it off.

jin:
> So, even if this gets done, I think there should be an option to turn it off.

I agree. It's also possible for the rule, or use `bazel query`, to provide the flat list of artifacts that are actually used, so it's simple to switch to this behavior.

PR #237, opened 2019-09-03
Add `strict_visibility` attr to hide transitive dependencies (#94)

beasleyr-vmw:
This commit adds an attribute, `strict_visibility`, to allow users to hide
transitive Maven dependencies, forcing the user to explicitly name artifacts
in their `maven_install`'s `artifacts` list before being able to reference
in downstream build rules.

Resolves https://github.com/bazelbuild/rules_jvm_external/issues/94.

Testing Done:
- `bazel test //tests/unit/build_tests/...` succeeded.
  `bazel test //tests/unit/build_tests:strict_version_neg` failed, as expected.


coursier.bzl
```diff
@@ -333,6 +349,10 @@ def _generate_imports(repository_ctx, dep_tree, neverlink_artifacts, override_ta
 
             target_import_string.append("".join(target_import_labels) + "\t],")
             target_import_string.append("\ttags = [\"maven_coordinates=%s\"]," % artifact["coord"])
+
+            if (repository_ctx.attr.strict_visibility and explicit_artifacts.get(_strip_packaging_and_classifier_and_version(artifact["coord"]))):
```
jin:
```suggestion
            if repository_ctx.attr.strict_visibility and explicit_artifacts.get(target_label):
```

coursier.bzl
```diff
@@ -268,7 +268,23 @@ def _generate_imports(repository_ctx, dep_tree, neverlink_artifacts, override_ta
...
+            # 	name = "org_hamcrest_hamcrest_library",
+            # 	jars = ["https/repo1.maven.org/maven2/org/hamcrest/hamcrest-library/1.3/hamcrest-library-1.3.jar"],
+            # 	srcjar = "https/repo1.maven.org/maven2/org/hamcrest/hamcrest-library/1.3/hamcrest-library-1.3-sources.jar",
+            # 	deps = [
+            # 		":org_hamcrest_hamcrest_core",
+            # 	],
+            #   tags = ["maven_coordinates=org.hamcrest:hamcrest.library:1.3"],
+            #   neverlink = True,
+            #   visibility = ["//visibility:public"],
+            if (repository_ctx.attr.strict_visibility and explicit_artifacts.get(_strip_packaging_and_classifier_and_version(artifact["coord"]))):
```
jin:
```suggestion
            if repository_ctx.attr.strict_visibility and explicit_artifacts.get(target_label):
```

coursier.bzl
```diff
@@ -866,6 +898,14 @@ pinned_coursier_fetch = repository_rule(
         "generate_compat_repositories": attr.bool(default = False),  # generate a compatible layer with repositories for each artifact
         "maven_install_json": attr.label(allow_single_file = True),
         "override_targets": attr.string_dict(default = {}),
+        "strict_visibility": attr.bool(
+            doc = """Controls visibility of transitive dependencies
```
jin:
```suggestion
            doc = """Controls visibility of transitive dependencies.
```

coursier.bzl
```diff
@@ -866,6 +898,14 @@ pinned_coursier_fetch = repository_rule(
...
+
+            If "True", transitive dependencies are private and invisible to user's rules.
+            If "False", transitive dependencies are public and visible to user's rules.
+            """,
+            default = True,
```
jin: Let's set this to False to prevent this from being a breaking change, and also be consistent with the macro.

coursier.bzl
```diff
@@ -897,6 +937,14 @@ coursier_fetch = repository_rule(
         ),
         "maven_install_json": attr.label(allow_single_file = True),
         "override_targets": attr.string_dict(default = {}),
+        "strict_visibility": attr.bool(
+            doc = """Controls visibility of transitive dependencies
+
+            If "True", transitive dependencies are private and invisible to user's rules.
+            If "False", transitive dependencies are public and visible to user's rules.
+            """,
+            default = True,
```
jin: False as well.

WORKSPACE
```diff
@@ -220,6 +220,19 @@ load(
...
 
+maven_install(
+    name = "strict_visibility_testing",
+    artifacts = [
+        # https://github.com/bazelbuild/rules_jvm_external/issues/94
+        "org.apache.tomcat:tomcat-catalina:9.0.24",
+    ],
+    repositories = [
+        "https://repo1.maven.org/maven2",
+        "https://maven.google.com",
```
jin: rm Google Maven

coursier.bzl
jin: Oh, I made a mistake here. target_label is escaped, but you're comparing with the non-escaped version. Feel free to revert.

beasleyr-vmw: Ha, yeah, _just_ ran into that. Backing out.

beasleyr-vmw:
> Looks good, can you also add a section in the README about this? You can follow the documentation convention of other attributes like `generate_compat_repositories`.

Sure thing! Pushed an update; hope the text reads okay.

jin:
Hmm, it looks like the versioned aliases of the top level targets don't have the right visibility assigned to them:

```diff
diff --git a/tests/unit/build_tests/BUILD b/tests/unit/build_tests/BUILD
index 6d3fa04..356e525 100644
--- a/tests/unit/build_tests/BUILD
+++ b/tests/unit/build_tests/BUILD
@@ -56,6 +56,7 @@ build_test(
     name = "strict_version",
     targets = [
         "@strict_visibility_testing//:org_apache_tomcat_tomcat_catalina",
+        "@strict_visibility_testing//:org_apache_tomcat_tomcat_catalina_9_0_24",
     ],
 )
```

```
$ bazel query --output=build @strict_visibility_testing//:org_apache_tomcat_tomcat_catalina_9_0_24
# /private/var/tmp/_bazel_jingwen/5e64b778ca5f4ff8a6ca994b3f33d8f8/external/strict_visibility_testing/BUILD:50:1
alias(
  name = "org_apache_tomcat_tomcat_catalina_9_0_24",
  actual = "@strict_visibility_testing//:org_apache_tomcat_tomcat_catalina",
)
```

beasleyr-vmw:
> Hmm, it looks like the versioned aliases of the top level targets don't have the right visibility assigned to them:

Ack. Good catch. Pushed two fixups to address. (The first resolves the alias issue. The second is just optional sugar.)

PR #245, opened 2019-09-15
add resolve_timeout attribute

cheister:
Some of our resolves are starting to take longer than the default 600 second timeout for `repository_ctx.execute`.  

This PR adds a `resolve_timeout` attribute to `maven_install` so the timeout can be changed if needed.

### DIFF
--- a/defs.bzl
+++ b/defs.bzl
@@ -29,7 +29,8 @@ def maven_install(
         version_conflict_policy = "default",
         maven_install_json = None,
         override_targets = {},
-        strict_visibility = False):
+        strict_visibility = False,
+        resolve_timeout = 600):
     """Resolves and fetches artifacts transitively from Maven repositories.
 
     This macro runs a repository rule that invokes the Coursier CLI to resolve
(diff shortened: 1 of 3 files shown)


jin:
Instead of exposing another API, what do you think about setting the default to a high value? Also, how many artifacts are you resolving, if you can say?

cheister:
Setting a high value penalizes users who might have connection issues and don't need to resolve very many artifacts.  I could also imagine users wanting a lower timeout than 600 seconds.  Seems better to allow users to configure it to what their system needs than try to guess a time that will work for everyone.

In our case we are resolve around 650 artifacts.

defs.bzl
```diff
@@ -29,7 +29,8 @@ def maven_install(
         version_conflict_policy = "default",
         maven_install_json = None,
         override_targets = {},
-        strict_visibility = False):
+        strict_visibility = False,
+        resolve_timeout = 600):
```
jin: Please add argument documentation in the docstring below.

cheister:
Yeah, we're definitely using pinned artifacts / maven_install.json.

jin:
Thanks!

PR #248, opened 2019-09-18
Add missing case for primary_artifact_path

puffnfresh:
I have an entry of:

```
maven_install(
    name = "zzz",
    artifacts = [
        "org.bouncycastle:bcpkix-jdk15on:1.61",
    ],
    repositories = [
        "https://xxx.yyy/repo",
    ),
)
```

And I have something in my `~/.config/coursier/credentials.properties` to give my creds. This confuses things currently because Coursier includes my name, but the repo doesn't know about it. I was previously getting an error because of `primary_artifact_path` being referenced before assigned.

### DIFF
--- a/coursier.bzl
+++ b/coursier.bzl
@@ -776,6 +776,9 @@ def _coursier_fetch_impl(repository_ctx):
                     # So, we use rfind to get the index of the final '@' in the
                     # repository url instead.
                     primary_artifact_path = primary_url[len(url):]
+                elif url.find("@") == -1 and primary_url.find("@") != -1:
+                    username_len = primary_url.find("//") + 2 + primary_url.find("@") + 1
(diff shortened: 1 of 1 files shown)


coursier.bzl
```diff
@@ -776,6 +776,9 @@ def _coursier_fetch_impl(repository_ctx):
                     # So, we use rfind to get the index of the final '@' in the
                     # repository url instead.
                     primary_artifact_path = primary_url[len(url):]
+                elif url.find("@") == -1 and primary_url.find("@") != -1:
+                    username_len = primary_url.find("//") + 2 + primary_url.find("@") + 1
```
jin: Could you add some comments here to help visualize the values? And also a comment on when this codepath will be used.

jin:
Thanks!

PR #254, opened 2019-10-07
Allow strict visibility to work for artifacts with classifiers

cheister:
strict_visibility does not work for artifacts with classifiers.  e.g.

```
maven_install(
    name = "strict_visibility_testing",
    artifacts = [
        maven.artifact(
            group = "org.eclipse.jetty",
            artifact = "jetty-http",
            version = "9.4.20.v20190813",
            classifier = "tests",
        ),
    ],
    repositories = [
        "https://repo1.maven.org/maven2",
    ],
    strict_visibility = True,
)
```
will fail with
```
ERROR: rules_jvm_external/tests/unit/build_tests/BUILD:71:1: in genrule rule //tests/unit/build_tests:strict_version_classifier_0__deps: target '@strict_visibility_testing//:org_eclipse_jetty_jetty_http_tests' is not visible from target '//tests/unit/build_tests:strict_version_classifier_0__deps'. Check the visibility declaration of the former target if you think the dependency is legitimate
```
this PR adds classifier to the `explicit_artifacts` dictionary used to lookup the strict visibility targets.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -225,6 +225,13 @@ maven_install(
     artifacts = [
         # https://github.com/bazelbuild/rules_jvm_external/issues/94
         "org.apache.tomcat:tomcat-catalina:9.0.24",
+        # https://github.com/bazelbuild/rules_jvm_external/issues/255
+        maven.artifact(
+            group = "org.eclipse.jetty",
+            artifact = "jetty-http",
+            version = "9.4.20.v20190813",
+            classifier = "tests",
+        ),
     ],
     repositories = [
         "https://repo1.maven.org/maven2",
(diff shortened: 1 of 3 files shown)


WORKSPACE
```diff
@@ -225,6 +225,12 @@ maven_install(
     artifacts = [
         # https://github.com/bazelbuild/rules_jvm_external/issues/94
         "org.apache.tomcat:tomcat-catalina:9.0.24",
+        maven.artifact(
+            group = "org.eclipse.jetty",
+            artifact = "jetty-http",
+            version = "9.4.20.v20190813",
+            classifier = "tests",
```
jin: can you add a link to this issue here?

cheister:
Fixes https://github.com/bazelbuild/rules_jvm_external/issues/255

PR #257, opened 2019-10-09
Add support for custom maven_install.json filenames and locations

### LINKED ISSUE #251: [FR] - Create Pinned json in a different directory?
Would it be possible for `@repo//:pin` and `@unpinned//:pin` commands to generate the pinned json file under a different directory?

I.e. given:

```
WORKSPACE
bazel_extensions/
  pinned_dependencies/
    repo_install.json
```

Could we write to the same place we read the file from? I.e. `maven_install_json: "//:bazel_extensions/pinned_dependencies/repo_install.json" will read from that file, but when running `@unpinned_repo//:pin` won't update that file and instead creates a new one in the top dir.

jin:
This PR adds support for custom `maven_install.json` locations, so they don't always need to be at `//:maven_install.json`.

e.g.

```
maven_install(
    name = "maven_install_in_custom_location",
    artifacts = ["com.google.guava:guava:27.0-jre"],
    repositories = ["https://repo1.maven.org/maven2"],
    maven_install_json = "@rules_jvm_external//tests/custom_maven_install:maven_install.json",
)
```

By default, the location is still `//:maven_install.json`. However, during an artifact pinning update and the `maven_install_json` attribute points to a custom file target, the pinning script respects the new location and applies the changes accordingly, instead of generating the json file at the default location.

Fixes #251


jin:
Thanks for the quick review @c-parsons!

PR #264, opened 2019-10-12
Fix Stardoc build breakage, update Stardoc, and generate new docs

jin:


WORKSPACE
```diff
@@ -35,11 +35,11 @@ bazel_skylib_workspace()
 
 load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")
 
-git_repository(
+http_archive(
     name = "io_bazel_skydoc",
```
c-parsons: Could you also rename this io_bazel_stardoc now? (If it's too much effort in this PR, feel free to defer

jin:
@c-parsons the breakage was from https://github.com/bazelbuild/rules_jvm_external/pull/262, where the `stardoc` rule in //:BUILD wasn't updated with a newly added bzl. It wasn't from Stardoc.

WORKSPACE
jin: Sure, done

PR #282, opened 2019-11-01
Add documentation on repo_mapping for maven_jar migration purposes

jin:
If upstream dependencies hardcodes the maven_jar repository name, this is a good solution / workaround without having to update the dependencies themselves.


jin:
@laurentlb ptal?

jin:
thanks.

PR #296, opened 2019-11-22
Drop dependency on Python by implementing a small Java SHA256 hasher

### LINKED ISSUE #291: Drop dependency on python for sha256 hashing
Multiple reports:

- https://github.com/bazelbuild/rules_jvm_external/issues/281
- https://stackoverflow.com/questions/58827854/bazel-sha256-checksum/58843439#58843439

It's a major issue on Windows, not so much on *nix platforms.

jin:
Fixes #291 

Many Windows users do not have Python installed (see #291), but that's what we use for sha256 hashing currently (`@bazel_tools//tools/build_defs/hash:sha256.py`). This PR implements a small hasher written in Java.

Also regenerate `maven_install.json` for a couple of examples to test that the generated sha256 is correct.


private/tools/java/rules/jvm/external/Hasher.java
```diff
@@ -0,0 +1,50 @@
...
+    MessageDigest digest = MessageDigest.getInstance("SHA-256");
+    try (BufferedInputStream bufferedInputStream =
+        new BufferedInputStream(new FileInputStream(filename))) {
+      while ((count = bufferedInputStream.read(buffer)) > 0) {
+        digest.update(buffer, 0, count);
+      }
+    }
+    // Convert digest byte array to a hex string.
+    byte[] hashDigest = digest.digest();
+    StringBuffer hexString = new StringBuffer();
```
artem-zinnatullin: pls don't use `StringBuffer`, all of its methods are `synchronized`, `StringBuilder` is what you want :)

See `StringBuffer` javadoc for details https://docs.oracle.com/javase/10/docs/api/java/lang/StringBuffer.html

private/tools/java/rules/jvm/external/Hasher.java
```diff
@@ -0,0 +1,50 @@
...
+import java.io.IOException;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+
+/** A tool to compute the sha256 hash of a file. */
+public class Hasher {
+
+  public static void main(String[] args) throws NoSuchAlgorithmException, IOException {
+    // Since this tool is for private usage, just do a simple assertion for the filename argument.
+    assert (args.length == 1) : "Please specify the path of the file to hash.";
```
artem-zinnatullin: but do you ever expect to run with assertions enabled? `-ea`

artem-zinnatullin: Resolved :)

artem-zinnatullin: Or I guess you can just remove the assertion now since you throw exception otherwise

jin: done, thanks.

private/tools/java/rules/jvm/external/Hasher.java
jin: done.

PR #299, opened 2019-11-27
Import tools/build_defs/repo/jvm.bzl from Bazel

jin:
This facilitates maven_jar migration, because jvm_maven_import_external's API is closer to maven_jar than rules_jvm_external. 

By bringing it into rules_jvm_external, we can also decouple the release process from Bazel.

I imported it into the same tree location at `//tools/build_defs/repo:jvm.bzl` to simplify migration: `s/bazel_tools/rules_jvm_external/`.

There's no automated syncing between upstream Bazel and this file, but the file doesn't change too frequent anyway.


jin:
Dropping plans for this for now.

PR #303, opened 2019-12-03
Fix handling of user:pass@host urls so that http auth can work for pinned artifacts

justhecuke:
The plumbing underneath the url handling still needs to be fixed (which may include changes to bazel itself) so that http auth can work, but this is a necessary part that needed to be fixed anyways.

It was the case that user:pass@host urls would be consistently parsed incorrectly for the maven artifact path which led to nonsensical urls in the mirror_urls list. This should fix that parsing and allow us to properly append the maven artifact path to the list of repositories and generate urls that will actually fetch the artifacts we want.


coursier.bzl
```diff
@@ -272,6 +272,35 @@ def _pinned_coursier_fetch_impl(repository_ctx):
...
+    return new_url
+
+def infer_artifact_path_from_primary_and_repos(primary_url, repository_urls):
+    """Returns the artifact path inferred by comparing primary_url with urls in repository_urls.
+
+    Returns:
+        String of the artifact path used by maven to find a particular artifact. Does not have a leading slash (`/`).
+    """
+    repository_urls = [remove_auth_from_url(r.rstrip("/")) for r in repository_urls]
+    primary_url = remove_auth_from_url(primary_url)
```
jin: nit: prefer creating new variables instead of reassignment

coursier.bzl
```diff
@@ -272,6 +272,35 @@ def _pinned_coursier_fetch_impl(repository_ctx):
...
+    host = host[host.find("@") + 1:]
+    new_url = "{}://{}".format(protocol, "/".join([host] + url_parts[1:]))
+    return new_url
+
+def infer_artifact_path_from_primary_and_repos(primary_url, repository_urls):
+    """Returns the artifact path inferred by comparing primary_url with urls in repository_urls.
+
+    Returns:
+        String of the artifact path used by maven to find a particular artifact. Does not have a leading slash (`/`).
+    """
```
jin: nit: could you add a couple of examples here?

justhecuke: Added one and bit more doc. Is this what you wanted?

coursier.bzl
justhecuke: Done.

Issue #186, opened 2019-07-09
Pinned artifacts does not work with basic auth


natros:
```
INFO: Call stack for the definition of repository 'com_fasterxml_jackson_jaxrs_jackson_jaxrs_base_2_9_9' which is a http_file (rule definition at /private/var/tmp/_bazel_fsousa/5eb111cf39e94839e993e7dc498f4d7c/external/bazel_tools/tools/build_defs/repo/http.bzl:314:13):
 - /private/var/tmp/_bazel_fsousa/5eb111cf39e94839e993e7dc498f4d7c/external/maven/defs.bzl:83:5
 - /Users/fsousa/Projects/demo/demos/WORKSPACE:112:1
ERROR: An error occurred during the fetch of repository 'com_fasterxml_jackson_jaxrs_jackson_jaxrs_base_2_9_9':
   java.io.IOException: Error downloading [https://nexus%40devel.ipb.pt/nexus3/repository/maven-public/com/fasterxml/jackson/jaxrs/jackson-jaxrs-base/2.9.9/jackson-jaxrs-base-2.9.9.jar] to /private/var/tmp/_bazel_fsousa/5eb111cf39e94839e993e7dc498f4d7c/external/com_fasterxml_jackson_jaxrs_jackson_jaxrs_base_2_9_9/file/downloaded: Unknown host: nexus%40devel.ipb.pt
ERROR: /private/var/tmp/_bazel_fsousa/5eb111cf39e94839e993e7dc498f4d7c/external/maven/BUILD:892:1: no such package '@com_fasterxml_jackson_jaxrs_jackson_jaxrs_base_2_9_9//file': java.io.IOException: Error downloading [https://nexus%40devel.ipb.pt/nexus3/repository/maven-public/com/fasterxml/jackson/jaxrs/jackson-jaxrs-base/2.9.9/jackson-jaxrs-base-2.9.9.jar] to /private/var/tmp/_bazel_fsousa/5eb111cf39e94839e993e7dc498f4d7c/external/com_fasterxml_jackson_jaxrs_jackson_jaxrs_base_2_9_9/file/downloaded: Unknown host: nexus%40devel.ipb.pt and referenced by '@maven//:com_fasterxml_jackson_jaxrs_jackson_jaxrs_base_2_9_9_extension'
```

```json
 {
                "coord": "com.fasterxml.jackson.jaxrs:jackson-jaxrs-base:2.9.9",
                "dependencies": [
[...]

PR #306, opened 2019-12-04
Automatically generate a netrc file according to basic http auth in url

justhecuke:
For every mirror url in a pinned json, try to extract netrc credentials. Then, write those credentials into a file that every generated http_file for the various maven artifacts will use.

Alternative way of implementing would be by allowing you to pass a pre-made netrc and that can be added in later if we should choose to do so, although it may be somewhat difficult to do so due to there being at least three different repositories in play making it difficult to use raw paths.

With this change, basic http auth will appear to be supported even though it technically isn't since we are translating basic http auth into netrc auth with all of netrc's shortcomings.

In the majority of cases I expect for this to be enough, but for complex auth schemes we may need something in addition.

Fixes https://github.com/bazelbuild/rules_jvm_external/issues/186

### DIFF
--- a/coursier.bzl
+++ b/coursier.bzl
@@ -131,6 +131,72 @@ def _compute_dependency_tree_signature(artifacts):
         signature_inputs.append(":".join(artifact_group))
     return hash(repr(sorted(signature_inputs)))
 
+def extract_netrc_from_auth_url(url):
+    """Return a dict showing the netrc machine, login, and password extracted from a url.
+
+    Returns:
+        A dict that is empty if there were no credentials in the url.
+        A dict that has three keys -- machine, login, password -- with their respective values. These values should be
+        what is needed for the netrc entry of the same name except for password whose value may be empty meaning that
+        there is no password for that login.
+    """
+    if "@" not in url:
+        return {}
+    protocol, url_parts = split_url(url)
+    login_password_host = url_parts[0]
+    if "@" not in login_password_host:
+        return {}
+    login_password, host = login_password_host.rsplit("@", 1)
+    login_password_split = login_password.split(":", 1)
+    login = login_password_split[0]
+    # If password is not provided, then this will be a 1-length split
+    if len(login_password_split) < 2:
+        password = None
+    else:
+        password = login_password_split[1]
+    if not host:
+        fail("Got a blank host from: {}".format(url))
+    if not login:
+        fail("Got a blank login from: {}".format(url))
+    # Do not fail for blank password since that is sometimes a thing
+    return {
+        "machine": host,
+        "login": login,
+        "password": password,
+    }
+
+def add_netrc_entries_from_mirror_urls(netrc_entries, mirror_urls):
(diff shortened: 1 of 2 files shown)


justhecuke:
Hmm, seems like I need to fix this for bazel 25.3.

jin:
Wow, awesome! I'll make a deeper review ASAP, but feel free to ignore the failing tests before 0.27.2, since netrc wasn't added before then. It's about time we release a new major version that only works with 1.x.

jin:
Could you please rebase to https://github.com/bazelbuild/rules_jvm_external/commit/a62917b62833aa5d5ac5e1451fdb0f29d92545d5?

joprice:
It seems that this broke the default behavior where the netrc in the home directory is used https://github.com/bazelbuild/bazel/blob/e9bef0cf46a71039e45f797c98acc201e13ae602/tools/build_defs/repo/http.bzl#L52. 

Filtering out empty netrc dicts works for me:  https://github.com/bazelbuild/rules_jvm_external/compare/master...joprice:emptyNetrc, but I can image someone might expect the two sources to be merged instead.

justhecuke:
> It seems that this broke the default behavior where the netrc in the home directory is used https://github.com/bazelbuild/bazel/blob/e9bef0cf46a71039e45f797c98acc201e13ae602/tools/build_defs/repo/http.bzl#L52.
> 
> Filtering out empty netrc dicts works for me: [master...joprice:emptyNetrc](https://github.com/bazelbuild/rules_jvm_external/compare/master...joprice:emptyNetrc), but I can image someone might expect the two sources to be merged instead.

That would make a lot of sense. Sorry for missing this use-case. My team is leery of using non-repo netrc because it breaks hermeticity. I'll see what I can do to fix.

joprice:
I was curious about the workflow that this supports. Do you check credentials into the WORKSPACE file? Our use of netrc was to avoid doing that.

jin:
This adds support for the inlined auth params, and technically the auth params don't need to be in the WORKSPACE -- they can be loaded from a .bzl that's not checked in. 

Since Bazel supports reading ~/.netrc, we should find a way to combine both files.

justhecuke:
We have actually checked-in our credentials but I'm led to believe that this is more for usage-tracking rather than real security. The server we are authing into is in our VPN and is mostly a cache for other maven repos to decrease latency and increase bandwidth.

We also use a generated `.bzl` that we create in a bazel wrapper that is used by the WORKSPACE so that we can keep different systems in sync. We are currently moving to bazel and are concurrently supporting multiple build systems.

joprice:
I'm not a heavy user of netrc, only found it useful recently because of tools like curl supporting it. I can try the non-checked in bazel file. And I like the idea of having some bootstrap task to generate it. Would this be the encouraged approach either way so that changing credentials invalidates the existing cache?

justhecuke:
@joprice https://github.com/bazelbuild/rules_jvm_external/pull/319

Is this something that you can try to see if it works for you? I haven't set up a local test yet and I wanted to make sure that your use-case was addressed.

PR #313, opened 2019-12-09
Improve artifact hashing performance

jin:
(creating this PR on behalf of @joprice)

3.0 introduced a simple Java-based hasher to drop the dependency on Python, but the JVM startup caused a huge performance regression. This fixes it.


joprice:
@jin I was going to put up another one with a unit test and remove the change I had in there two handle double-symlinking (I reported that in another issue and not sure that my workaround is ideal).

jin:
Yep, sorry, we'll take a look at yours!

PR #314, opened 2019-12-09
improve artifact hashing performance

joprice:


coursier.bzl
```diff
@@ -552,15 +556,25 @@ def _coursier_fetch_impl(repository_ctx):
...
+
+    exec_result = repository_ctx.execute(hasher_command)
+    if exec_result.return_code != 0:
+        fail("Error while obtaining the sha256 checksums: " + exec_result.stderr)
+
+    shas = {}
+    for line in exec_result.stdout.splitlines():
+      parts = line.split(" ")
+      path = repository_ctx.path(parts[1])
+      shas[path] = parts[0]
```
jin: use `artifact["coord"]` as the key? that should solve the CI failure.

jin:
and thank you so much for this!

coursier.bzl
joprice: I didn't realize path wasn't hashable on some versions. How do you suggest going about that for coord? Pass a series of concatenations of file and coord to the java process like `coord:file coord:file`, then return `coord:sha coord:sha`?

joprice:
I'll remove the mods file. It was leftover from a series of optimization attempts I made using graal, newer jdk, java modules, etc to see if I could shave off more of the startup time. They all ended up being about the same or in the case of graal, worse.

coursier.bzl
jin: Hmm, you could also cast path to `str()`. Should be a fairly fast opertation.

coursier.bzl
```diff
@@ -508,16 +519,17 @@ def _coursier_fetch_impl(repository_ctx):
         primary_url_parts = []
         filepath_parts = artifact["file"].split("/")
         protocol = None
+
         # Only support http/https transports
         for part in filepath_parts:
             if part == "http" or part == "https":
-                    protocol = part
+                protocol = part
```
jin: I believe you need to rebase this part on master to add `break` after this line.

coursier.bzl
joprice: Ok I'll try that first

PR #319, opened 2019-12-10
Pre-pend HOME netrc to generated netrc for inline http auth

justhecuke:
http_file will, by default, take in the $HOME/.netrc file to use for authentication.

This was broken by the earlier attempt to auto-generate netrc as a workaround for a lack of
inline http auth in http_file.

We now prepend the HOME netrc to the generated netrc file so that the HOME netrc serves as a
base which will get overlaid with the credentials in mirror_urls.

### DIFF
--- a/coursier.bzl
+++ b/coursier.bzl
@@ -206,8 +206,8 @@ def get_home_netrc_contents(repository_ctx):
     if "HOME" in repository_ctx.os.environ:
         if not repository_ctx.os.name.startswith("windows"):
             netrcfile = "%s/.netrc" % (repository_ctx.os.environ["HOME"],)
-            if repository_ctx.execute(["test", "-f", netrcfile]).return_code == 0:
-                return repository_ctx.read(netrcfile)
+            if repository_ctx.which("test") and repository_ctx.execute(["test", "-f", netrcfile]).return_code == 0:
+                    return repository_ctx.read(netrcfile)
     return ""
 
 def _pinned_coursier_fetch_impl(repository_ctx):
--- a/coursier.bzl
+++ b/coursier.bzl
@@ -302,7 +302,7 @@ def _pinned_coursier_fetch_impl(repository_ctx):
             http_files.append("    )")
     repository_ctx.file("defs.bzl", "\n".join(http_files), executable = False)
     repository_ctx.file("netrc", "\n".join(
-        get_netrc_lines_from_entries(netrc_entries) + get_home_netrc_contents(repository_ctx).splitlines()),
+        get_home_netrc_contents(repository_ctx).splitlines() + get_netrc_lines_from_entries(netrc_entries)),
         executable = False)
 
     repository_ctx.report_progress("Generating BUILD targets..")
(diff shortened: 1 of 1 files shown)


coursier.bzl
```diff
@@ -200,6 +200,16 @@ def get_netrc_lines_from_entries(netrc_entries):
                 netrc_lines.append("password {}".format(password))
     return netrc_lines
 
+def get_home_netrc_contents(repository_ctx):
+    # Copied with a ctx -> repository_ctx rename from tools/build_defs/repo/http.bzl's _get_auth.
+    # Need to keep updated with improvements in source since we cannot load private methods.
+    if "HOME" in repository_ctx.os.environ:
+        if not repository_ctx.os.name.startswith("windows"):
+            netrcfile = "%s/.netrc" % (repository_ctx.os.environ["HOME"],)
+            if repository_ctx.execute(["test", "-f", netrcfile]).return_code == 0:
```
jin: use `repository_ctx.which`

joprice:
I'll trying it out now!

joprice:
Looks like this works for me

justhecuke:
@jin I saw an email with a comment from you but I can't seem to find it in the GitHub UI. I don't think I can use `repository_ctx.which` for the netrc file (indeed I tried but it doesn't find ".netrc" and doesn't handle paths) but it does work for "test".

Is that what you were thinking of?

coursier.bzl
```diff
@@ -200,6 +200,16 @@ def get_netrc_lines_from_entries(netrc_entries):
                 netrc_lines.append("password {}".format(password))
     return netrc_lines
 
+def get_home_netrc_contents(repository_ctx):
+    # Copied with a ctx -> repository_ctx rename from tools/build_defs/repo/http.bzl's _get_auth.
+    # Need to keep updated with improvements in source since we cannot load private methods.
```
jin: please apply buildifier on this file. a couple of indents look off.

coursier.bzl
```diff
@@ -291,7 +301,9 @@ def _pinned_coursier_fetch_impl(repository_ctx):
                 http_files.append("        urls = [\"%s\"]," % artifact["url"])
             http_files.append("    )")
     repository_ctx.file("defs.bzl", "\n".join(http_files), executable = False)
-    repository_ctx.file("netrc", "\n".join(get_netrc_lines_from_entries(netrc_entries)), executable = False)
+    repository_ctx.file("netrc", "\n".join(
+        get_home_netrc_contents(repository_ctx).splitlines() + get_netrc_lines_from_entries(netrc_entries)),
```
jin: can you add a comment that netrc entries are order-dependent, and we respect the HOME netrc contents over the inlined auth info?

coursier.bzl
justhecuke: Done. FYI: running buildifier on the whole repo creates many changes. Would it be OK to do a buildifier run on the whole repo and commit that? I know that the rules suggest not doing that.

jin: Thanks. Yeah, I'm OK with a formatting PR.

Issue #311, opened 2019-12-06
"Could not create symlink" when using use_unsafe_shared_cache = True


joprice:
When activating the `use_unsafe_shared_cache` setting, I see the error below:

```
'@scala_2_11//': Traceback (most recent call last):
	File "/private/var/tmp/_bazel_adamjohnson/09e8271e29391b7bad9dc3e7f8bde013/external/rules_jvm_external/coursier.bzl", line 503
		artifact.update(<1 more arguments>)
	File "/private/var/tmp/_bazel_adamjohnson/09e8271e29391b7bad9dc3e7f8bde013/external/rules_jvm_external/coursier.bzl", line 503, in artifact.update
		_relativize_and_symlink_file(<2 more arguments>)
	File "/private/var/tmp/_bazel_adamjohnson/09e8271e29391b7bad9dc3e7f8bde013/external/rules_jvm_external/coursier.bzl", line 69, in _relativize_and_symlink_file
		repository_ctx.symlink(<2 more arguments>)
java.io.IOException: Could not create symlink from /Users/adamjohnson/Library/Caches/Coursier/v1/http/localhost%3A8080/iheart/amp-bazel-external/com/fasterxml/jackson/core/jackson-databind/2.7.8/jackson-databind-2.7.8.jar to /private/var/tmp/_bazel_adamjohnson/09e8271e29391b7bad9dc3e7f8bde013/external/scala_2_11/v1/http/localhost%3A8080/iheart/amp-bazel-external/com/fasterxml/jackson/core/jackson-databind/2.7.8/jackson-databind-2.7.8.jar: /private/var/tmp/_bazel_adamjohnson/09e8271e29391b7bad9dc3e7f8bde013/external/scala_2_11/v1/http/localhost%3A8080/iheart/amp-bazel-external/com/fasterxml/jackson/core/jackson-databind/2.7.8/jackson-databind-2.7.8.jar (File exists)
```

I have tried deleting the cache and re-running and get the same error.

joprice:
This seems to be caused by the dependency tree having dupes on https://github.com/bazelbuild/rules_jvm_external/blob/0e365338b26ff9879ab67683730f50a184b3f3be/coursier.bzl#L493.

jin:
Do you have a `maven_install` I can reproduce this on?

joprice:
Not at hand, but I can try to minimize one that doesn't contain internal repos.

gergelyfabian:
[...]

PR #325, opened 2019-12-13
Added option to use safe shared cache

gergelyfabian:
Comparing to unsafe shared cache we do not do symlinking to Bazel's cache
from Coursier's cache but copy, to avoid possible problems if coursier's
cache change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -311,6 +311,32 @@ The default value of `use_unsafe_shared_cache` is `False`. This means that Bazel
 will create independent caches for each `maven_install` repository, located at
 `$(bazel info output_base)/external/@repository_name/v1`.
 
+### Using a safe persistent artifact cache
+
+To download artifacts into a shared and persistent directory in your home
+directory (while keeping Bazel caching safe), set `use_safe_shared_cache = True` in `maven_install`.
+
+```python
+maven_install(
+    artifacts = [
+        # ...
+    ],
+    repositories = [
+        # ...
+    ],
+    use_safe_shared_cache = True,
+)
+```
+
+
+This is considered safe as it copies the downloaded artifacts to Bazel's cache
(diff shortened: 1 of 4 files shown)


gergelyfabian:
The intention with this PR was to enable shared caching between Coursier executions (artifact pinnings), but keep Bazel's caching safe. I hope my understanding is correct, that downloading to Coursier and then copying over to Bazel's cache should be the same from Bazel's viewpoint as if the artifacts were downloaded to Bazel's cache right away.

If my understanding is incorrect, or you'd like to propose a change for this option (e.g. the name), please let me know.

jin:
Hey, thanks for this PR!

This does not address the "unsafe" part of the term, where bad actors can make modifications to the artifacts in the shared cache and Coursier wouldn't know, and Bazel will naively copy/symlink it to be used. 

This is why we encourage the usage of artifact pinning wherever possible, where we lock down the exact sha256 checksums of the artifacts.

gergelyfabian:
Ok, I got your point about copy also being unsafe (because someone may modify the Courier cache in between before we copy it to Bazel cache). So maybe I could rework this PR to become a fix for #311 - that was the original trigger for creating this change (it removes that problem for my case)?

Related to that issue, what should happen if there are duplicate artifacts? For copying it's not an issue, but symlinking fails. Should we copy, or do some additional steps with symlinking (e.g. bypass symlinking if the symlink is already there)?

Solving #311 here would be satisfactory for me and then I do not need a new option for symlinking/copying for unsafe shared caching.

Please suggest, and I'd be more than happy to rework.

gergelyfabian:
The solution for use_unsafe_shared_cache in https://github.com/bazelbuild/rules_jvm_external/pull/327 works for my case, so I'm closing this PR (as that is a better option).

PR #344, opened 2019-12-26
Add Jetifier support

jin:


WORKSPACE
```diff
@@ -290,6 +290,19 @@ maven_install(
...
 
+maven_install(
+    name = "jetify_test",
+    artifacts = [
+        "com.google.guava:guava:27.0-jre",
+        "com.android.support:appcompat-v7:28.0.0"
+    ],
+    repositories = [
+        "https://jcenter.bintray.com/",
+        "https://maven.google.com",
```
artem-zinnatullin: Let's reorder and hit maven.google.com first? JCenter is not very reliable 😬

jin: I don't think this really matters (and it's just for tests), we'll need to hit jcenter for guava anyway.

jin:
Yeah, there's really no need to jetify everything, but we can leave that for a follow up.

private/rules/jetifier.bzl
```diff
@@ -0,0 +1,59 @@
...
+def _jetify_impl(ctx):
+    srcs = ctx.attr.srcs
+    outfiles = []
+    for src in srcs:
+        for artifact in src.files.to_list():
+            jetified_outfile = ctx.actions.declare_file("jetified_" + artifact.basename, sibling = artifact)
+            jetify_args = ctx.actions.args()
+            jetify_args.add_all(["-l", "error"])
+            jetify_args.add_all(["-o", jetified_outfile])
+            jetify_args.add_all(["-i", artifact])
```
arturdryomov: WDYT about this? Less list allocations.
```starlark
jetify_args.add("-l", "error")
jetify_args.add("-o", jetified_outfile)
jetify_args.add("-i", artifact)
```

jin: nice catch! thanks.

PR #348, opened 2019-12-28
Add support for jetify_include_patterns

artem-zinnatullin:
Allows user to include only certain artifacts for jetification. 

Since Starlark doesn't support Regex, I decided to allow few Starlark supported string comparison patterns applied against simple coords `groupId:artifactId`:

- `equals(x)`
- `contains(x)`
- `startswith(x)`
- `endswith(x)`

This is not complete yet (but already works!), I'd like to get feedback on the approach first, hope you'll find it let's say "interesting" 😅

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -321,6 +321,9 @@ maven_install(
         "https://maven.google.com",
     ],
     jetify = True,
+    jetify_exclude_patterns = [
+        "startswith(com.google)",
+    ],
 )
 
 RULES_KOTLIN_VERSION = "8ca948548159f288450516a09248dcfb9e957804"
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -321,8 +321,9 @@ maven_install(
         "https://maven.google.com",
     ],
     jetify = True,
-    jetify_exclude_patterns = [
-        "startswith(com.google)",
+    jetify_include_patterns = [
+        "startswith(com.google:)",
+        "startswith(com.android.support:)",
     ],
 )
 
--- a/private/dependency_tree_parser.bzl
+++ b/private/dependency_tree_parser.bzl
@@ -119,8 +119,9 @@ def _generate_imports(repository_ctx, dep_tree, explicit_artifacts, neverlink_ar
                 import_rule = "aar_import"
             else:
                 fail("Unsupported packaging type: " + packaging)
-            if repository_ctx.attr.jetify:
+            if repository_ctx.attr.jetify and not _coord_matches_exclude_pattern(simple_coord, repository_ctx.attr.jetify_exclude_patterns):
                 import_rule = "jetify_" + import_rule
+
             target_import_string = [import_rule + "("]
 
             # 2. Generate the target label.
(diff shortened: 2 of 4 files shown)


artem-zinnatullin:
@arturdryomov please also take a look

artem-zinnatullin:
Actually, a reverse variant might be more feasible — `jetify_include_patterns` as most of the libraries already migrated to Androidx

I'll update the code

artem-zinnatullin:
Ok, updated with reverse logic, now it's `jetify_include_patterns`

private/dependency_tree_parser.bzl
```diff
@@ -330,3 +331,39 @@ def _generate_imports(repository_ctx, dep_tree, explicit_artifacts, neverlink_ar
 parser = struct(
     generate_imports = _generate_imports,
 )
+
+def _simple_coord_matches_pattern(simple_coord, patterns):
+    for raw_pattern in patterns:
+        pattern_and_value=_parse_raw_pattern(raw_pattern)
```
justhecuke: Need spaces around `=`

private/dependency_tree_parser.bzl
```diff
@@ -330,3 +331,39 @@ def _generate_imports(repository_ctx, dep_tree, explicit_artifacts, neverlink_ar
...
+
+        pattern = pattern_and_value[0]
+        token = pattern_and_value[1]
+
+        if (
```
justhecuke: You can use `any` to encapsulate an or-chain. Up to you.

private/dependency_tree_parser.bzl
```diff
@@ -330,3 +331,39 @@ def _generate_imports(repository_ctx, dep_tree, explicit_artifacts, neverlink_ar
...
+                (pattern == "equals" and simple_coord == token)
+                or
+                (pattern == "contains" and token in simple_coord)
```
justhecuke: Is simple_coord a list? If so, make it a dict or something so that the `in` operator is o(1) and not o(n).

private/dependency_tree_parser.bzl
```diff
@@ -119,8 +119,9 @@ def _generate_imports(repository_ctx, dep_tree, explicit_artifacts, neverlink_ar
                 import_rule = "aar_import"
             else:
                 fail("Unsupported packaging type: " + packaging)
-            if repository_ctx.attr.jetify:
+            if repository_ctx.attr.jetify and _simple_coord_matches_pattern(simple_coord, repository_ctx.attr.jetify_include_patterns):
```
justhecuke: I can't see context on a phone. Is this run in a loop? Of so then we may need to cache things in the function so we don't recalculate const things.

WORKSPACE
```diff
@@ -321,6 +321,10 @@ maven_install(
         "https://maven.google.com",
     ],
     jetify = True,
+    jetify_include_patterns = [
+        "startswith(com.google:)",
+        "startswith(com.android.support:)",
```
arturdryomov: While this is a clever trick, I think a Bazel-idiomatic approach would be listing labels explicitly:
```starlark
jetify_include = [
    "@maven//:junit_junit",
]
```

artem-zinnatullin: That's a bit too much when you need to jetify entire groupId for example, but I don't disagree really

The reason why I went with pattern list is because I found it very nice when we were using OkBuck https://github.com/uber/okbuck/blob/a0eb9e9daf27a17715d238fd2ed9fd24c5ad74f9/buildSrc/src/main/java/com/uber/okbuck/extension/JetifierExtension.java#L43

jin:
I'm not a fan of bringing the starlark dsl into this. Two thoughts:

1) Can we just encode a whitelist of artifacts to jetify, and not exposing another API for this? Shamelessly ~~stealing~~ inspired from https://github.com/square/bazel_maven_repository/blob/52cb857ded93912a3ee419a0653a7cd351a63a9d/maven/jetifier.bzl#L4-L108. 

2) If we have to have an API, I prefer using `*` as a glob. `*` means every artifact, `groupid:*` means everything that starts with `groupid`, and so on.

arturdryomov:
Following @jin motto I think Jetifier should die soon and it is not even worth to support it 😉Unfortunately it is still alive (OH HAI Facebook SDK). The proposed hardcode solution sounds great actually, it can be refined further when the actual need arises.

artem-zinnatullin:
Alright, I'll redo to make it a `jetify_include_list` 👍

jin:
Not as an user configurable API, right? that is, hardcode the list of artifact coordinates to jetify.

artem-zinnatullin:
Mm I was thinking of user configurable, there are thousands of libraries of
different versions some of which are updated to androidx and some are not,
we can't nor prob want to track them all :)

artem-zinnatullin:
Closing this one in favor of #356

artem-zinnatullin:
Thanks for reviews and discussions, I agree that it's better to have explicit list 👍

Issue #353, opened 2020-01-02
Stamped-prefix jars do not inter-operate with mvn and a pom_file rule


mmikitka:
The `@bazel_common//tools/maven:pom_file.bzl` pom_file rule does not play nicely with jvm_import deps prefixed with `stamped_`.

When I make the following local modification to private/rules/jvm_import.bzl, Maven (mvn) is able to find the dependencies.

```
--- a/private/rules/jvm_import.bzl
+++ b/private/rules/jvm_import.bzl
@@ -45,11 +45,11 @@ def _jvm_import_impl(ctx):
 
     return [
         DefaultInfo(
-            files = depset([outjar]),
+            files = depset([injar]),
         ),
         JavaInfo(
-            compile_jar = outjar,
-            output_jar = outjar,
+            compile_jar = injar,
+            output_jar = injar,
             source_jar = ctx.file.srcjar,
             deps = [
                 dep[JavaInfo]
```

Please advise on how to use the `pom_file` rule with `maven_install`-managed dependencies, in particular the `stamped_`-prefixed generated jar files.

mmikitka:
My genrule-wrapped mvn command looks something like this:

`/usr/bin/mvn -X --offline --fail-fast -f $(execpath :pom) -Dmaven.repo.local=$(BINDIR)/external/maven/v1/https/repo.maven.apache.org/maven2`

jin:
What does "does not play nicely" mean? Could you give me something to reproduce this with, please?

mmikitka:
I created a branch that reproduces my problem. Please checkout the following branch: https://github.com/mmikitka/rules_jvm_external/tree/351-json-deps-jvm-import-failure

After checking out the branch, do the following:
```
bazel run @unpinned_maven_genrule_testing//:pin
bazel build //tests/integration/maven:mvn
```

You will see the following Maven error message
```
[ERROR] Plugin org.apache.maven.plugins:maven-resources-plugin:2.6 or one of its dependencies could not be resolved: Cannot access central (https://repo.maven.apache.org/maven2) in offline mode and the artifact org.apache.maven.plugins:maven-resources-plugin:jar:2.6 has not been downloaded from it before. -> [Help 1]
[...]

PR #354, opened 2020-01-03
Include the input jar in the outputs

mmikitka:
Issue https://github.com/bazelbuild/rules_jvm_external/issues/353


jin:
No, this is not the solution, the jvm_import rule should only return one jar in its default filegroup output.

PR #355, opened 2020-01-06
Add a build setting `--//setting:stamp_manifest` to disable Target-Label stamping.

jin:
There are a few reports of jar stamping being non-deterministic. This PR adds an ability to enable or disable the `Target-Label` stamping by passing the flag `--@rules_jvm_external//setting:stamp_manifest`.


lukaciko:
I believe #738 fixed the non-determinism problem. @jin @shs96c  perhaps `stamp_manifest` can now be removed?

PR #356, opened 2020-01-08
Add jetify_include_list to jetify only specific artifacts

artem-zinnatullin:
Allows user to include only certain artifacts for jetification.

This is not a breaking change as it keeps the default behavior intact, if user only sets `jetify = True` — all artifacts are jetified.

If user sets `jetify = True` **and** `jetify_include_list = [ … ]` then only included artifacts are jetified.

This allows to improve build times by avoiding jetification of artifacts that don't need it and avoid Jetifier bugs when it corrupts an artifact (haven't seen that in a while, but it was the case in past with some libraries).

This is a replacement for #348 as discussed in that PR.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -321,6 +321,9 @@ maven_install(
         "https://maven.google.com",
     ],
     jetify = True,
+    jetify_include_list = [
+        "com.android.support:appcompat-v7",
+    ]
 )
 
 RULES_KOTLIN_VERSION = "8ca948548159f288450516a09248dcfb9e957804"
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -311,7 +311,7 @@ maven_install(
 )
 
 maven_install(
-    name = "jetify_test",
+    name = "jetify_all_test",
     artifacts = [
         "com.google.guava:guava:27.0-jre",
         "com.android.support:appcompat-v7:28.0.0"
@@ -321,9 +321,23 @@ maven_install(
         "https://maven.google.com",
     ],
     jetify = True,
+)
+
+maven_install(
+    name = "jetify_include_list_test",
+    artifacts = [
+        "com.google.guava:guava:27.0-jre",
+        "com.android.support:appcompat-v7:28.0.0",
+        "com.android.support:swiperefreshlayout:28.0.0",
+    ],
+    repositories = [
+        "https://jcenter.bintray.com/",
+        "https://maven.google.com",
+    ],
+    jetify = True,
     jetify_include_list = [
         "com.android.support:appcompat-v7",
-    ]
+    ],
 )
 
 RULES_KOTLIN_VERSION = "8ca948548159f288450516a09248dcfb9e957804"
--- a/coursier.bzl
+++ b/coursier.bzl
@@ -781,7 +781,7 @@ pinned_coursier_fetch = repository_rule(
             """,
             default = False,
         ),
-        "jetify": attr.bool(doc = "Runs the AndroidX Jetifier tool on all artifacts.", default = False),
+        "jetify": attr.bool(doc = "Runs the AndroidX [Jetifier](https://developer.android.com/studio/command-line/jetifier) tool on artifacts specified in jetify_include_list. If jetify_include_list is not specified, run Jetifier on all artifacts.", default = False),
         "jetify_include_list": attr.string_list(doc = "List of artifacts that need to be jetified in `groupId:artifactId` format. By default all artifacts are jetified if `jetify` is set to True.", default = JETIFY_INCLUDE_LIST_JETIFY_ALL),
     },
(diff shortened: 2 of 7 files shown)


artem-zinnatullin:
cc @justhecuke @arturdryomov

private/dependency_tree_parser.bzl
```diff
@@ -19,6 +19,8 @@ into target declarations (jvm_import) for the final @maven//:BUILD file.
 
 load("//:private/coursier_utilities.bzl", "SUPPORTED_PACKAGING_TYPES", "escape", "strip_packaging_and_classifier", "strip_packaging_and_classifier_and_version")
 
+JETIFY_INCLUDE_LIST_JETIFY_ALL = ["_PRIVATE_RULES_JVM_EXTERNAL_JETIFY_ALL_"]
+
```
jin: use `*`? 

this default value will show up in the generated `docs/api.md`.

coursier.bzl
```diff
@@ -815,7 +822,8 @@ coursier_fetch = repository_rule(
             default = False,
         ),
         "resolve_timeout": attr.int(default = 600),
-        "jetify": attr.bool(doc = "Runs the AndroidX Jetifier tool on all artifacts.", default = False)
+        "jetify": attr.bool(doc = "Runs the AndroidX Jetifier tool on all artifacts.", default = False),
```
jin:
```suggestion
        "jetify": attr.bool(doc = "Runs the AndroidX Jetifier tool on artifacts specified in jetify_include_list. If jetify_include_list is not specified, run jetifier on all artifacts.", default = False),
```

here and above.

WORKSPACE
```diff
@@ -311,7 +311,7 @@ maven_install(
 )
 
 maven_install(
-    name = "jetify_test",
+    name = "jetify_all_test",
```
arturdryomov: `jetify_without_include_list_test`? Anything is good really.

coursier.bzl
```diff
@@ -775,7 +781,8 @@ pinned_coursier_fetch = repository_rule(
             """,
             default = False,
         ),
-        "jetify": attr.bool(doc = "Runs the AndroidX Jetifier tool on all artifacts.", default = False)
+        "jetify": attr.bool(doc = "Runs the AndroidX Jetifier tool on all artifacts.", default = False),
+        "jetify_include_list": attr.string_list(doc = "List of artifacts that need to be jetified in `groupId:artifactId` format. By default all artifacts are jetified if `jetify` is set to True.", default = JETIFY_INCLUDE_LIST_JETIFY_ALL),
```
arturdryomov: Too bad it is impossible to use labels at this stage.

coursier.bzl
arturdryomov: Nit: use either `jetifier` or Jetifier in docs.

Also think about providing URL to the Jetifier page (Stardoc supports Markdown).

private/dependency_tree_parser.bzl
arturdryomov: Great suggestion. I’ve struggled to suggest something more meaningful than an empty list / a magic value.

private/dependency_tree_parser.bzl
```diff
@@ -78,6 +80,13 @@ def _generate_imports(repository_ctx, dep_tree, explicit_artifacts, neverlink_ar
                     if repository_ctx.attr.maven_install_json:
                         all_imports.append(_genrule_copy_artifact_from_http_file(artifact))
 
+    jetify_all = repository_ctx.attr.jetify and repository_ctx.attr.jetify_include_list == JETIFY_INCLUDE_LIST_JETIFY_ALL
+
+    # Write artifacts to dict to achieve O(1) lookup instead of O(n).
+    jetify_include_dict = {}
+    for jetify_include_artifact in repository_ctx.attr.jetify_include_list:
+        jetify_include_dict[jetify_include_artifact] = None
```
arturdryomov: Welp, not sure that creating a dict is cheaper than searching in a list but I’ll bite.

private/dependency_tree_parser.bzl
artem-zinnatullin: ah, totally agree, will do

WORKSPACE
artem-zinnatullin: I think _all captures the intention really good here

artem-zinnatullin:
Updated

arturdryomov:
:shipit:

jin:
We'd probably want to update the docs at https://github.com/bazelbuild/rules_jvm_external#jetifier too.

I can do that soon, or if you wanna pick it up that's fine too!

artem-zinnatullin:
Here you go https://github.com/bazelbuild/rules_jvm_external/pull/359

Issue #357, opened 2020-01-08
3.1 seems to have a limit to the number of artifacts it will hash


justhecuke:
> ERROR: An error occurred during the fetch of repository 'unpinned_maven':
>    Error while obtaining the sha256 checksums: java.io.IOException: Cannot run program "/var/tmp/_bazel_jwei/install/4f12cbbb3a294bfa314e578758559d54/process-wrapper" (in directory "/private/var/tmp/_bazel_jwei/3e5cafa66bb3fb562cc9f4a87160325d/external/unpinned_maven"): error=7, Argument list too long

I've verified that this does not happen in 3.0 but does happen in 3.1. I assume this is due to https://github.com/bazelbuild/rules_jvm_external/pull/314 but I haven't tried a local test to see if this is actually the case.

jin:
Looks like we need to batch this up.

cc @joprice

joprice:
That makes sense. I hadn't hit that for the number of deps I'm currently using. @justhecuke Thanks for the quick fix!

justhecuke:
@joprice Thanks for making the optimization in the first place. Our maven loads take about 20-30% of the time now.

PR #360, opened 2020-01-14
Fix hashing args to take in an argsfile

justhecuke:
Fixes https://github.com/bazelbuild/rules_jvm_external/issues/357

Adds special handling for a very special case in the Hasher where we read in files to hash in an argsfile if `--argsfile <file>` are the only args passed.


justhecuke:
Wanted to get feedback on the direction this is going. I'm in the process of writing tests, although I don't think it'll take long.

jin:
Yeah, this looks reasonable to me.

justhecuke:
@jin This is ready for review now.

private/tools/java/rules/jvm/external/HasherTest.java
```diff
@@ -56,13 +61,78 @@ public void sha256_multipleFiles() throws IOException, FileNotFoundException {
...
 
-    String result = Hasher.hashFiles(files);
+    String result = Hasher.hashFiles(Stream.of(files));
     String[] lines = result.split("\n");
     assertThat(lines, arrayWithSize(3));
 
-    checkLine(lines[0], "7f83b1657ff1fc53b92dc18148a1d65dfc2d4b1fa3d677284addd200126d9069", "test-1.file");
-    checkLine(lines[1], "334d016f755cd6dc58c53a86e183882f8ec14f52fb05345887c8a5edd42c87b7", "test-2.file");
-    checkLine(lines[2], "7f83b1657ff1fc53b92dc18148a1d65dfc2d4b1fa3d677284addd200126d9069", "test-3.file");
+    checkLine(lines[0], "7f83b1657ff1fc53b92dc18148a1d65dfc2d4b1fa3d677284addd200126d9069",
```
jin: could you run google-java-format on this?

private/tools/java/rules/jvm/external/Hasher.java
```diff
@@ -17,20 +17,38 @@
...
-    System.out.print(hashFiles(args));
+    Stream<String> fileStream = argsToStream(args);
+    System.out.print(hashFiles(fileStream));
   }
 
-  static String hashFiles(String[] files) {
-    return Stream.of(files)
+  static Stream<String> argsToStream(String[] args) throws IOException {
+    Stream<String> fileStream;
+
```
jin: since this hasher is only used for rules_jvm_external, I don't think there's a need to keep the original non-argsfile approach.

private/tools/java/rules/jvm/external/HasherTest.java
justhecuke: Done.

private/tools/java/rules/jvm/external/Hasher.java
justhecuke: Removed the non-argsfile approach and had it raise an IAE

PR #366, opened 2020-01-23
Preserver jar filename when stamping

darl:
After updating to version 3.1 (from 2.9) all external libraries in idea got a prefix `stamped_`

screenshot 2020-01-24 в 0 51 15

Its not a problem on its own.
But there is problem with idea plugin:
[It looks](https://github.com/bazelbuild/intellij/blob/c240371fbc3b3756b5faa2fa7f60130cb22327b2/scala/src/com/google/idea/blaze/scala/sync/BlazeScalaSyncPlugin.java#L78) for libraries which start with `scala-library` to configure scala sdk.

This fix will keep original jar names so they will be imported to idea with their original names

### DIFF
--- a/private/rules/jvm_import.bzl
+++ b/private/rules/jvm_import.bzl
@@ -21,7 +21,7 @@ def _jvm_import_impl(ctx):
         },
     )
 
-    outjar = ctx.actions.declare_file("stamped_" + injar.basename, sibling = injar)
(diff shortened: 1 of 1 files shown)


jin:
Seems like this change caused a test to break. Could you please take a look at that?

darl:
Didn't dig deep enough, but I think it caused by different path separator on windows.

Changed pr a little to keep jar in same directory.
Tests look green now.

jin:
Thanks!

Issue #369, opened 2020-02-05
The json packaging rules match artifacts with name json as well


cheister:
https://github.com/bazelbuild/rules_jvm_external/pull/352 matches the maven coordinate against `:json:` which can match the artifactId as well as the packaging type in the maven coordinate.  This breaks using the `org.json:json` artifact.

jin:
Fixed by #370

PR #370, opened 2020-02-05
Selecting json deps should only apply to dependencies that have packaging=json not artifactId=json

cheister:
This changes the logic for json coordinates so they match the packaging part of the coordinate only and not the artifactId part
https://github.com/bazelbuild/rules_jvm_external/issues/369


private/coursier_utilities.bzl
```diff
@@ -28,18 +28,49 @@ SUPPORTED_PACKAGING_TYPES = [
 ]
 
 def strip_packaging_and_classifier(coord):
-    # We add "pom" into _COURSIER_PACKAGING_TYPES here because "pom" is not a
```
cheister: I originally implemented this as
```
def strip_packaging_and_classifier(coord):
    # Strip packaging and classifier from the following maven coordinate formats
    # groupId:artifactId:version
    # groupId:artifactId:packaging:version
    # groupId:artifactId:packaging:classifier:version
    coordinates = coord.split(":")
    return ":".join([coordinates[0], coordinates[1], coordinates[-1]])
```
but stripping all classifiers and types causes problems for resolving some artifacts so I went back to the more selective removal of classifiers and package types

private/coursier_utilities.bzl
```diff
@@ -28,18 +28,49 @@ SUPPORTED_PACKAGING_TYPES = [
...
-        coord = coord.replace(":%s:" % packaging_type, ":")
-    for classifier_type in ["sources", "natives"]:
-        coord = coord.replace(":%s:" % classifier_type, ":")
+    # Strip some packaging and classifier values based on the following maven coordinate formats
+    # groupId:artifactId:version
+    # groupId:artifactId:packaging:version
+    # groupId:artifactId:packaging:classifier:version
+    coordinates = coord.split(":")
+    if len(coordinates) > 4:
+        if coordinates[3] in ["sources", "natives"]:
```
cheister: I think `natives` should just be `native` but I didn't want to change it in this PR.

private/coursier_utilities.bzl
```diff
@@ -28,18 +28,49 @@ SUPPORTED_PACKAGING_TYPES = [
...
+            coordinates.pop(3)
 
-    return coord
+    if len(coordinates) > 3:
+        # We add "pom" into _COURSIER_PACKAGING_TYPES here because "pom" is not a
```
jin: nit, while you're here: _COURSIER_PACKAGING_TYPES -> SUPPORTED_PACKAGING_TYPES

private/coursier_utilities.bzl
jin: SGTM

private/coursier_utilities.bzl
jin: SGTM

private/coursier_utilities.bzl
```diff
@@ -28,18 +28,49 @@ SUPPORTED_PACKAGING_TYPES = [
...
+        if coordinates[2] in SUPPORTED_PACKAGING_TYPES + ["pom"]:
+            coordinates.pop(2)
+
+    return ":".join(coordinates)
 
 def strip_packaging_and_classifier_and_version(coord):
     return ":".join(strip_packaging_and_classifier(coord).split(":")[:-1])
 
+# TODO: Should these methods be combined with _parse_maven_coordinate_string in specs.
+def get_packaging(coord):
```
jin: add unit tests?

cheister: Added

Issue #371, opened 2020-02-05
If you resolve more than one bom or pom file it will be discarded in the pinned json file because the filepath is null


cheister:
When a pom artifact or bom artifact is resolved by coursier it returns a null filepath.  The deduplication used the filepath to remove duplicate dependencies.  Since multiple pom artifacts all have null filepath only the first one will be add to the dependencies and all other will be removed by the deduplication.

Coursier doesn't support boms yet https://github.com/bazelbuild/rules_jvm_external/issues/229 but if they are brought in by other dependencies then dependency resolution shouldn't fail.

jin:
Fixed by #372, I believe.

PR #372, opened 2020-02-05
Don't remove duplicate dependencies that have null file paths

cheister:
Fix for https://github.com/bazelbuild/rules_jvm_external/issues/371 which keeps all the artifacts with "null" paths from being de-dupped out of existence.  This happens anytime you have multiple pom dependencies.

The real fix for boms would be getting https://github.com/coursier/coursier/issues/1390 working

I assumed that we wouldn't need to further de-dup the artifacts with null filepaths on maven coordinates since I assume coursier would have taken care of that.


cheister:
@jin Any ideas on what is wrong with CI?  It seems like the failures are not related to this change?

jin:
Yeah, it's an unrelated breakage with Android builds on Windows.

PR #376, opened 2020-02-13
reproducibly produce the stamped jar

kevingessner:
Since 1739774daaf7fb4f3fca5997d99e31d8af51542d, `jar` is used during stamping.  Unlike singlejar, it does not produce jars reproducibly because it embeds the current timestamp.  This change overrides that timestamp (thanks to jars being zips in disguise) so the same outjar is always created from a given injar.

### DIFF
--- a/private/rules/jvm_import.bzl
+++ b/private/rules/jvm_import.bzl
@@ -33,8 +33,16 @@ def _jvm_import_impl(ctx):
             "chmod +w {output_jar}".format(output_jar = outjar.path),
             # If the jar is signed do not modify the manifest because it will
             # make the signature invalid. Otherwise append the Target-Label
-            # manifest attribute using `jar umf`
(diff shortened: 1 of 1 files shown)


jin:
Hi @kevingessner, thanks for this PR! It looks like the Windows tests are failing from timeout because the genrule doesn't terminate. 

I'm thinking of disabling this feature and making a release, as I've also received reports that this is breaking remote caching and execution. After that, we can reinvestigate a more principled way of embedding the `Target-Label` entry (perhaps using a custom tool, like we did for hashing in `//private/tools`.

kevingessner:
Thanks for looking @jin.  Moving this out of shell entirely does seem like a better move!

JaredNeil:
For reference, Bazel has the a tool designed for creating reproducible jars.
https://github.com/bazelbuild/bazel/blob/master/src/java_tools/buildjar/java/com/google/devtools/build/buildjar/jarhelper/JarHelper.java
The https://github.com/higherkindness/rules_scala project uses a copy of it to accomplish this same jar stamping.

PR #379, opened 2020-02-17
Add attribute to override BUILD file template

Yannic:


jin:
I'd prefer limiting the API surface as much as possible. This exposes the internal implementation of the repository rule. 

What use cases are you thinking of?

PR #385, opened 2020-02-20
Enable parsing COURSIER_OPTS env var to add parameters to coursier invocation

gergelyfabian:
This enables e.g. increasing available memory settings for couriser (-Xmx).

### DIFF
--- a/coursier.bzl
+++ b/coursier.bzl
@@ -77,19 +77,21 @@ def _relativize_and_symlink_file(repository_ctx, absolute_path):
 # location of `java`.
 def _generate_java_jar_command(repository_ctx, jar_path):
     java_home = repository_ctx.os.environ.get("JAVA_HOME")
+    coursier_opts = repository_ctx.os.environ.get("COURSIER_OPTS", "")
+    coursier_opts = coursier_opts.split(" ") if len(coursier_opts) > 0 else []
 
     if java_home != None:
         # https://github.com/coursier/coursier/blob/master/doc/FORMER-README.md#how-can-the-launcher-be-run-on-windows-or-manually-with-the-java-program
         # The -noverify option seems to be required after the proguarding step
         # of the main JAR of coursier.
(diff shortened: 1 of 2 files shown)


gergelyfabian:
This change is necessary for really big dependency trees. It can happen then, that couriser runs out of memory, which can be fixed by providing proper -Xmx and -Xms settings for it.

jin:
The environment variable name is misleading, this is more for the JVM running Coursier right?

gergelyfabian:
Yes, it is for the JVM running Coursier.
If you have a proposal for a better name I would be glad to fix it :)

Regards,

Gergely

Jin <notifications@github.com> ezt írta (időpont: 2020. febr. 20., Cs
17:49):

gergelyfabian:
I changed the name of the env var to COURSIER_OPTS. This is similar to SBT_OPTS or MAVEN_OPTS for SBT and Maven respectively (and used for those build tools to enable passing in JVM flags).

coursier.bzl
```diff
@@ -77,19 +77,21 @@ def _relativize_and_symlink_file(repository_ctx, absolute_path):
 # location of `java`.
 def _generate_java_jar_command(repository_ctx, jar_path):
     java_home = repository_ctx.os.environ.get("JAVA_HOME")
+    coursier_opts = repository_ctx.os.environ.get("COURSIER_OPTS")
```
jin: use the `default` param: https://docs.bazel.build/versions/2.1.0/skylark/lib/dict.html#get

gergelyfabian: Changed default value to empty string. I still prefer returning in the next line an empty list if this option wasn't set, so I applied len(...) > 0.

gergelyfabian:
> COURSIER_OPTS sounds good by convention, since it's the same for SBT and Maven. Please also add documentation to the README.

Documentation to README added.

PR #386, opened 2020-02-22
Fix issue where a jetified maven artifact did not also jetify the deps

justhecuke:
There are two main changes in this PR:
- we now use directDependencies rather than the full transitive closure of maven dependencies for maven targets
- dependencies for jetified artifacts are redirected towards their jetified versions

This fixes an issue where a jetified maven artifact would still depend on un-jetified maven artifacts which led to a lot of problems with duplicate classes and resources since having both a jetfied and un-jetified android support library in your classpath is a very bad idea.


justhecuke:
This PR still needs some work because maven_install.json itself should be changed to provide the jetified deps because we need coursier to figure out the dep graph of the jetified dependencies.

The approach here of just changing in ingestion of maven_install.json isn't sufficient since it requires that you either explicitly add the jetified artifacts into your list of maven artifacts to install or that someone else explicitly depend on it in POM which is unreliable.

jin:
Thanks. I've been wanting to generate the BUILD file using directDependencies instead of dependencies, ever since Coursier added it.

We don't need Coursier to figure out the dep  graph of the dependencies. Since we now have the map of coordinate transformations, we can modify this code path (https://github.com/bazelbuild/rules_jvm_external/blob/2a2c0794cbbc255f2dc50acc483ce809a243d94b/coursier.bzl#L564-L628) to replace the Coursier-provided maven_install.json in-place, to reflect the Jetifier changes.

justhecuke:
> Thanks. I've been wanting to generate the BUILD file using directDependencies instead of dependencies, ever since Coursier added it.
> 
> We don't need Coursier to figure out the dep graph of the dependencies. Since we now have the map of coordinate transformations, we can modify this code path (
> 
> https://github.com/bazelbuild/rules_jvm_external/blob/2a2c0794cbbc255f2dc50acc483ce809a243d94b/coursier.bzl#L564-L628
> 
> ) to replace the Coursier-provided maven_install.json in-place, to reflect the Jetifier changes.

I'm not too sure this works because we'd also need to add in the entries for androidx deps and transitive deps and version conflict resolution.

justhecuke:
> > Thanks. I've been wanting to generate the BUILD file using directDependencies instead of dependencies, ever since Coursier added it.
> > We don't need Coursier to figure out the dep graph of the dependencies. Since we now have the map of coordinate transformations, we can modify this code path (
> > https://github.com/bazelbuild/rules_jvm_external/blob/2a2c0794cbbc255f2dc50acc483ce809a243d94b/coursier.bzl#L564-L628
> > 
> > ) to replace the Coursier-provided maven_install.json in-place, to reflect the Jetifier changes.
> 
> I'm not too sure this works because we'd also need to add in the entries for androidx deps and transitive deps and version conflict resolution.

@jin Would you be OK with a 2-pass coursier fetch where we add an extra fetch to grab the added androidx dependencies and then use this second fetch as the basis for the pinned json?

justhecuke:
@jin I think this is ready for review now. Take a look when you can.

private/rules/jetifier_maven_map.bzl
```diff
@@ -0,0 +1,126 @@
+jetifier_maven_map = {
```
jin: How did you generate this map? Can you add some docs/instructions/tool please?

private/rules/jetifier_maven_map.bzl
```diff
@@ -0,0 +1,126 @@
...
+    ("com.android.support", "animated-vector-drawable"): {"28.0.0" : ("androidx.vectordrawable", "vectordrawable-animated", "1.0.0")},
```
jin: Should the versions be hardcoded?

specs.bzl
```diff
@@ -32,7 +32,7 @@ def _maven_repository(url, user = None, password = None):
         credentials = {"user": user, "password": password}
         return {"repo_url": url, "credentials": credentials}
 
-def _maven_artifact(group, artifact, version, packaging = None, classifier = None, override_license_types = None, exclusions = None, neverlink = None, testonly = None):
+def _maven_artifact(group, artifact, version = None, packaging = None, classifier = None, override_license_types = None, exclusions = None, neverlink = None, testonly = None):
```
jin: Version cannot be done, since this is used in maven_install and directly in coursier, and we want to enforce that the version is specified.

specs.bzl
```diff
@@ -79,6 +80,15 @@ def _maven_artifact(group, artifact, version, packaging = None, classifier = Non
 
     return maven_artifact
 
+def _maven_artifact_to_coord_str(artifact):
+    return "{}:{}{}{}{}".format(
```
jin: nice.

specs.bzl
```diff
@@ -94,9 +104,15 @@ def _maven_exclusion(group, artifact):
...
     return {"group": group, "artifact": artifact}
 
+def _maven_coord_str_to_unversioned(coord_str):
+    coord_split = coord_str.split(":")
+    return "{}:{}".format(coord_split[0], coord_split[1])
+
 maven = struct(
     repository = _maven_repository,
     artifact = _maven_artifact,
+    artifact_to_coord_str = _maven_artifact_to_coord_str,
```
jin: Can you inline these into the jetifier bzl file instead, since that's the only caller?

specs.bzl
```diff
@@ -94,9 +104,15 @@ def _maven_exclusion(group, artifact):
...
+    coord_str_to_unversioned = _maven_coord_str_to_unversioned,
```
jin: Same for this, inline at callsite: I don't expect this to be a public API in the `maven` specification struct.

jin: Or move into the `utils` struct.

private/rules/jetifier_maven_map.bzl
justhecuke: That's how it is in jetifier. For compatibility reasons it maps one specific version to another since older/new versions aren't guaranteed to be compatible.

private/rules/jetifier_maven_map.bzl
justhecuke: I took https://android.googlesource.com/platform/frameworks/support/+/refs/heads/androidx-master-dev/jetifier/jetifier/core/src/main/resources/default.config and manually modified it's `pomRules` section and inlined `versions["latestReleased"]`. I'll write that into the file.

justhecuke:
> Thanks for this.
> 
> This took a while to review because of the large refactoring delta, and the code diverged away from a linear-ish read into functions, which I don't personally prefer. It'd be great it if the diff was smaller without the unrelated changes.
> 
> Do you think you can submit this change with a smaller delta for the jetifier specific work instead?

I wanted to avoid copy-pasting the big chunk of code to fetch using coursier twice so I was hoping to keep that one function (`make_coursier_dep_tree`).

I can certainly inline the other functions I used.

private/rules/jetifier_maven_map.bzl
jin: I see. Thanks.

jin:
Sounds good. A 2-pass fetch seems fine to me, as long as it's clear to end users that there's a cost to this. Perhaps note it in the documentation or surface in the `repository_ctx.report_progress` line?

specs.bzl
justhecuke: I'll inline.

specs.bzl
justhecuke: Done.

specs.bzl
justhecuke: Reverted.

private/rules/jetifier_maven_map.bzl
justhecuke: Added comments.

justhecuke:
@jin I think I've addressed your comments and you can take another look.

PR #394, opened 2020-03-11
support additional netrc lines for downloading pinned artifacts

### LINKED ISSUE #393: inject custom netrc lines for pinned installs in the workspace
For various reasons, we can't use the existing netrc/auth support for pinned deps: we can't change `~/.netrc`, nor can we check in the credentials in the json's mirror urls.  We _can_ generate the necessary netrc lines in a custom repository rule, but rules_jvm_external won't accept them that way.  It would be a great addition to support that: a new `additional_netrc_lines` or similar parameter on `pinned_maven_install`, which would mix them in to the existing netrc.

kevingessner:
Fixes #393

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,6 +144,12 @@ JSON file.
 
 Since all artifacts are persisted locally in Bazel's cache, it means that
 **fully offline builds are possible** after the initial `bazel fetch @maven//...`.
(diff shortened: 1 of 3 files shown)


kevingessner:
Done!  added a note about `~/.netrc` too, which was already in code.

PR #396, opened 2020-03-19
use jq instead of python to format the pinned JSON (fixes #395)

### LINKED ISSUE #395: pinned JSON file sort is unstable
The `pin.sh` script to generate the pinned install JSON uses `python -m json.tool` to format the JSON.  Unfortunately, `json.tool` changed its default behavior between python2 and python3: objects were sorted by key in py2, but not in py3.  This causes a lot of churn in the file, depending on what `python` is in the PATH of the last person to generate the file.  It would be better to use a consistent formatter, e.g.:

1. explicitly use `python2 -m json.tool` or `python3 -m json.tool`
2. use `python`'s json lib directly to format the JSON with code that is compatible with both 2 and 3
3. use something like `jq` instead (brought in via the workspace)
4. make the JSON formatter configurable in the repository rule

I can put together a PR for one of the above or a different idea.  Do you have a preference for fixing this?

kevingessner:
The formatted output is consistent with python2's `json.tool`.  No changes are required to how rules_jvm_external is added in the workspace nor how :pin is run.  I've tested on linux (centos) and OS X.

I didn't add jq for windows because I wasn't sure if pinning worked for windows at all, and I can't test it.  Should I add that as well?

### DIFF
--- a/coursier.bzl
+++ b/coursier.bzl
@@ -31,6 +31,10 @@ package(default_visibility = ["//visibility:{visibility}"])
 load("@rules_jvm_external//private/rules:jvm_import.bzl", "jvm_import")
 load("@rules_jvm_external//private/rules:jetifier.bzl", "jetify_aar_import", "jetify_jvm_import")
 
+{imports}
+"""
+
+_BUILD_PIN = """
 sh_binary(
     name = "pin",
     srcs = ["pin.sh"],
@@ -40,8 +44,6 @@ sh_binary(
         "@bazel_tools//src/conditions:windows": ["@rules_jvm_external_jq_windows//file"],
     }}),
 )
-
-{imports}
 """
 
 def _is_windows(repository_ctx):
@@ -226,7 +228,7 @@ def get_home_netrc_contents(repository_ctx):
                 return repository_ctx.read(netrcfile)
     return ""
 
-def get_jq_http_files():
+def _get_jq_http_files():
     '''Returns repository targets for the `jq` dependency that `pin.sh` needs.'''
     lines = []
     for jq in JQ_VERSIONS:
@@ -335,7 +337,7 @@ def _pinned_coursier_fetch_impl(repository_ctx):
                 http_files.append("        urls = [\"%s\"]," % artifact["url"])
             http_files.append("    )")
 
-    http_files.extend(get_jq_http_files())
+    http_files.extend(_get_jq_http_files())
 
     repository_ctx.file("defs.bzl", "\n".join(http_files), executable = False)
     repository_ctx.file(
@@ -783,7 +785,7 @@ def _coursier_fetch_impl(repository_ctx):
 
     repository_ctx.file(
         "BUILD",
-        _BUILD.format(
+        (_BUILD + _BUILD_PIN).format(
             visibility = "private" if repository_ctx.attr.strict_visibility else "public",
             repository_name = repository_name,
             imports = generated_imports,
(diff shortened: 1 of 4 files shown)


jin:
Yes, pinning does work on Windows. We test them on CI here: https://github.com/bazelbuild/rules_jvm_external/blob/7f29105eb4ac2c1a1582197337656d6e59b0acb9/.bazelci/presubmit.yml#L85-L101

kevingessner:
OK, green now on all platforms!

coursier.bzl
```diff
@@ -217,6 +226,22 @@ def get_home_netrc_contents(repository_ctx):
                 return repository_ctx.read(netrcfile)
     return ""
 
+def get_jq_http_files():
```
jin: prefix with `_`?

coursier.bzl
```diff
@@ -745,23 +774,23 @@ def _coursier_fetch_impl(repository_ctx):
         override_targets = repository_ctx.attr.override_targets,
     )
 
+    # This repository rule can be either in the pinned or unpinned state, depending on when
```
jin: Why is this here? This is the implementation of the pinned repo, so the repository name should never have `unpinned_`?

tests/integration/BUILD
```diff
@@ -22,6 +22,7 @@ genquery(
         "@testonly_testing//:com_google_code_findbugs_jsr305",
         "@testonly_testing//:com_google_auto_value_auto_value_annotations_1_6_3",
         "@testonly_testing//:com_google_auto_value_auto_value_annotations",
+        "@testonly_testing//:pin",
```
jin: The pinned repo should not have the pin build target.

coursier.bzl
```diff
@@ -22,16 +22,25 @@ load(
...
 
 _BUILD = """
 package(default_visibility = ["//visibility:{visibility}"])
 
-exports_files(["pin"])
-
 load("@rules_jvm_external//private/rules:jvm_import.bzl", "jvm_import")
 load("@rules_jvm_external//private/rules:jetifier.bzl", "jetify_aar_import", "jetify_jvm_import")
 
+sh_binary(
```
jin: this cases `bazel build @maven//...` to fail:

```
ERROR: missing input file 'external/maven/pin.sh', owner: '@maven//:pin.sh'
ERROR: /private/var/tmp/_bazel_jingwen/5e64b778ca5f4ff8a6ca994b3f33d8f8/external/maven/BUILD:7:1: @maven//:pin: missing input file 'LabelCause{label=@maven//:pin.sh, msg=missing input file 'external/maven/pin.sh', owner: '@maven//:pin.sh'}'
ERROR: /private/var/tmp/_bazel_jingwen/5e64b778ca5f4ff8a6ca994b3f33d8f8/external/maven/BUILD:7:1 1 input file(s) do not exist
```

This is because we don't expect the pinned repo to contain a `pin.sh` file, since pinning is done on the unpinned repo. Do you think you can find a way to continue making `bazel build @maven//...` work?

coursier.bzl
kevingessner: This was already here, I just moved it up so i can use it in BUILD.  I didn't check if it was needed in practice.

tests/integration/BUILD
kevingessner: `testonly_testing` isn't pinned (no maven_install_json), so it should have a `:pin` target to create the initial JSON. right? https://github.com/bazelbuild/rules_jvm_external/blob/7f29105eb4ac2c1a1582197337656d6e59b0acb9/WORKSPACE#L226

coursier.bzl
kevingessner: This is `_coursier_fetch_impl` which is called with `name = name if maven_install_json == None else "unpinned_" + name,` -- so the check is required afaict.

coursier.bzl
kevingessner: builds now.

tests/integration/BUILD
jin: Oops, you're right.

Issue #389, opened 2020-03-03
Feature: override coursier download location


peter-janssen:
Currently coursier is retrieved from: https://github.com/coursier/coursier/releases/download/{COURSIER_CLI_VERSION}/coursier.jar
This might not always be possible due to corporate proxies etc. 

The possibility to override this url to a private repository would be nice.

illicitonion:
Does anyone (cc @jin) have an idea for what the interface for this should be?

jin:
Since we're already using an environment variable for `COURSIER_CACHE`, `COURSIER_URL` and `COURSIER_SHA256` would work to replace the values here:

https://github.com/bazelbuild/rules_jvm_external/blob/bad9e2501279aea5268b1b8a5463ccc1be8ddf65/coursier.bzl#L571-L574

wisechengyi:
Thanks for the pointers, @jin!

However I've tried the following but it still tried to reach the default urls.
```
./bazel build --repo_env='COURSIER_URL=https://<some internal host>/io/get-coursier/bazel/v2.0.0-RC5-3/coursier.jar'
```

Looking back at the bzl files:
```
$ rg COURSIER_CACHE
coursier.bzl
470:    location = env_dict.get("COURSIER_CACHE")
472:    # If we're not using the unsafe shared cache, or if COURSIER_CACHE is not defined,
548:        # If not using the shared cache and the user did not specify a COURSIER_CACHE, set the default
552:        environment = {"COURSIER_CACHE": str(repository_ctx.path(coursier_cache_location))}
938:        "COURSIER_CACHE",
```

`COURSIER_CACHE` is used, but didn't see `COURSIER_URL` getting picked up anywhere.

(sorry newbie to bazel here in case I missed any basics)

jin:
This hasn't been implemented yet. I was responding to @illicitonion's question on how this would look like, design-wise.

wisechengyi:
Ah thanks for clarifying! I will sync with @illicitonion and try to take a stab at it.

illicitonion:
[...]

PR #402, opened 2020-04-11
Allow for custom coursier download url via COURSIER_URL env var

wisechengyi:
### Problem

Coursier download urls are hardcoded, hurdling firewalled environments without additional proxy support.

### Solution

Use `COURSIER_URL` specify a custom download url.

Addresses #389


jin:
Could you also add a section about this to the README please?

wisechengyi:
Gladly!

jin:
Thanks!

Issue #283, opened 2019-11-03
Buildozer suggestions broken in rules_jvm_external?


dhalperi:
Bazel 1.1.0 started warning about sha128 uses, which has pushed me to switch to `rules_jvm_external` from `rules_maven`.

Am I doing something wrong, or is Bazel now unable to give useful `buildozer` suggestions?

For example, in this case `A` and `B` both require Guava.

```python
java_library(
    name = "A",
    srcs = ["A.java"],
    deps = ["@maven//:com_google_guava_guava"],
)

java_library(
    name = "B",
    srcs = ["B.java"],
    deps = [
        ":A",
        "@maven//:com_google_guava_guava",
    ],
)

java_library(
    name = "B_no_dep",
    srcs = ["B.java"],
    deps = [
        ":A",
    ],
)
```

The error I get is not especially helpful at identifying the missing target that needs to be added to `B_no_dep`:

```
INFO: Writing tracer profile to '/private/var/tmp/_bazel_dan/55aaa51ea5620bd26651800f3bc9bddd/command.profile.gz'
INFO: Invocation ID: 57a6ccc4-bb94-4891-b8d4-6bdb1badfabd
INFO: Analyzed 3 targets (0 packages loaded, 0 targets configured).
INFO: Found 3 targets...
ERROR: /Users/dan/git-workspace/rules_jvm_demo/rules_jvm/BUILD:16:1: Building libB_no_dep.jar (1 source file) failed (Exit 1)
B.java:6: error: [strict] Using type com.google.common.collect.Lists from an indirect dependency (TOOL_INFO: "external/maven/v1/https/repo1.maven.org/maven2/com/google/guava/guava/26.0-jre/guava-26.0-jre.jar").
        System.err.println(Lists.reverse(A.INTEGERS));
                           ^
INFO: Elapsed time: 0.438s, Critical Path: 0.26s
INFO: 0 processes.
FAILED: Build did NOT complete successfully
```

With `rules_maven`, I got a `buildozer` command that worked:

```
INFO: Analyzed 3 targets (1 packages loaded, 5 targets configured).
INFO: Found 3 targets...
ERROR: /Users/dan/git-workspace/rules_jvm_demo/rules_maven/BUILD:16:1: Building libB_no_dep.jar (1 source file) failed (Exit 1)
B.java:6: error: [strict] Using type com.google.common.collect.Lists from an indirect dependency (TOOL_INFO: "@com_google_guava_guava//jar"). See command below **
[...]

PR #405, opened 2020-04-25
Add stamp_manifests option to maven_install so you can stamp manifests on a per maven_install basis

cheister:
When trying to use the `--@rules_jvm_external//settings:stamp_manifest` option in a`.bazelrc` file the IJ plugin fails to sync with the message `ERROR: unknown key: '--@rules_jvm_external//settings:stamp_manifest'`.

Instead this PR adds a `stamp_manifests` option to `maven_install` so you can configure manifest stamping for each `maven_install` you setup and don't have to set the global option for all uses of the plugin.

### DIFF
--- a/README.md
+++ b/README.md
@@ -818,6 +818,26 @@ maven_install(
 )
 ```
 
+### Stamping manifests for buildozer missing dependency support
+
+If you want more helpful messages when you are missing dependencies from this
+`maven_install` you can enable `stamp_manifests` to have Target-Label entries
+added to the manifests of the imported jar files.  The Target-Label entry provides
+Bazel with enough information to give you a `buildozer` command when a target is
+missing a dependency instead of just the path to the jar file.
+
+```python
+maven_install(
+    artifacts = [
+        # ...
+    ],
+    repositories = [
+        # ...
+    ],
+    stamp_manifests = True
+)
+```
+
 ### Jetifier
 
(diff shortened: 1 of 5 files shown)


jin:
I'm a bit hesitant of expanding the API surface for this, because we still haven't fixed the reproducibility issues with it yet. See https://github.com/bazelbuild/rules_jvm_external/issues/283#issuecomment-586762265 

Are you running into these reproducibility issues yourself?

Alternatively, we can fix the IntelliJ plugin to support starlark build settings flags.

cc @gregce as FYI - the IntelliJ plugin doesn't support SBC flags in .bazelrc files.

cheister:
We haven't seen the reproducibility errors because we haven't had stamping enabled because it does not work with the IntelliJ plugin.  I'll try running this branch and see what happens.  

Are the reproducibility errors just that the same jar can be stamped but "look different" to the cache and cause things to re-compile?

jin:
> Are the reproducibility errors just that the same jar can be stamped but "look different" to the cache and cause things to re-compile?

See https://github.com/bazelbuild/rules_jvm_external/pull/376 - essentially, `jar` embeds timestamps into the new jar, and that breaks remote caching.

cheister:
Ok.  Going to close this PR since it doesn't fix the timestamp problem.

Issue #324, opened 2019-12-11
Pinning artifacts on Windows does not work?


Michael-Nolan:
Windows 10
Bazel 1.2.1
rules_jvm_external 3.0 and 2.10 were tried. 

I have a small project that uses rules_jvm_external. If I run `bazel build //...` Then everything works fine with just Coursier  But when trying to pin the jars by running `bazel run @maven//:pin` I get the following error: 

```
Windows PowerShell
Copyright (C) Microsoft Corporation. All rights reserved.

PS C:\Users\USER_NAME> cd C:\PROJECT_LOCATION
PS C:\PROJECT_LOCATION> bazel run @maven//:pin
Starting local Bazel server and connecting to it...
INFO: Analyzed target @maven//:pin (1 packages loaded, 1 target configured).
INFO: Found 1 target...
INFO: Deleting stale sandbox base C:/users/USER_NAME/_bazel_USER_NAME/x2wgegak/sandbox
INFO: Elapsed time: 583.469s, Critical Path: 0.04s
INFO: 0 processes.
INFO: Build completed successfully, 1 total action
INFO: Build completed successfully, 1 total action
FATAL: ExecuteProgram(C:\users\USER_NAME\_bazel_USER_NAME\x2wgegak\external\maven\pin) failed: ERROR: src/main/native/windows/process.cc(199): CreateProcessW("C:\users\USER_NAME\_bazel_USER_NAME\x2wgegak\external\maven\pin"): %1 is not a valid Win32 application.
 (error: 193)
PS C:\PROJECT_LOCATION>
```

If i go look at the pin file that was created, I can see that it is a shell script:

```
#!/usr/bin/env bash

set -euo pipefail
readonly maven_install_json_loc=$BUILD_WORKSPACE_DIRECTORY/maven_install.json
readonly execution_root=$(bazel info execution_root)
readonly workspace_name=$(basename $execution_root)
cat <<"RULES_JVM_EXTERNAL_EOF" | python -m json.tool > $maven_install_json_loc
<THE EXPECTED OUTPUT IS HERE>
RULES_JVM_EXTERNAL_EOF

if [ "False" = "True" ]; then
    echo "Successfully pinned resolved artifacts for @maven, $maven_install_json_loc is now up-to-date."
else
    echo "Successfully pinned resolved artifacts for @maven in $maven_install_json_loc." \
      "This file should be checked in your version control system."
    echo
[...]

PR #407, opened 2020-05-09
Use coursier's default cache directory when fetching jars to update the install.json file

cheister:
Every time I run `bazel run @unpinned_maven//:pin` it downloads every jar defined in the `maven_install` target because the @unpinned_maven repository is deleted every time it is "fetched".  Basically the Coursier cache used for fetching at `external/unpinned_maven/v1` is never re-used when you are updating the `maven_install.json` file.  

This PR changes the unpinned fetch so it uses Coursier's default cache locations or the COURSIER_CACHE env var for the cache.

By re-using the cache a `bazel run @unpinned_maven//:pin` update now takes around 60 seconds instead of 10-15 minutes when the cache is empty.

### DIFF
--- a/coursier.bzl
+++ b/coursier.bzl
@@ -49,14 +49,14 @@ sh_binary(
 def _is_verbose(repository_ctx):
     return bool(repository_ctx.os.environ.get("RJE_VERBOSE"))
 
-def _is_windows(repository_ctx):
-    return repository_ctx.os.name.find("windows") != -1
+def _is_windows(os_name):
+    return os_name.find("windows") != -1
 
-def _is_linux(repository_ctx):
-    return repository_ctx.os.name.find("linux") != -1
+def _is_linux(os_name):
+    return os_name.find("linux") != -1
 
-def _is_macos(repository_ctx):
-    return repository_ctx.os.name.find("mac") != -1
+def _is_macos(os_name):
+    return os_name.find("mac") != -1
 
 # The representation of a Windows path when read from the parsed Coursier JSON
 # is delimited by 4 back slashes. Replace them with 1 forward slash.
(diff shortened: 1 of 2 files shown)


coursier.bzl
```diff
@@ -466,16 +467,24 @@ def _deduplicate_artifacts(dep_tree):
...
+
+    coursier_cache_env_var = os_env.get("COURSIER_CACHE")
+    if coursier_cache_env_var:
         # This is an absolute path.
-        return location
+        return coursier_cache_env_var
+
+    # locations from https://get-coursier.io/docs/1.1.0-M14-1/cache.html#default-location
+    if _is_windows(os_name):
+        return "%s\\Coursier\\Cache\\v1" % os_env.get("LOCALAPPDATA")
```
jin, line 483: Sorry for the delay, I just managed to try this on a Windows machine. 

While pinning artifacts on Windows is still pretty broken with the `pin` script not working properly (https://github.com/bazelbuild/rules_jvm_external/issues/324), this should be `"%s/Coursier/cache/v1" % os_env.get("LOCALAPPDATA").replace("\\", "/")`, based on what I've tested.

coursier.bzl
```diff
@@ -466,16 +467,24 @@ def _deduplicate_artifacts(dep_tree):
...
         return "v1"
-    else:
+
+    coursier_cache_env_var = os_env.get("COURSIER_CACHE")
+    if coursier_cache_env_var:
         # This is an absolute path.
-        return location
+        return coursier_cache_env_var
+
+    # locations from https://get-coursier.io/docs/1.1.0-M14-1/cache.html#default-location
```
jin, line 481: The current version is https://get-coursier.io/docs/2.0.0-RC5-3/cache.html#default-location

jin:
Thank you so much for this! Just tried it out on this repo itself, it is a *big* speedup.

coursier.bzl
cheister: Thanks for testing it on Windows!  Is it supposed to be lower case `cache` in the path or captilized `Cache` ?

coursier.bzl
cheister: updated

coursier.bzl
jin: It is `cache` on my machine.

PR #412, opened 2020-05-26
Add maven_deploy rule

cheister:
@jin I wanted to get feedback on a proposal to add a maven_deploy rule to rules_jvm_external.

This is a basic implementation based on a similar idea to https://github.com/graknlabs/bazel-distribution and https://github.com/google/dagger/tree/master/tools.  Basically, it's a rule that uses  an aspect to collect transitive dependencies and generate pom files.

It lets you make a `maven_deploy` target with bazel targets. If you run the rule it collects and builds the jars, generates pom files and then executes a script to upload the jars to maven.   For example if you created the following BUILD file.

```
load("@rules_jvm_external//deploy:maven_deploy.bzl", "maven_deploy")

maven_deploy(
    name = "maven_deploy",
    targets = [
        "//foo:lib",
        "//bar:lib",
    ],
)
```
You could run 
```
bazel run //:maven_deploy --@rules_jvm_external//deploy:maven_version=test123
```
to deploy the jars and all of the jars they depend on to maven.  The rule runs maven to upload the jars and assumes you have the credentials setup to run the `maven-deploy-plugin:deployfile` target.

It will use the `maven_coordinates` tag for each of the maven jars it uploads.

This is a very basic start to the rule, there are lots of maven cases that could be addressed such as test only and shaded jars and a more elaborate pom template.  It would also be helpful to template the maven deploy script that is generated so you could use anything you wanted to upload the jars.

I'm posting this to find out if you have interest in adding something like this to rules_jvm_external and for feedback on what you think of the approach.  

If it's something you feel is worth adding then I would need to flush out a better API and more documentation.

### DIFF
--- a/deploy/BUILD
+++ b/deploy/BUILD
@@ -1,0 +1,12 @@
+# tools/maven/BUILD
+
+load("//deploy:maven_version.bzl", "maven_version")
+
+package(default_visibility = ["//visibility:public"])
+
+exports_files(["pom.xml.template"])
+
+maven_version(
+    name = "maven_version",
+    build_setting_default = "SNAPSHOT",
+)
--- a/deploy/maven_deploy.bzl
+++ b/deploy/maven_deploy.bzl
@@ -1,0 +1,184 @@
+load(":maven_version.bzl", "MavenVersionProvider")
+
+MavenDeployInfo = provider(
+    fields = {
+        "deps": "MavenDeployInfo providers of dependencies",
+        "jar": "jar file to deploy",
+        "maven_coordinates": "maven coordinates for this jar",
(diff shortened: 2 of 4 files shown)


jin:
Wow, thanks for this! I have no experience with Maven publishing, so I'll take some time to learn about it and try this out, before I'll be able to properly review this.

A while ago, I was also discussing adding Maven deploy rules with @shs96c, who I believe also has his own Maven deploy rules, which he planned to upstream to this repo. I just pinged him about this PR -- do you mind if we also wait for a bit for his version, and then potentially merge the two approaches?

cheister:
Declining in favor of #413

PR #413, opened 2020-05-26
Upstream java export

shs96c:
Creates a suite of rules to allow us to upload artifacts to maven repositories. The normal entry point for most users will be `java_export`

The rules ultimately lead to `maven_publish`, which allows us to push to the Sonatype OSS maven repository. This requires gpg signing, and the presence of javadocs and source jars.

These are based on rules derived from Selenium and authored by myself. Where appropriate, the copyright headers have been retained.


shs96c:
I've kept the different commits separate to try and make it easier to review. Each can be reviewed in isolation, though they mesh to form the common whole. The expected entry point for most users is `java_export`, which is essentially a drop-in replacement for `java_library` (in the same way that docker's `java_image` is: that is, there are some extra parameters, but it should be easy enough to use).

shs96c:
It may be that there are elements of #412 that should be merged into this. I'd appreciate a quick sanity check of the approach in this PR before going further.

shs96c:
I'll add the test and docs. I guess I should hook in some skydoc stuff, right?

I'd be happy to maintain this and review PRs. The plan is to switch Selenium to this once we add support for generating JPMS module-info files again (which I'll be doing in a follow up PR once this has landed)

shs96c:
@cheister does this PR cover your use-cases as well? Is there anything that we should consider adding or changing?

jin:
Skydoc will be excellent as well.

> I'd be happy to maintain this and review PRs. The plan is to switch Selenium to this once we add support for generating JPMS module-info files again (which I'll be doing in a follow up PR once this has landed)

Perfect! Let me iron out the administrative processes internally and add you as a maintainer.

davido:
@shs96c Very impressive, indeed.

@ronshapiro added a Maven publish tool chain for Dagger project. Would this PR work for his use case out of the box?

shs96c:
@davido it should do. If it doesn't, then there's clearly something missing.

jin:
@shs96c I have granted you write access to this repository.

shs96c:
Thank you, @jin. I've added some more docs. The next step is the test, then I think we're ready for review :)

shs96c:
Tests are in. We use some Java 11 features in the helper tools, which is why the Java 8 builds are failing. I can fix those, but I think this is ready for a review with that caveat.

jin:
@shs96c you can exclude specific tests from the JDK 8 and Windows setups in the .bazelci/presubmit.yml file with the target subtraction syntax (`-- -//package:test`). Until we have a JDK11 on macOS CI configuration (cc @philwo), I think it's sufficient to exclude them and document that java_export is only tested and maintained on Linux and (maybe) macOS with JDK11 for now. We don't have to support all three platforms for the scope of this PR.

tests/integration/java_export/BUILD
```diff
@@ -0,0 +1,60 @@
...
+    ],
+    deps = [
+        ":project",
+        artifact("com.google.guava:guava"),
+        artifact("org.hamcrest:hamcrest"),
+    ],
+    data = [
+        # It's not apparent why it's not sufficient to simply add this:
+        # one must also include this in the `deps` too, else the test
+        # can't find the file.
```
jin, line 35: `deps` adds the jar to the test's compile/runtime classpath, whereas `data` adds to the java-agnostic runfiles. `data` basically lets you call `Paths.get` in the test.

private/rules/has_maven_deps.bzl
```diff
@@ -0,0 +1,123 @@
+MavenInfo = provider(
+    fields = {
+        "coordinates": "Maven coordinates for the project, which may be None",
+        "artifact_jars": "Depset of runtime jars that are unique to the artifact",
+        "jars_from_maven_deps": "Depset of jars from all transitive maven dependencies",
+        "artifact_source_jars": "Depset of source jars that unique to the artifact",
+        "source_jars_from_maven_deps": "Depset of jars from all transitive maven dependencies",
+        "maven_deps": "Depset of first order maven dependencies",
+        "as_maven_dep": "Depset of this project if used as a maven dependency",
+        "deps_java_infos": "Despet of JavaInfo instances of dependencies not included in the project",
```
jin, line 10: "Depset"

README.md
```diff
@@ -913,6 +914,47 @@ java_library(
...
+  deps = [
+    "//user_project/utils",
+    "@maven//:com_google_guava_guava",
+  ],
+)
+```  
+
+In order to publish the artifact, use `bazel run`:
+
+`bazel run --define "maven_repo=file://$HOME/.m2/repository" //user_project:exported_lib.publish`
```
dsilvasc, line 942: To publish multiple artifacts from the same project (like jersey or jackson split their publications), would that be feasible with a single run target?

shs96c: At the moment, no. In the Selenium project, we have a script that uses a `bazel query` to find the items to deploy, and then calls `bazel run` in a loop. It'd be nice to have a single target for deploying everything….

shs96c:
Created #422 to update the docs too.

README.md
emidln: I realize I'm late to the party, but I handle this problem via constructing a parent pom with modules based on a bazel query. We generate pom.xml files for each bazel package via a macro augmenting our `java_library` derivative as well as a standalone macro to declare a maven package comprised of arbitrary artifacts. This allows `mvn` to optimize the artifact upload in cases where we have hundreds of artifacts. I think this approach is feasible to adopt here to allow multiple targets to be uploaded at once.

shs96c: Probably better if we talked about this on the #maven channel in the Bazel Slack. Generating a parent pom would be very useful.

Issue #417, opened 2020-06-01
Unable to pin due to incorrect coursier cache path


joprice:
I'm seeing the error `Error while trying to parse the path of file in the coursier cache` when trying to pin with the latest changes meant to improve performance https://github.com/bazelbuild/rules_jvm_external/commit/b16b35fa7160fc6d09752efb80baacb254bce801. I added a print statement and found that somehow, even though I'm on mac, my coursier cache is at ` $HOME .coursier/cache` instead of the expected `$HOME/Library/Caches/Coursier/v1`. Is this some incorrect setup or something I have cached? I have no env vars related to coursier exported. It surprises me that failing to find a cached file when pinning would not be something the rules would attempt to recover from, but I assume I'm in a broken state that wasn't anticipated.

jin:
cc @cheister 

That somewhat means that your machine isn't going into the [`_is_macos()`](https://github.com/bazelbuild/rules_jvm_external/commit/b16b35fa7160fc6d09752efb80baacb254bce801#diff-062be76e7bf90a2fa889e8a9ec966b5cR487) branch here. Do you mind trying a local copy of rules_jvm_external and confirming that?

I'll add this to the release blockers.

joprice:
Yes I'll check that now

joprice:
Looks like it's this line https://github.com/coursier/coursier/blob/f48c1c6b01ac5b720e66e06cf93587b21d030e8c/modules/paths/src/main/java/coursier/paths/CoursierPaths.java#L60. It uses the `$HOME/.coursier` folder if present. Maybe for backwards compat, but maybe also to allow using the same folder across different os'es? I moved my folder out of the way and then it used `$HOME/Library/Caches/Coursier/v1` and was able to pin. So I guess that logic in `get_coursier_cache_or_default` needs to be extended to account for this?

jin:
Looks like it, we should maintain backwards compat with existing directories. I think doing a `test` for the existence of the directory suffices? Similar to what we're doing for netrc: 

https://github.com/bazelbuild/rules_jvm_external/blob/b16b35fa7160fc6d09752efb80baacb254bce801/coursier.bzl#L231

joprice:
I hit another issue just now, which may be a bug in coursier: error:

```org.scalafmt.dynamic.exceptions.ScalafmtException: failed to resolve Scalafmt version '2.4.2'
[...]

PR #419, opened 2020-06-04
Fix coursier cache path logic for https://github.com/bazelbuild/rules_jvm_external/issues/417

cheister:
Update how the default coursier cache path is calculated based on the existence of a $HOME/.coursier directory.  

Based on  https://github.com/coursier/coursier/blob/f48c1c6b01ac5b720e66e06cf93587b21d030e8c/modules/paths/src/main/java/coursier/paths/CoursierPaths.java#L60

From https://github.com/bazelbuild/rules_jvm_external/issues/417

Additional information: https://get-coursier.io/docs/cache.html#compatibility-with-former-versions

### DIFF
--- a/coursier.bzl
+++ b/coursier.bzl
@@ -49,14 +49,20 @@ sh_binary(
 def _is_verbose(repository_ctx):
     return bool(repository_ctx.os.environ.get("RJE_VERBOSE"))
 
-def _is_windows(os_name):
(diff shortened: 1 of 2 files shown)


jin:
I added a link to https://get-coursier.io/docs/cache.html#compatibility-with-former-versions in the PR description. 

Merging!

PR #420, opened 2020-06-09
Unpin Android SDK's build_tools and api_level versions

jin:


shs96c:
Not quite sure why the Windows build is failing because of this. It should be fine. The alternative is to just document the required version of Android.

jin:
It's actually a problem with the Android rules where newer versions of the Android build tools is incompatible on Windows.  See https://github.com/bazelbuild/continuous-integration/pull/984#issuecomment-641257701

I'll close this for now, until the Android rules are fixed for Windows.

PR #424, opened 2020-06-25
Make behaviour when merging duplicates configurable

shs96c:
Although the behaviour before was stable, it wasn't obvious, so this
change makes the behaviour of MergeJars explicit and configurable via
the command line.

This change also adds some (much needed) unit tests for MergeJars to
ensure that it behaves as we expect it should.


jin:
Thanks!

Issue #409, opened 2020-05-18
Pin command requires pinned_maven_install due to jq dependency


jfancher:
Following the documentation in the main `README`, I created a simple `maven_install`:

```starlark
workspace(name = "missing_jq_repro")

load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")

http_archive(
    name = "rules_jvm_external",
    sha256 = "9508c97baa2b3a5a952b4d3acaa06ee71e6b36bed7d62db2d919d6e450078862",
    strip_prefix = "rules_jvm_external-b500b01137f75e0f8442bb4429ef5ce59135864a",
    url = "https://github.com/bazelbuild/rules_jvm_external/archive/b500b01137f75e0f8442bb4429ef5ce59135864a.zip",
)

load("@rules_jvm_external//:defs.bzl", "maven_install")

maven_install(
    artifacts = ["com.google.guava:guava:27.0-jre"],
    repositories = ["https://repo1.maven.org/maven2"],
)
```

Then I followed the [instructions](https://github.com/bazelbuild/rules_jvm_external#pinning-artifacts-and-integration-with-bazels-downloader) for pinning, which say the next step is to run `@maven//:pin`. It fails:

```console
$ bazel run @maven//:pin

ERROR: [...]/external/maven/BUILD:103:1: no such package '@rules_jvm_external_jq_linux//file': The repository '@rules_jvm_external_jq_linux' could not be resolved and referenced by '@maven//:pin'
ERROR: Analysis of target '@maven//:pin' failed; build aborted: no such package '@rules_jvm_external_jq_linux//file': The repository '@rules_jvm_external_jq_linux' could not be resolved
```

It seems this dependency repo is pulled in as part of the pinned install itself. Adding the following fixed it:

```starlark
load("@maven//:defs.bzl", "pinned_maven_install")

pinned_maven_install()
```

I'm not sure whether this is expected (it seems a bit backwards), but if this is the required order now the docs should probably be updated.

jin:
Running the initial pin should not depend on the existence of the pinned repo, so this is a bug.

PR #426, opened 2020-06-26
Allow `bazel run @maven//:pin` to run successfully

shs96c:
To do this, we download each version of jq that might be used and
ensure it's available. Decide at run time which version we'll use.

Fixes https://github.com/bazelbuild/rules_jvm_external/issues/409

### DIFF
--- a/coursier.bzl
+++ b/coursier.bzl
@@ -35,14 +35,26 @@ load("@rules_jvm_external//private/rules:jetifier.bzl", "jetify_aar_import", "je
 """
 
 _BUILD_PIN = """
+genrule(
+    name = "jq-binary",
+    cmd = "cp $< $@",
+    outs = ["jq"],
+    srcs = select({{
+        "@bazel_tools//src/conditions:linux_x86_64": ["jq-linux"],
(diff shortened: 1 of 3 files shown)


alexeagle:
Hmm, this isn't working for me

```
% bazel run @unpinned_maven//:pin
/private/var/tmp/_bazel_alex.eagle/03c50bd9f989d25b88a1be300d53820c/execroot/rh/bazel-out/darwin-fastbuild/bin/external/unpinned_maven/pin: line 10: external/unpinned_maven/jq: No such file or directory
```

Changing this arg
https://github.com/bazelbuild/rules_jvm_external/pull/426/files#diff-062be76e7bf90a2fa889e8a9ec966b5cR53
from `$(location :jq-binary)` to just `jq` fixes it for me.
(note, `$(rootpath :jq-binary)` didn't fix it either, still has the extra "external/unpinned_maven" segments)

We go to lengths to genrule-copy the file to just "jq" so it doesn't seem like we should have to do this? unless it's for windows, but then you still don't have a bash.runfiles helper used in the pin.sh script

PR #427, opened 2020-06-26
Add pinning step to the simple example for user journey coverage

jin:
Doesn't work on Windows: https://github.com/bazelbuild/rules_jvm_external/issues/324


shs96c (approved):
Ship it!

PR #434, opened 2020-07-08
Avoid calling out to Bazel

gmishkin:
I was thinking about the case where user doesn't have bazel in their path but only bazelisk, or is using our vendored bazelisk Python script, etc. Are you open to this? It seemed like a lot of effort just to print the workspace name into the terminal output help text, and isn't it optional to begin with?

Let me know and if not I can try to rework this to get this info some other way.

### DIFF
--- a/private/pin.sh
+++ b/private/pin.sh
@@ -3,8 +3,7 @@
 set -euo pipefail
 readonly maven_install_json_loc={maven_install_location}
 # This script is run as a `sh_binary`, so ensure we are in the calling workspace before running Bazel.
-readonly execution_root=$(cd "$(dirname "$maven_install_json_loc")" && bazel info execution_root)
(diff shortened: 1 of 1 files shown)


gmishkin:
@jin understood. I wonder if the `BAZEL_REAL` environment variable is available here and I could run that instead. Does Bazel always set that or is it a bazelisk-only feature? I'm thinking it'd have to always be set, otherwise it would be hard to write a `tools/bazel` script without bazelisk.

jin:
I think that's a variable exported only to the `tools/bazel` script.

gmishkin:
I think https://github.com/bazelbuild/bazelisk/pull/142 will help me. If someone's directly running a "bazel" that's _not_ a bazelisk, we'd have bigger problems.

PR #437, opened 2020-07-17
Fix compatibility with --incompatible_disable_depset_items

laurentlb:
The first depset argument will be named 'direct' instead of 'items'.
Most of the time, it is passed positionally though.

The flag will be enabled by default in Bazel 4.0.


jin:
I've created a PR here: https://github.com/bazelbuild/rules_jvm_external/pull/431

PR #444, opened 2020-08-12
Handle tags in java_export rule

gibfahn:
Using "Tags" attribute on java_export() rule generates:

```
java_library(...)
got multiple values for parameter 'tags' via java_export rules
```


gibfahn:
cc/ @shs96c

PR #451, opened 2020-08-25
Updating Android API version to 29

comius:
Build kite continuous integration environment was updated.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -2,8 +2,8 @@ workspace(name = "rules_jvm_external")
 
 android_sdk_repository(
     name = "androidsdk",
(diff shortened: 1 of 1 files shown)


jin:
Error from the android_instrumentation_test example: 

```
/bin/bash -c 'source external/bazel_tools/tools/genrule/genrule-setup.sh; echo           This rule requires that the Android SDK used by Bazel has the           following system image installed: emulator_images_android_23_x86. Please install this system           image through the Android SDK Manager and try again. ;           exit 1
--
  | ')
```

Was the Android Emulator API level 23 removed?

PR #452, opened 2020-08-26
Spell auth scheme as "Basic" for artifactory

dmivankov:
JFrog Artifactory seems to accept only "Basic" casing.
Even though authorization scheme should be case-insensitive https://tools.ietf.org/html/rfc7617#section-2

### DIFF
--- a/private/tools/java/rules/jvm/external/maven/MavenPublisher.java
+++ b/private/tools/java/rules/jvm/external/maven/MavenPublisher.java
@@ -171,7 +171,7 @@ public class MavenPublisher {
       if (credentials.getUser() != null) {
         String basicAuth = Base64.getEncoder().encodeToString(
           String.format("%s:%s", credentials.getUser(), credentials.getPassword()).getBytes(US_ASCII));
-        connection.setRequestProperty("Authorization", "BASIC " + basicAuth);
+        connection.setRequestProperty("Authorization", "Basic " + basicAuth);
       }
       connection.setRequestProperty("Content-Length", "" + Files.size(toUpload));
 


brymck:
Thanks, @dmivankov, I've experienced the same issue today and found this to be the fix. As another note in support of this PR, [Section 2 of RFC 2617](https://tools.ietf.org/html/rfc2617#section-2) defines the format for basic auth as `"Basic" 

> The "basic" authentication scheme is based on the model that the client must authenticate itself with a user-ID and a password for each realm.  The realm value should be considered an opaque string which can only be compared for equality with other realms on that server. The server will service the request only if it can validate the user-ID and password for the protection space of the Request-URI. There are no optional authentication parameters.
>
> For Basic, the framework above is utilized as follows:
>
> &nbsp;&nbsp;&nbsp;challenge   = "Basic" realm
> &nbsp;&nbsp;&nbsp;credentials = "Basic" basic-credentials
>
> ...
>
> If the user agent wishes to send the userid "Aladdin" and password "open sesame", it would use the following header field:
>
> &nbsp;&nbsp;&nbsp;Authorization: Basic QWxhZGRpbjpvcGVuIHNlc2FtZQ==

While the generic syntax is case-insensitive, I don't believe this conforms to that syntax ([Section 2.1 of RFC 2650](https://tools.ietf.org/html/rfc6750#section-2.1) for bearer tokens calls this out specifically), and `"Basic"` rather than `"BASIC"` is the safest approach. At the least, it's ambiguous, which is why Artifactory and others (including Go's `net/http` and Rails) may impose case-sensitive restrictions on `"Basic"`. Merging this PR will maximize compatibility with stricter parsers.

jin (approved):
LGTM, thanks!

jin:
This appears to fix https://github.com/bazelbuild/rules_jvm_external/issues/476.

PR #453, opened 2020-08-26
Output leading zeroes in maven export checksums

dmivankov:


jin:
Could we add some tests for this?

dmivankov:
Sure. Though sadly I won't have a chance to add some tests until October

dmivankov:
@jin and now we've got tests :)

jin (approved):
Awesome, thank you :)

PR #454, opened 2020-09-02
Fix jq path in :pin runnable

alexeagle:
Problem reported on #426

The build typically takes place in the user's workspace, so using
`$(location)` (or more correctly, `$(rootpath)`) gives an undesirable
"external/" leading segment. This would be useful if the binary being
run is in the users workspace, but we know our binary is in the same
workspace defining the BUILD file.

This change also uses the runfiles helper since I think that's required
to make this work on Windows, though I only tested on Mac.

### DIFF
--- a/coursier.bzl
+++ b/coursier.bzl
@@ -50,7 +50,7 @@ sh_binary(
     name = "pin",
     srcs = ["pin.sh"],
     args = [
-      "$(location :jq-binary)",
+      "$(rootpath :jq-binary)",
     ],
     data = [
         ":jq-binary",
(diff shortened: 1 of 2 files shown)


alexeagle:
Hmm, CI failure is just
```
subprocess.CalledProcessError: Command '['set -e\nbazel run @unpinned_regression_testing//:pin\nbazel run @unpinned_maven_install_in_custom_location//:pin\nbazel run @duplicate_artifacts_test//:pin']' returned non-zero exit status 1.
```
I ran those locally on Mac and all three `bazel run` commands succeed

alexeagle:
ping @jin @shs96c

alexeagle:
update: the problem is only observable under `--nolegacy_external_runfiles`

alexeagle:
See https://github.com/bazelbuild/bazel/issues/9306 where the rollout of that incompatible change is stalled maybe?

shs96c:
Discussed with @alexeagle on the Slack. Looks like his change has just been pushed.

shs96c (approved):
LGTM

PR #457, opened 2020-09-22
Handle '@' in username

ablyes:
handle the case where the username contains one or more '@' char (like an email address used as username).

### DIFF
--- a/coursier.bzl
+++ b/coursier.bzl
@@ -440,7 +440,8 @@ def remove_auth_from_url(url):
     host = url_parts[0]
     if "@" not in host:
         return url
-    userless_host = host[host.find("@") + 1:]
+    last_index=host.rfind("@", 0, None)
+    userless_host=host[last_index + 1:]
     new_url = "{}://{}".format(protocol, "/".join([userless_host] + url_parts[1:]))
     return new_url
 


ablyes:
@googlebot I signed it!

coursier.bzl
```diff
@@ -440,7 +440,8 @@ def remove_auth_from_url(url):
     host = url_parts[0]
     if "@" not in host:
         return url
-    userless_host = host[host.find("@") + 1:]
+    last_index=host.rfind("@", 0, None)
```
plaird, line 443: FYI there are some split_url() tests that have weird urls that this probably won't work for. Not sure if those are real cases somebody has, or not. 

https://github.com/bazelbuild/rules_jvm_external/blob/master/tests/unit/coursier_test.bzl#L208

ablyes: I think it's not realistic to have a host with @ 
My commit should fix those 2 cases:
username@whatever.com:password@host
username@whatever.com@itLikes@icanputanotheronehere@:password_with_@_why_not:host

plaird: I agree, I don't think having that @ char in the path, in unencoded form, is even legal. But just FYI someone thought it was a valid use case.

jin (commented):
Could you add tests?

ablyes:
Yes, I will do that. 
May be just delete the none realistic tests also.

shs96c (approved):
This makes sense. Thank you for the PR!

PR #463, opened 2020-10-04
Add google cloud storage support to java_export

### LINKED ISSUE #458: Java export - publish to private google cloud storage bucket
Currently, the only way to publish a library is to local FS or HTTP site, so if you want to publish a private lib, you've to create a maven repo and grant access through user/password.
It will be nice to allow sending the files to GCS instead using application-default credentials, which will allow a quick private maven repo setup.
Also - if we do that, we should also support GCS as a source repo (should probably be a different issue).

Can take this issue on myself if we're up to support it, should be a simple change.

zacharya19:
Closes #458 (this issue is for publishing only, I would probably open another one to add GCS support to maven_install, hopefully I'll get to it).
@shs96c LMK what you think.


README.md
```diff
@@ -952,6 +960,10 @@ bazel run --stamp \
   //user_project:exported_lib.publish`
 ```
 
+It's also possible to publish to a Google Cloud Storage bucket (make sure to not add a trailing slash):
```
shs96c, line 963: Nit: could we just remove a trailing slash if we see one?

defs.bzl
```diff
@@ -154,6 +154,22 @@ def _parse_artifact_str(artifact_str):
     else:
         return parse.parse_maven_coordinate(artifact_str)
 
+JAVA_EXPORT_REPO_NAME = "internal_java_exports_maven"
+
+def java_export_deps():
+    """
+        Adds the maven dependencies that are needed to run java_export rule.
+    """
+    maven_install(
```
shs96c, line 163: Any reason not to use a `jvm_import` to pull in a jar we add to the repo? Other than the vast number of dependencies? 

The current way of using `rules_jvm_external` doesn't need anyone to go through the cumbersome "import the ruleset, load the rule set deps, run the deps setup" dance. That's why the current deployer doesn't use anything outside the Java 8 JRE.

shs96c (changes requested):
This looks like it's in good shape. Thank you for writing it! 

@jin can weigh in on the request to avoid asking people to add extra steps to their workspaces. Is there any way we can import them into the repo itself? 

It'd be nice to have some tests, but I guess it's hard to add those for publishing.

If credentials are needed, how are they detected and specified?

README.md
zacharya19: Thought about it, but I didn't want to mess with the users args, in case s/he really want the extra slash from some reason.
Now that I think about maybe it doesn't really make sense, will add it.

defs.bzl
zacharya19: Maybe I'm wrong, but AFAIK bazel never supported transitive deps, so the only real way to avoid forcing the user from adding those deps is to save the jars in our repo, which isn't a solution IMHO.
Am I wrong about it? Even with jvm_import it will include target from different repo, which the user will need to add.

zacharya19:
> If credentials are needed, how are they detected and specified?
Right now I'm using the application default, since it's GCS best practice.
I can add support for providing credentials file, but I think it's better to use application default.

README.md
zacharya19: Done.

defs.bzl
jin: If it's just a single (or few) jar dependency, we could vendor it.

How about doing it this way instead of requiring a load statement? https://github.com/bazelbuild/rules_jvm_external#exporting-and-consuming-artifacts-from-external-repositories

zacharya19: The GCS lib has 23 deps (so 24 total jars), we can vendor them but I'm not sure it's the right approach.
LMK what you think about it, if you prefer it to be vendored I'll change it.

jin: @shs96c and I had a discussion offline about this, and I think that it's would be a cleaner approach to go with your implementation, with an internal `maven_install` repo.

zacharya19: :+1:

defs.bzl
```diff
@@ -154,6 +154,22 @@ def _parse_artifact_str(artifact_str):
...
+        name = JAVA_EXPORT_REPO_NAME,
+        artifacts = [
+            "com.google.cloud:google-cloud-storage:1.111.2",
+        ],
+        repositories = [
+            "https://jcenter.bintray.com/",
```
jin, line 169: The repo list should be parameterized through `java_export_deps()` for users to switch repos.

defs.bzl
```diff
@@ -154,6 +154,22 @@ def _parse_artifact_str(artifact_str):
     else:
         return parse.parse_maven_coordinate(artifact_str)
 
+JAVA_EXPORT_REPO_NAME = "internal_java_exports_maven"
```
jin, line 157: Prefix with `_INTERNAL` to make it clear that this is not meant for consumption by end users.

README.md
```diff
@@ -916,6 +916,14 @@ Any version conflicts or duplicate artifacts will resolved automatically.
 
 ## Publishing to External Repositories
 
+In order to use this feature you must import the java_export maven deps in your workspace:
```
jin, line 919: nit:

`java_export` Maven dependencies in your `WORKSPACE` file:

jin (approved):
LGTM, please also get approval from @shs96c.

defs.bzl
zacharya19: I changed to `INTERNAL_JAVA_EXPORT_REPO_NAME` since vars starts with underscore are internal and can't be imported in the build file.

shs96c:
Sorry for the slow feedback on this. After discussing with @jin, I think the best thing to do to avoid vendoring in all the jars we need is to move `rules_jvm_external` to a two-step initialisation process (load the deps, run the setup). I'll prep a PR for that for you to build on.

shs96c:
After a far too long, I've created #481 to help make it easier for us to pull in the bits that we need to rules_jvm_external

shs96c:
I've landed #481 so it would be helpful if you could rebase on top of that, and use those features to pull in any dependencies, please.

zacharya19:
@shs96c Done, LMK what you think, it's a much smaller change now, most of it is maven_pin file.

shs96c (approved):
Thanks for you patience while we got the infrastructure in to support this, and thank you again for the patch. LGTM!

Issue #445, opened 2020-08-13
Depending on embedded JDK


spietras:
As far as I understand, Bazel by default [provides an embedded JDK]( https://docs.bazel.build/versions/master/bazel-and-java.html#configuring-the-jdk) (or at least some rules to get it), so we don't have to have one locally.

```rules_jvm_external``` use Coursier, which is invoked by running ```java```: 

https://github.com/bazelbuild/rules_jvm_external/blob/aed0eb3747682e64611bf98bb1ca590a44b470de/coursier.bzl#L109-L129

This assumes java is available at ```JAVA_HOME```, which is unset with no local JDK. If I understand correctly Bazel sets [```JAVABASE```](https://docs.bazel.build/versions/0.19.1/be/make-variables.html#predefined_variables) to the location of the embedded JDK.

So is it possible to use the embedded JDK with ```rules_jvm_external```? Maybe it is possible now and I am missing something? And if it's not currently possible, wouldn't it be sufficient to just check if ```JAVABASE``` is set too? I guess the order of checking is important because someone can have a local JDK, while still wanting to use the embedded one.

For now, a cheap workaround is to just set ```JAVA_HOME``` relatively, knowing that both ```maven``` and ```jdk``` are in the ```external``` workspace. I can confirm that adding this line:
```
build --action_env=JAVA_HOME=../bazel_tools/jdk
```
 to the ```.bazelrc``` file works and the embedded JDK is used.

jin:
Looks like it's actually possible, but I just haven't got to doing this yet. Would you wanna take a stab at this and submit a PR?

spietras:
> Would you wanna take a stab at this and submit a PR?

Sure, I'll give it a try soon.

> Looks like it's actually possible to use the downloaded JDK (not embedded one, since the embedded one is minimal enough to run Bazel, and probably not other Java applications)

Yeah, sorry, by "embedded" I meant the downloaded JDK, not the embedded JRE. So from now on, I will refer to the downloaded JDK as the "remote JDK", because that's how it's called in the bazel_tools workspace.

jin:
Thank you for taking the time to look into this, @spietras. It sounds like there's a usability gap between toolchains and repository rules, and the current API is not sufficient to express such a relationship.

[...]

PR #465, opened 2020-10-12
Add @maven//:outdated to check for updated versions of artifacts

cheister:
This adds an outdated action to a maven_install repository so you can run
```
bazel run @maven//:outdated
```
to check for updated versions of artifacts. e.g.

```
Checking for updates of 23 artifacts against the following repositories:
	https://repo1.maven.org/maven2
	https://digitalassetsdk.bintray.com/DigitalAssetSDK
	https://maven.google.com
	https://packages.confluent.io/maven/

org.pantsbuild:jarjar [1.6.6 -> 1.7.2]
junit:junit [4.12 -> 4.13.1]
org.jetbrains.kotlin:kotlin-test [1.3.21 -> 1.4.10]
com.digitalasset:damlc [100.12.1 -> 100.13.45]
com.squareup:javapoet [1.11.1 -> 1.13.0]
```


private/outdated.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+if [ -f "private/tools/prebuilt/outdated_deploy.jar" ]; then
```
cheister, line 5: I'm not sure if there is a better way to check if the action is running inside the rules_jvm_external repo or outside of it.

private/tools/java/rules/jvm/external/maven/Outdated.java
```diff
@@ -0,0 +1,126 @@
...
+      String groupId = artifactParts[0];
+      String artifactId = artifactParts[1];
+      String version = artifactParts[2];
+
+      String releaseVersion = null;
+      for (String repository : repositories) {
+        releaseVersion = getReleaseVersion(repository, groupId, artifactId);
+        if (releaseVersion != null) {
+          verboseLog(String.format("Found version [%s] for %s:%s in %s", releaseVersion, groupId, artifactId, repository));
+          // Should we search all repositories in the list for latest version instead of just the first
```
cheister, line 113: Using the first one found has been working for me but may not find the truly latest version of an artifact in the list of repositories.

private/outdated.sh
```diff
@@ -0,0 +1,9 @@
...
+    java -jar private/tools/prebuilt/outdated_deploy.jar external/{repository_name}/outdated.artifacts external/{repository_name}/outdated.repositories
```
cheister, line 6: Also, another todo, this probably doesn't work if proxy vars are set.

shs96c (approved):
The logic in `Outdated.java` looks good to me.

private/outdated.sh
cheister: As far as I can tell https://github.com/bazelbuild/rules_jvm_external/issues/445 is still an open issue because JAVABASE is not exposed to repository rules.  If we fix JAVABASE for the java that coursier uses we should fix it here also but I don't currently see a straightforward fix otherwise.

private/tools/java/rules/jvm/external/maven/Outdated.java
```diff
@@ -0,0 +1,133 @@
...
+      System.exit(1);
+    }
+
+    String artifactsFilePath = args[0];
+    String repositoriesFilePath = args[1];
+
+    List<String> artifacts = Files.readAllLines(Paths.get(artifactsFilePath), StandardCharsets.UTF_8);
+    List<String> repositories = Files.readAllLines(Paths.get(repositoriesFilePath), StandardCharsets.UTF_8);
+
+    System.out.println(String.format("Checking for updates of %d artifacts against the following repositories:", artifacts.size()));
```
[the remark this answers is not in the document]
cheister: I mainly kept it with `String.format` since it matches all of the other logging in the file that uses `verboseLog` instaed of `System.out.println`

private/tools/java/rules/jvm/external/maven/Outdated.java
```diff
@@ -0,0 +1,133 @@
...
+    //<metadata>
+    //  <versioning>
+    //    <latest>1.14.0-SNAPSHOT</latest>
+    //    <release>1.13.0</release>
+    //  </versioning>
+    //</metadata>
+    Element metadataElement = document.getDocumentElement();
+    Element versioningElement = getFirstChildElement(metadataElement, "versioning");
+    if (versioningElement != null) {
+      releaseVersion = versioningElement.getElementsByTagName("release").item(0).getTextContent();
```
cheister, line 56: Note: Another possible feature is allowing people to look for updates from `latest` instead of `release`.

private/tools/java/rules/jvm/external/maven/Outdated.java
```diff
@@ -0,0 +1,133 @@
...
+
+    System.out.println(String.format("Checking for updates of %d artifacts against the following repositories:", artifacts.size()));
+    for (String repository: repositories) {
+      System.out.println(String.format("\t%s", repository));
+    }
+    System.out.println();
+
+    boolean foundUpdates = false;
+
+    for (String artifact: artifacts) {
```
cheister, line 104: It should also be straightforward to make this multithreaded if we wanted to check for several updates at once and reduce runtime.

jin: Add this note as a comment?

coursier.bzl
```diff
@@ -265,16 +277,43 @@ def _get_jq_http_files():
...
     _windows_check(repository_ctx)
 
+    repositories = []
+    for repository in repository_ctx.attr.repositories:
+        repositories.append(json_parse(repository))
+
     artifacts = []
-    for a in repository_ctx.attr.artifacts:
-        artifacts.append(json_parse(a))
+    for artifact in repository_ctx.attr.artifacts:
```
jin, line 315: In the future, please separate refactoring changes from feature changes for a smaller diff :)

private/tools/java/rules/jvm/external/maven/Outdated.java
jin: I think this is reasonable, but to reduce potential confusion, I think the repository where this lookup succeeded should be printed regardless of `RJE_VERBOSE`.

jin (approved):

private/tools/java/rules/jvm/external/maven/Outdated.java
cheister: Do you mind if we leave this only in verbose mode?  Mainly because it keeps the output clean and I'm not sure I know of any other outdated tool that tells you where it found the update by default.

PR #466, opened 2020-10-13
Use merge_jars to add Target-Label to jar because it keeps consistent timestamps

cheister:
Use merge_jars to add Target-Label to the stamped jars because it maintains consistent timestamps.

Also rename all BUILD.bazel files to BUILD.


cheister:
Declining this because merge_jars doesn't work with jars that have zip entries that look the same to a case-insensitive filesystem.

PR #467, opened 2020-10-13
Include classifier in label if present

gmishkin:
My use case is, I have a bzl file with a function that returns a list of all our artifacts. The function is called for both the artifacts attr of maven_install in the WORKSPACE and in another rule where I pass `[artifact(a) for a in get_rje_artifacts()]` as inputs to a custom rule we have. I also use strict_visibility

What happened was that for the artifact that has a classifier, the label returned wasn't valid and so my custom rule failed to build.

I was able to fix it like this.

### DIFF
--- a/defs.bzl
+++ b/defs.bzl
@@ -139,7 +139,10 @@ def maven_install(
 
 def artifact(a, repository_name = DEFAULT_REPOSITORY_NAME):
     artifact_obj = _parse_artifact_str(a) if type(a) == "string" else a
-    return "@%s//:%s" % (repository_name, _escape(artifact_obj["group"] + ":" + artifact_obj["artifact"]))
+    parts = [artifact_obj["group"], artifact_obj["artifact"]]
+    if "classifier" in artifact_obj:
+        parts.append(artifact_obj["classifier"])
+    return "@%s//:%s" % (repository_name, _escape(":".join(parts)))
 
 def maven_artifact(a):
     return artifact(a, repository_name = DEFAULT_REPOSITORY_NAME)


jin (commented):
Hm, the target labels are generated *without* the classifier. I'm not sure if this is the right change.

https://github.com/bazelbuild/rules_jvm_external/blob/08a31605c677956ab3ca6f6668f10b662c28ab6d/private/dependency_tree_parser.bzl#L78
https://github.com/bazelbuild/rules_jvm_external/blob/08a31605c677956ab3ca6f6668f10b662c28ab6d/private/coursier_utilities.bzl#L48

jin:
Do you have a repro example?

gmishkin:
@jin I'll try to construct a minimal repro, but I think the gist was:

- I have a function that returns the list to pass to maven_install, which includes
```starlark
def get_rje_artifacts():
  return [
    maven.artifact("com.google.googlejavaformat", "google-java-format", "1.8", classifier = "all-deps")`,
    # ...
  ]
```

- I have a rule takes in jars that I pass `artifact()` (from defs.bzl) on the above to

```starlark
jar_index(
    name = "external_jar_index",
    jars = [artifact(a) for a in get_rje_artifacts()]
)
```

Without this patch, the label it ends up passing to the jars attr is not valid because it doesn't have the classifier in the labe.

SanyaLuc:
@gmishkin It is sad that nobody from repo owners followed up with you, but I found that classifier already exists, but just slightly not intuitive format https://groups.google.com/g/bazel-discuss/c/ssIF7MBbN-k

PR #468, opened 2020-10-15
Use AddJarManifestEntry to add Target-Label to jars manifest

cheister:
Use custom java class to add Target-Label to the jar manifest so we maintain consistent modification time of the jar Manifest.

### DIFF
--- a/private/rules/jvm_import.bzl
+++ b/private/rules/jvm_import.bzl
@@ -14,35 +14,28 @@ def _jvm_import_impl(ctx):
         fail("Please only specify one jar to import in the jars attribute.")
 
     injar = ctx.files.jars[0]
-    manifest_update_file = ctx.actions.declare_file(injar.basename + ".target_label_manifest", sibling = injar)
-    ctx.actions.expand_template(
-        template = ctx.file._manifest_template,
-        output = manifest_update_file,
-        substitutions = {
-            "{TARGETLABEL}": "%s" % ctx.label,
-        },
-    )
-
     outjar = ctx.actions.declare_file("processed_" + injar.basename, sibling = injar)
     ctx.actions.run_shell(
-        inputs = [injar, manifest_update_file] + ctx.files._host_javabase,
+        inputs = [injar] + ctx.files._add_jar_manifest_entry,
         outputs = [outjar],
-        command = " && ".join([
-            # Make a copy of the original jar, since `jar(1)` modifies the jar in place.
-            "cp {input_jar} {output_jar}".format(input_jar = injar.path, output_jar = outjar.path),
-            # Set the write bit on the copied jar.
-            "chmod +w {output_jar}".format(output_jar = outjar.path),
+        command = "\n".join([
             # If the jar is signed do not modify the manifest because it will
-            # make the signature invalid. Otherwise append the Target-Label
-            # manifest attribute using `jar umf`
-            "(unzip -l {output_jar} | grep -qE 'META-INF/.*\\.SF') || ({jar} umf {manifest_update_file} {output_jar} > /dev/null 2>&1 || true)".format(
-                jar = "%s/bin/jar" % ctx.attr._host_javabase[java_common.JavaRuntimeInfo].java_home,
-                manifest_update_file = manifest_update_file.path,
-                output_jar = outjar.path,
+            # make the signature invalid.
+            "if unzip -l {injar} | grep -qE 'META-INF/.*\\.SF'; then".format(injar = injar.path),
+            "  cp {injar} {outjar}".format(injar = injar.path, outjar = outjar.path),
+            "else",
+            "  set -e",
+            "  {add_jar_manifest_entry} --source {injar} --manifest-entry 'Target-Label:{target_label}' --output {outjar}".format(
+                add_jar_manifest_entry = ctx.attr._add_jar_manifest_entry.files_to_run.executable.path,
+                injar = injar.path,
+                target_label = ctx.label,
+                outjar = outjar.path,
             ),
+            "fi",
         ]),
         mnemonic = "StampJarManifest",
         progress_message = "Stamping the manifest of %s" % ctx.label,
+        tools = [ctx.attr._add_jar_manifest_entry.files_to_run],
     )
 
     if not ctx.attr._stamp_manifest[StampManifestProvider].stamp_enabled:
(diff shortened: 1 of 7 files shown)


private/rules/jvm_import.bzl
```diff
@@ -14,35 +14,28 @@ def _jvm_import_impl(ctx):
...
-            "(unzip -l {output_jar} | grep -qE 'META-INF/.*\\.SF') || ({jar} umf {manifest_update_file} {output_jar} > /dev/null 2>&1 || true)".format(
-                jar = "%s/bin/jar" % ctx.attr._host_javabase[java_common.JavaRuntimeInfo].java_home,
-                manifest_update_file = manifest_update_file.path,
-                output_jar = outjar.path,
+            # make the signature invalid.
+            "if unzip -l {injar} | grep -qE 'META-INF/.*\\.SF'; then".format(injar = injar.path),
+            "  cp {injar} {outjar}".format(injar = injar.path, outjar = outjar.path),
+            "else",
+            "  set -e",
+            "  {add_jar_manifest_entry} --source {injar} --manifest-entry 'Target-Label:{target_label}' --output {outjar}".format(
```
cheister, line 28: Note: I also tried this with `"  {ijar} {injar} {outjar} --nostrip_jar --target_label {target_label}"` where `ijar` is `@bazel_tools//tools/jdk:ijar` but it looks like `ijar` does not preserve compression in jar files so the stamped jar files were typically 2-3 times larger than the unstamped jars.

private/rules/jvm_import.bzl
```diff
@@ -14,35 +14,28 @@ def _jvm_import_impl(ctx):
...
+        command = "\n".join([
             # If the jar is signed do not modify the manifest because it will
-            # make the signature invalid. Otherwise append the Target-Label
-            # manifest attribute using `jar umf`
-            "(unzip -l {output_jar} | grep -qE 'META-INF/.*\\.SF') || ({jar} umf {manifest_update_file} {output_jar} > /dev/null 2>&1 || true)".format(
-                jar = "%s/bin/jar" % ctx.attr._host_javabase[java_common.JavaRuntimeInfo].java_home,
-                manifest_update_file = manifest_update_file.path,
-                output_jar = outjar.path,
+            # make the signature invalid.
+            "if unzip -l {injar} | grep -qE 'META-INF/.*\\.SF'; then".format(injar = injar.path),
```
shs96c, line 24: This assumes that `unzip` is on the machine, which may not be true on Windows. Maybe delegate this to the tool?

private/tools/java/rules/jvm/external/jar/AddJarManifestEntry.java
```diff
@@ -0,0 +1,171 @@
...
+    Objects.requireNonNull(out, "Output path must be set.");
+
+    byte[] buffer = new byte[2048];
+    int bytesRead;
+    boolean manifestUpdated = false;
+
+    try (InputStream fis = Files.newInputStream(source);
+         ZipInputStream zis = new ZipInputStream(fis)) {
+
+      try (OutputStream fos = Files.newOutputStream(out);
```
shs96c, line 77: You can safely merge these two `try` blocks into a single one.

private/tools/java/rules/jvm/external/jar/AddJarManifestEntry.java
```diff
@@ -0,0 +1,171 @@
...
+          Manifest manifest = new Manifest();
+          manifest.getMainAttributes().put(Attributes.Name.MANIFEST_VERSION, "1.0");
+          manifest.getMainAttributes().put(new Attributes.Name("Created-By"), "AddJarManifestEntry");
+          String[] manifestEntryParts = manifestEntry.split(":", 2);
+          manifest.getMainAttributes().put(new Attributes.Name(manifestEntryParts[0]), manifestEntryParts[1]);
+
+          ZipEntry newManifestEntry = new ZipEntry(JarFile.MANIFEST_NAME);
+          newManifestEntry.setTime(DEFAULT_TIMESTAMP);
+          zos.putNextEntry(newManifestEntry);
+          manifest.write(zos);
```
shs96c, line 140: `JarInputStream` will only look at the first two entries of a zip when looking for a manifest. You need to write the manifest first, and _then_ write the contents of the jar.

tests/com/jvm/external/jar/AddJarManifestEntryTest.java
```diff
@@ -0,0 +1,136 @@
...
+import java.util.jar.Manifest;
+import java.util.zip.ZipEntry;
+import java.util.zip.ZipInputStream;
+import java.util.zip.ZipOutputStream;
+
+import static java.nio.charset.StandardCharsets.UTF_8;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
+public class AddJarManifestEntryTest {
```
shs96c, line 28: Can you please add a test that a jar without a manifest and with "more than two entries" gets a manifest that that `JarInputStream` can find, please?

shs96c (changes requested):
Looks good, and the idea makes sense. One question: why do we limit this to adding only one manifest entry? Would it be desirable to support adding more than one? The answer may well be "no", but it's good to ask.

private/tools/java/rules/jvm/external/jar/AddJarManifestEntry.java
cheister: Yeah, these can be merged but it I thought keeping them separate made the blocks more clear.

private/tools/java/rules/jvm/external/jar/AddJarManifestEntry.java
cheister: This is a good use case but I don't think `JarInputStream` is used by anything in Bazel.  Is the goal to support external tools that might use `JarInputStream` to work with stamped jars?

tests/com/jvm/external/jar/AddJarManifestEntryTest.java
cheister: There is a test for a jar without a manifest at https://github.com/bazelbuild/rules_jvm_external/blob/master/tests/unit/manifest_stamp/BUILD#L9  but it would be good to add a test for the `JarInputStream`

private/tools/java/rules/jvm/external/jar/AddJarManifestEntry.java
cheister: Updated so if the manifest is missing it will be added as the first entry in the new jar.

private/rules/jvm_import.bzl
shs96c: We really need to avoid using `unzip` here, since it won't work on Windows. Sadly, it doesn't look like `@bazel_tools//tools/zip:zipper` supports just listing files, but we could use `jar tvf {injar}` here relatively safely.

shs96c (changes requested):

private/rules/jvm_import.bzl
cheister: I would rather address that in another PR.  This PR was just meant to replace using jar to add the Target-Label entry to the manifest in a cache friendly manner.  The `unzip` method was here before and as far as I know worked fine on Windows.  Removing it is probably better but I think it's a separate PR.  

Please also keep in mind that this feature is not enabled by default.  People have to opt into the using the stamping functionality with the `--@rules_jvm_external//settings:stamp_manifest` option.

jin: `unzip` is also a suggested prereq for using Bazel on Windows: https://docs.bazel.build/versions/master/install-windows.html#installing-compilers-and-language-runtimes

We can add that as a setup step in the README.

jin (approved):

private/rules/jvm_import.bzl
shs96c: A follow up PR is fine.

shs96c (approved):
A follow up PR for the `unzip` thing is fine. LGTM

PR #469, opened 2020-10-15
Use JarHelper jar entry timestamp logic in MergeJars

cheister:
I came across this difference between MergeJars and   https://github.com/bazelbuild/bazel/blob/master/src/java_tools/buildjar/java/com/google/devtools/build/buildjar/jarhelper/JarHelper.java#L124 when I was looking at jar tools.  Seems like it's better to use the same logic as JarHelper here.


shs96c (approved):
Any chance that you could add a test for this, please? I realise that code isn't the best tested out there, and my apologies for that.

Issue #448, opened 2020-08-21
Error publishing Maven with proto dependencies during javadoc


jastice:
We're trying to publish the bazel-bsp adapter to maven using the `java_export` rule, and having some troubles. Our experiment setting it up is here: https://github.com/JetBrains/bazel-bsp/tree/rules_jvm_external
When trying to build with `bazel run --stamp \
  --define "maven_repo=file://$HOME/.m2/repository" //:bsp.publish`

it fails on the docs:
```
INFO: From Action bsp-docs.jar:
[...]

PR #475, opened 2020-10-23
Rework how maven project jars are calculated

shs96c:
Previously, we assumed that all the information required for a project
all code to be included was inside the main repository. This was put
in place largely to prevent code depending on `rules_proto` from
pulling in the classes from protobuf itself.

Issue #448 demonstrated that this assumption wasn't correct.

We now gather all the JavaInfos that should be included in the
artifact, as well as those of dependencies of the artifact (called
"dep_infos" in the MavenInfo provider) We then calculate the
difference of those sets to determine the input files to add to the
generated project jar.

However, this is not sufficient on its own in the case of
protocbuf. As such, we also scan for classes to exclude from the
project jar by examaning the contents of the "dep_infos" runtime jars.

Furthermore, jars that are created by aspects have no way of
expressing what their maven dependencies are. To work around this,
rules may now express `MavenHintInfo` which will be used to help
calculate the contents of the artifact jars and maven dependencies.

### DIFF
--- a/private/rules/maven_project_jar.bzl
+++ b/private/rules/maven_project_jar.bzl
@@ -1,13 +1,14 @@
-load(":has_maven_deps.bzl", "MavenInfo", "has_maven_deps")
+load(":has_maven_deps.bzl", "MavenInfo", "calculate_artifact_jars", "calculate_artifact_source_jars", "has_maven_deps")
 
-def _combine_jars(ctx, merge_jars, inputs, output):
+def _combine_jars(ctx, merge_jars, inputs, excludes, output):
     args = ctx.actions.args()
     args.add("--output", output)
     args.add_all(inputs, before_each = "--sources")
+    args.add_all(excludes, before_each = "--exclude")
 
     ctx.actions.run(
         mnemonic = "MergeJars",
-        inputs = inputs,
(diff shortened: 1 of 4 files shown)


private/rules/maven_project_jar.bzl
```diff
@@ -17,12 +18,26 @@ def _maven_project_jar_impl(ctx):
...
+    artifact_srcs = calculate_artifact_source_jars(info)
+
     # Merge together all the binary jars
     bin_jar = ctx.actions.declare_file("%s.jar" % ctx.label.name)
-    _combine_jars(ctx, ctx.executable._singlejar, info.artifact_jars.to_list(), bin_jar)
+    _combine_jars(
+        ctx,
+        ctx.executable._singlejar,
+        artifact_jars,
+        depset(transitive = [ji.transitive_runtime_jars for ji in info.dep_infos.to_list()]).to_list(),
```
jin, line 31: This line looks like it's creating the transitive dep_info jars to be excluded from artifact_jars, where artifact_jars already excluded this same list of transitive_runtime_jars from `calculate_artifact_jars`. Am I misreading something or is this doing redundant work?

jin (approved):

PR #481, opened 2020-11-13
Introduce a deps and a setup macro

shs96c:
This allows us to pull in dependencies that we may need to ensure that
rules_jvm_external functions. At present, we use this to make sure we
use `rules_java` for our own java targets.


jin:
It looks like the `@rules_java` load requirement is put on hold: https://github.com/bazelbuild/bazel/issues/8746#issuecomment-703514000

We should still have the scaffolding, maybe we could use this PR to build Hasher and MergeJars from source, instead of relying on a deploy jar?

shs96c:
I don’t see any harm in using those rules, but building from source would be good. Want me to do that in a follow up PR?

I think we’ll need to have the maven_install in the repositories file.

jin:
I'd prefer to add a `load` dependency if we don't need to, and it's not clear what the plan is for enforcing `@rules_java`. Certainly not happening in Bazel 4.0, from release notes.

We can add the scaffolding for deps and setup in this PR, and follow up with a PR to build those binaries from source.

jin:
Discussion about `load`ing `rules_java` on bazel-discuss today: https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/bazel-discuss/XNvpWcge4AE/dVPDrQDhBQAJ

repositories.bzl
```diff
@@ -0,0 +1,14 @@
+load("//:defs.bzl", "maven_install")
+
+def rules_jvm_external_deps():
+    maven_install(
+        name = "rules_jvm_external_deps",
+        artifacts = [
+        ],
+        maven_install_json = "@rules_jvm_external//:rules_jvm_external_deps_install.json",
+        repositories = [
```
jin, line 9: Parameterize this with default values?

jin (approved):

repositories.bzl
shs96c: Done.

jin (approved):

PR #484, opened 2020-11-20
Generate Coursier credentials from a netrc file

### LINKED ISSUE #483: The netrc file can not be used to generate `maven_install.json` with private repositories
Currently, the only way to generate the `maven_install.json` file with private repositories is to provide the credentials in the repository url, which is a committed file.

Instead, `rules_jvm_external` should be able to pick credentials from the netrc file.

nlewo:
A simple parser parse the user netrc file and generate a list of
credentials that can be used by Coursier.

Fixes #483

### DIFF
--- a/coursier.bzl
+++ b/coursier.bzl
@@ -24,6 +24,7 @@ load(
     "COURSIER_CLI_SHA256",
     "JQ_VERSIONS",
 )
+load("@bazel_tools//tools/build_defs/repo:utils.bzl", "read_netrc")
 
 _BUILD = """
 package(default_visibility = ["//visibility:{visibility}"])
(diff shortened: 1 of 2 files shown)


dmivankov:
There is a workaround - configure both netrc and default coursier location, like ~/.config/coursier/credentials.properties
But very nice to have single source of truth :+1: 

Seems to be safe to add extra --credentials https://get-coursier.io/docs/other-credentials default locations would still be considered

coursier.bzl
```diff
@@ -261,6 +261,50 @@ def get_home_netrc_contents(repository_ctx):
...
+    for _ in sanitized:
+        if i >= len(sanitized):
+            break
+        token = sanitized[i]
+        if token == "default":
+            machine = None
+            i = i + 1
+        elif token == "machine":
+            if i + 1 > len(sanitized):
+                fail("netrc file is not correct (error on token {})".format(token))
```
cheister, line 284: Should this fail the whole run or just print a warning and continue without any credentials set?

tests/unit/coursier_test.bzl
```diff
@@ -476,6 +477,42 @@ def _get_coursier_cache_or_default_enabled_with_home_dot_coursier_directory_test
...
+    asserts.equals(
+        env,
+        ["example.com foo:bar",
+         "with-newline foo:bar",
+         "machine foo:bar",
+         ],
+        generate_coursier_credentials_from_netrc(netrc_content)
+    )
+    return unittest.end(env)
+
```
cheister, line 502: Do you mind adding a test for an invalid netrc file since you've covered those cases in your code?

coursier.bzl
```diff
@@ -261,6 +261,50 @@ def get_home_netrc_contents(repository_ctx):
                 return repository_ctx.read(netrcfile)
     return ""
 
+def generate_coursier_credentials_from_netrc(netrc_content):
+    """Generate a list of coursier credentials from a netrc file. Only
+    'machine' entries containing a login and a password are taken into
+    account. The 'default' entry is ignored.
+    """
+    spacified = netrc_content.replace("\n", " ").replace("\t", " ")
+    sanitized = [e for e in spacified.split(" ") if e != ""]
```
gibfahn, line 270: Would be good to ignore `#` style comments. The two netrc parsers in Bazel both do this:

https://github.com/bazelbuild/bazel/blob/4c0febc02fab5e9fda3f40532a62854fb1c45631/tools/build_defs/repo/utils.bzl#L228

https://github.com/bazelbuild/bazel/pull/12130

gibfahn: Also the second one checks the `$NETRC` env var as an alternative to the hardcoded `~/.netrc`, would be nice to support that too.

coursier.bzl
gibfahn: Would be nice if netrc files with non-standard syntax didn't fail the build.

mrkkrp: I changed the function to use the `read_netrc` helper provided by Bazel.

tests/unit/coursier_test.bzl
mrkkrp: I changed the function to use the `read_netrc` helper provided by Bazel, so we are no longer testing parsing of `.netrc` files. This should resolve this thread.

coursier.bzl
mrkkrp: I changed the function to use the read_netrc helper provided by Bazel.

mrkkrp:
@nlewo Could you please tell @googlebot that you consent, like it says in the comment above?

nlewo:
@googlebot I consent.

shs96c:
@jin any idea why the clabot hasn't cleared the `cla:no` label?

shs96c (approved):
LGTM once the CLA bot gives the thumbs up.

jin:
There are two commits in this PR, one each by @nlewo and @mrkkrp. I forced a rescan and the CLA bot says that it needs the consent from @mrkkrp. 

@mrkkrp, could you respond with "@googlebot I consent." as well?

mrkkrp:
@googlebot I consent.

mrkkrp:
Please hold on with merging for now, there may be a minor problem with this patch. I'll look into this today.

mrkkrp:
I have pushed a new commit. Now the `coursier_fetch` and `pinned_coursier_fetch` rules inspect the filesystem to get credentials. To ensure that events like adding a `.netrc` file after an initial run have effect, we need to set the `local` attribute of the repository rules to `True`. I'm not totally sure what kind of undesirable consequences this may have though.

mrkkrp:
I don't think that the CI failures are related, but I may be wrong.

mrkkrp:
Nevermind, I removed the commit that adds `local = True` to the repo rules definition. I did not fix the issue. Apparently `bazel clean` is necessary in both cases. IMHO, the patch is good to go.

shs96c:
@mrkkrp This error looks valid: https://buildkite.com/bazel/rules-jvm-external/builds/1767#1e42648b-8170-4c69-ade7-99d88af96a7a/101-112

dmivankov:
>  Downloading https://releases.bazel.build/1.0.1/release/bazel-1.0.1-linux-x86_64
could be that read_netrc isn't available in bazel 1.0.1

FaBrand:
Is this still being pursued? I had to patch coursier as well to deal with authentication.
One followup that i encountered after this is that the user that did the actual pinning would end up in the pinned.json file:
```json
 "url": "https://<user>@<repository>/
 ```
I could remove it by reusing `remove_auth_from_url` for the `url` part in the parsed dep_tree.
```diff
+    for artifact in dep_tree["dependencies"]:
+        url = artifact.get("url", None)
+        if url == None:
+            continue
+        artifact.update({"url": remove_auth_from_url(url)})
```
As well as an additional method to do the same with the `file` path. That one contained the user as well
```json
"file": "v1/https/<username>%40<repo_url>/
```

shs96c:
This is still something we want to land.

nlewo:
fyi, I no longer work on this topic so i could not terminate this work :(

shs96c:
This has been superseded by #778

PR #485, opened 2020-11-24
Update documentation to use the latest names

jermowery:
Documentation was updated here: https://github.com/bazelbuild/rules_jvm_external/releases/tag/3.3
But it was not updated in this readme


jin:
cc @shs96c let's remove this from the README until we make a new release?

shs96c (approved):
Makes sense to leave these out until we actually make the release. Thank you for the PR!

Issue #494, opened 2020-12-04
add a flag to fetch javadoc for maven artifacts


erickj:
just a feature request, that it would be nice to be able to use coursier's ability to fetch javadocs as additional artifacts. this requires additional support, similar to the `fetch_sources` flags because the underlying coursier command must be modified to use the `--javadoc` flag

erickj:
created pull request https://github.com/bazelbuild/rules_jvm_external/pull/495/commits/6b7822a72532003a1c2bef6c01cb896369c7394a

jin:
Merged in #495.

PR #495, opened 2020-12-04
Adds a flag to fetch javadoc targets (#494)

erickj:
Adds flag fetch_javadoc to maven_install rules. When used, javadoc for
all targets will be fetched by coursier and placed in a similarly
named filegroup.


shs96c (approved):
Thanks for the PR! This looks good to me.

Issue #507, opened 2021-01-06
bazel run @maven//:outdated crashes with NPE (4.0, java11)


dhalperi:
WORKSPACE:

```
load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")

RULES_JVM_EXTERNAL_TAG = "4.0"

RULES_JVM_EXTERNAL_SHA = "31701ad93dbfe544d597dbe62c9a1fdd76d81d8a9150c2bf1ecf928ecdf97169"

http_archive(
    name = "rules_jvm_external",
    sha256 = RULES_JVM_EXTERNAL_SHA,
    strip_prefix = "rules_jvm_external-%s" % RULES_JVM_EXTERNAL_TAG,
    url = "https://github.com/bazelbuild/rules_jvm_external/archive/%s.zip" % RULES_JVM_EXTERNAL_TAG,
)

load("@rules_jvm_external//:defs.bzl", "maven_install")

maven_install(
    artifacts = [
        "javax.inject:javax.inject:1",
    ],
    repositories = [
        "https://repo1.maven.org/maven2",
    ],
)
```

java --version:

```
openjdk 11.0.8 2020-07-14
OpenJDK Runtime Environment AdoptOpenJDK (build 11.0.8+10)
OpenJDK 64-Bit Server VM AdoptOpenJDK (build 11.0.8+10, mixed mode)
```

os: `macOS Catalina 10.15.7 (19H114)`

bazel version:
```
Bazelisk version: 1.6.1
Build label: 3.7.2
Build target: bazel-out/darwin-opt/bin/src/main/java/com/google/devtools/build/lib/bazel/BazelServer_deploy.jar
Build time: Thu Dec 17 17:02:20 2020 (1608224540)
Build timestamp: 1608224540
Build timestamp as int: 1608224540
```

Output:
```
bazel run @maven//:outdated
INFO: Invocation ID: 7b754f1a-1fa5-4587-9a43-59f1e871c245
INFO: Analyzed target @maven//:outdated (3 packages loaded, 5 targets configured).
INFO: Found 1 target...
Target @maven//:outdated up-to-date:
  bazel-bin/external/maven/outdated
INFO: Elapsed time: 0.330s, Critical Path: 0.01s
INFO: 1 process: 1 internal.
INFO: Build completed successfully, 1 total action
INFO: Build completed successfully, 1 total action
Checking for updates of 1 artifacts against the following repositories:
	https://repo1.maven.org/maven2

[...]

PR #508, opened 2021-01-06
Outdated: handle more maven-metadata.xml files

dhalperi:
Fix bazelbuild/rules_jvm_external#507

The current `maven-metadata.xml` for `javax.inject:javax.inject` is missing a
release or latest tag.

```xml
    <metadata>
      <groupId>javax.inject</groupId>
      <artifactId>javax.inject</artifactId>
      <version>1</version>
      <versioning>
        <versions>
          <version>1</version>
        </versions>
        <lastUpdated>20100720032040</lastUpdated>
      </versioning>
    </metadata>
```

In this case, grab the first version in the list.

With the new code paths, the control flow was getting hard to track,
so I switched to returning at decision time.

---
This change is [Reviewable](https://reviewable.io/reviews/bazelbuild/rules_jvm_external/508)


dhalperi:
Some notes:
* I regenerated the deploy jar, but I guess a maintainer will want to (re)-do this for security?
* The stack trace in #507 does not line up with the existing source code (L56/L58). So I'm assuming the deploy jar was out of date.
* I didn't see any unit tests to update, please point me there if I should.

jin:
@cheister shall we switch Outdated.jar to be built from source when it's requested? I'm also thinking of doing the same for the hasher.

dhalperi:
Would that require users to have to add outdated's deps in some way to our own WORKSPACE files?

jin:
Yep. That's already needed today if you want to use `java_export` to Google Cloud Storage, with the deps.bzl setup macros.

cheister:
Thanks for the PR.  I'll look into adding better tests for this class

private/tools/java/rules/jvm/external/maven/Outdated.java
```diff
@@ -44,22 +42,48 @@ public static String getReleaseVersion(String repository, String groupId, String
...
+
+    // Note: we may want to add a flag to allow people to look for updates against
+    // "latest" instead of "release"
+    NodeList release = versioningElement.getElementsByTagName("release");
+    if (release != null && release.getLength() > 0) {
+      return release.item(0).getTextContent();
     }
-    return releaseVersion;
+
+    // No release info, default to the first version in the list.
```
cheister, line 79: Based on some example metadata files it seems like this should default to the last version in the list instead of the first. e.g https://repo1.maven.org/maven2/com/fasterxml/jackson/core/jackson-core/maven-metadata.xml, https://repo1.maven.org/maven2/com/google/guava/guava/maven-metadata.xml, https://repo1.maven.org/maven2/org/apache/commons/commons-lang3/maven-metadata.xml

cheister:
Do you mind rebasing this PR to get the tests from https://github.com/bazelbuild/rules_jvm_external/pull/511 ?

private/tools/java/rules/jvm/external/maven/Outdated.java
dhalperi: Done. I also fixed the logic and tested it.

dhalperi:
> Do you mind rebasing this PR to get the tests from #511 ?

Done. I also rewrote the (commented-out) test to use classpath resources so it doesn't need Internet.

cheister (approved):
Thanks for the fix!

dhalperi:
Great! Just checking, what are next steps? Obviously, I won't be able to (squash&)merge the PR.

jin (approved):

PR #509, opened 2021-01-08
Detect when pinned inputs have changed and stop the build

shs96c:
Optionally force people to repin if the inputs to maven_install have
changed. This is done by adding a new `fail_if_repin_required`
attribute to `maven_install`. In order to preserve existing behaviour,
this is currently set to `False`, but users can update to `True` if
required.

If a repin is not required, a warning is printed to the console and
the build continues as expected. If a repin is required, the build is
failed and steps to rectify the situation are printed.

In addition to the `maven_install` attribute the `REPIN`
environment variable is checked. This allows one to update the inputs
without needing to modify the `WORKSPACE` file.


README.md
```diff
@@ -206,6 +206,25 @@ prefix (e.g.`@unpinned_maven` or `@unpinned_<your_maven_install_name>`). For
...
+It can be easy to forget to update the `maven_install.json` lock file
+when updating artifacts in a `maven_install`. Normally,
+rules_jvm_external will print a warning to the console and continue
+the build when this happens, but by setting the
+`fail_if_repin_required` attribute to `True`, this will be treated as
+a build error, causing the build to fail. When this attribute is set,
+it is possible to update the `maven_install.json` file using:
+
+```shell
+$ REPIN=1 bazel run @unpinned_maven//:pin
```
cheister, line 220: It would be nice if running :pin did not require setting an env var, since running :pin implies that your are trying to re-generate the json file. 

I'm not sure how this would work in the repository rule though since I don't think there is a way to ask the repository context what action is being run.  You could maybe do something where the pin.sh script generates a file that the `_pinned_coursier_fetch_impl` could look for an delete but I'm not sure if it's reliable to share an output from one repository rule to another this way.   @jin do you have any ideas?

coursier.bzl
```diff
@@ -301,6 +316,12 @@ def _add_outdated_files(repository_ctx, artifacts, repositories):
         executable = True,
     )
 
+def _is_failing_build_if_inputs_change(repository_ctx):
```
cheister, line 319: nit: should this method just be called `_fail_if_repin_required` ?

coursier.bzl
```diff
@@ -346,7 +367,29 @@ def _pinned_coursier_fetch_impl(repository_ctx):
...
     dep_tree = maven_install_json_content["dependency_tree"]
 
-    dep_tree_signature = dep_tree.get("__AUTOGENERATED_FILE_DO_NOT_MODIFY_THIS_FILE_MANUALLY")
+    # Then, check to see if we need to repin our deps because inputs have changed
+    if dep_tree.get("__INPUT_ARTIFACTS_HASH") == None:
+        print("NOTE: %s_install.json does not contain a signature of the required artifacts. " % repository_ctx.name +
+              "This feature ensures that the build does not use stale dependencies when the inputs " +
+              "have changed. To generate this signature, run 'bazel run @unpinned_%s//:pin'." % repository_ctx.name)
+    else:
+        computed = compute_dependency_inputs_signature(repository_ctx.attr.artifacts)
```
cheister, line 376: nit: mind calling this `computed_artifacts_hash` ?

coursier.bzl
```diff
@@ -862,7 +905,9 @@ def _coursier_fetch_impl(repository_ctx):
         artifact.update({"sha256": shas[str(repository_ctx.path(file))]})
 
     dep_tree.update({
-        "__AUTOGENERATED_FILE_DO_NOT_MODIFY_THIS_FILE_MANUALLY": _compute_dependency_tree_signature(dep_tree["dependencies"]),
+        "__AUTOGENERATED_FILE_DO_NOT_MODIFY_THIS_FILE_MANUALLY": "THERE_IS_NO_DATA_ONLY_ZUUL",
+        "__RESOLVED_ARTIFACTS_HASH": _compute_dependency_tree_signature(dep_tree["dependencies"]),
```
cheister, line 909: Should the`__RESOLVED_ARTIFACTS_HASH` and `__INPUT_ARTIFACTS_HASH ` entries only be added if `fail_if_repin_required` is set to true?

README.md
shs96c: You're right that's the ideal way for this to work. Unfortunately for us, `bazel run` extends `bazel build`, and this requires the `WORKSPACE` file to be valid and properly configured, so we're caught in a catch-22: we're not correct so we can't run, but if we can't run, we can't be correct. *sigh*

To mitigate this, the default behaviour is to emit a warning rather than fail the build. The user must set `fail_if_repin_required` to `True` to get the build-breaking behaviour. To try and help more, we also print a warning to the console if the build needs to be repinned, and that command is listed.

coursier.bzl
shs96c: Done.

coursier.bzl
shs96c: Done

coursier.bzl
shs96c: I've retained the current behaviour we already have. I'd prefer not to change that in this PR, as we're already doing enough stuff here.

README.md
jin, line 220: Consider calling it `RJE_REPIN` or `RJE_FORCE_REPIN` to make it clear that this is a RJE env var. Useful in bashrc/zshrc(s).

jin, line 220: Could you make a note about how this feature is enabled only if your maven_install.json version is a newer one?

Maybe we could version maven_install.json with a `version` field with the value of the current RJE major release directly to make it clear for users.

jin (approved):
LGTM. The approach with the env vars and per-repo attribute is reasonable.

README.md
gibfahn: I think the idea was to have a common env var that would repin everything, that we could share between multiple _external repos.

shs96c: I've added a note that people should repin if they see the error. Personally, I think it's fine to say to people "repin when you see a new version of `rules_jvm_external`"

PR #511, opened 2021-01-09
Add tests for Outdated

cheister:
Add tests for outdated class.

Follow on to https://github.com/bazelbuild/rules_jvm_external/pull/508 so we can test fixes for the outdated function

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -26,6 +26,7 @@ tasks:
       - "//..."
       # These tests are currently incompatible with RBE
       - "-//tests/integration/override_targets"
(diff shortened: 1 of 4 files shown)


.bazelci/presubmit.yml
```diff
@@ -26,6 +26,7 @@ tasks:
       - "//..."
       # These tests are currently incompatible with RBE
       - "-//tests/integration/override_targets"
+      - "-//tests/com/jvm/external/maven:OutdatedTest"
```
cheister, line 29: Excluding this test from RBE for now since it looks like the RBE build doesn't allow connections to outside maven servers.

jin (approved):

PR #513, opened 2021-01-25
Remove artifact symlinks from .publish script to avoid files clash

dmivankov:
This allows to have multiple .publish targets within same run
invocation, for example using multirun rule like this:
```python
multirun(
  name = "publish-all",
  commands = [
    "target1.publish",
    "target2.publish",
    "target3.publish",
  ],
)
```
Without this fix bazel would complain that pom.xml and other symlinks
are overwriting each other between targets.

Uploader executable can be kept as shared symlink as it is same across
targets.

MavenPublisher doesn't seem to rely on exact filenames or paths, so
should be a safe change.


dmivankov:
not needed anymore after https://github.com/bazelbuild/rules_jvm_external/pull/552/files

PR #515, opened 2021-02-01
Ivy repo support

khogeland:
Coursier supports Ivy repo patterns, and this almost works out-of-the-box with these rules, but I ran into one place that needs special handling.

Also, I'm not sure if it's appropriate to include the change to `strip_packaging_and_classifier`, but it's not apparent why it's opinionated about which classifiers to remove. All of our Ivy artifacts have classifiers, and don't get friendly `group_name` labels because of it 🤷


jin:
Thank you for your PR! Please add tests to demonstrate that this works with the features we have in rules_jvm_external today, especially with integration tests (see the root WORKSPACE file).

khogeland:
@jin I added a test to verify `maven_install` will resolve Ivy artifacts and their dependencies, is that sufficient? I'm not sure what features you're referring to.

PR #516, opened 2021-02-02
Add aliases with classifiers stripped

khogeland:
If a resolved artifact includes a classifier, it won't have a `group_name`-style target created for it. This is an issue when a dependency does not specify a classifier, but the resolved artifact does. For example:

- Artifact `com.example:foo:classifier:1` is published. It is the only artifact matching `com.example:foo:1`.
- Artifact `com.example:bar` declares a dependency on `com.example:foo:1`
- Coursier resolves `com.example:foo:1`, to the artifact with coords `com.example:foo:classifier:1`
- A rule for the resolved `foo` is generated, named `@maven//:com_example_foo_classifier`
- The generated rule for `bar` depends on `@maven//:com_example_foo`, which does not exist
- Bazel explodes

This PR doesn't remove/change any targets that are currently created, but in the above example it will create an alias from `@maven//:com_example_foo` -> `@maven://com_example_foo_classifier` (and another alias with version numbers)

### DIFF
--- a/private/coursier_utilities.bzl
+++ b/private/coursier_utilities.bzl
@@ -27,23 +27,29 @@ SUPPORTED_PACKAGING_TYPES = [
     "scala-jar",
 ]
 
-def strip_packaging_and_classifier(coord):
-    # Strip some packaging and classifier values based on the following maven coordinate formats
-    # groupId:artifactId:version
-    # groupId:artifactId:packaging:version
-    # groupId:artifactId:packaging:classifier:version
+def strip_packaging(coord):
     coordinates = coord.split(":")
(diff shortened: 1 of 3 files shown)


cheister:
Thanks for the PR!  Do you have a test case we could add to [regression_testing](https://github.com/bazelbuild/rules_jvm_external/blob/master/WORKSPACE#L184) for this?

jin:
+1 to @cheister. Please add tests.

khogeland:
This is a very specific edge case that I cannot find a public example of (and we have a suitable workaround internally), so I'm closing this. If anyone runs into this in the wild, I'd be happy to revisit it.

PR #518, opened 2021-02-10
Generate pseudo-compile jars for bazel java compilation

shs96c:
Normally, bazel uses `ijar` to prepare a jar containing just
ABI-entries for classfiles. These are stable when non-API breaking
changes are made, and so allow bazel to compile code faster, and are
known as "compile jars"

Because of https://github.com/bazelbuild/bazel/issues/4549
`rules_jvm_external` doesn't use `ijar`, and instead uses the
downloaded jar as the "compile jar". Normally, this is fine, but Bazel
4.0.0 now sets `-Xlint:path` for javac invocations, and this means
that if there's a `Class-Path` manifest entry in a compile jar, the
jars within _that_ will be checked. This can lead to noisy builds:
https://github.com/bazelbuild/bazel/issues/12968

This change generates a "pseudo-compile jar" for `jvm_import`
targets. All it does is repack the input jar, excluding the
`META-INF/MANIFEST.MF` file. This means that we should avoid
compilation problems whilst still working well with Kotlin projects.

### DIFF
--- a/private/rules/jvm_import.bzl
+++ b/private/rules/jvm_import.bzl
@@ -15,27 +15,38 @@ def _jvm_import_impl(ctx):
 
     injar = ctx.files.jars[0]
     outjar = ctx.actions.declare_file("processed_" + injar.basename, sibling = injar)
-    ctx.actions.run_shell(
-        inputs = [injar] + ctx.files._add_jar_manifest_entry,
+    args = ctx.actions.args()
+    args.add_all(["--source", injar, "--output", outjar])
+    args.add_all(["--manifest-entry", "Target-Label:{target_label}".format(target_label = ctx.label)])
+    ctx.actions.run(
+        executable = ctx.executable._add_jar_manifest_entry,
+        arguments = [args],
+        inputs = [injar],
         outputs = [outjar],
-        command = "\n".join([
-            # If the jar is signed do not modify the manifest because it will
-            # make the signature invalid.
-            "if unzip -l {injar} | grep -qE 'META-INF/.*\\.SF'; then".format(injar = injar.path),
-            "  cp {injar} {outjar}".format(injar = injar.path, outjar = outjar.path),
-            "else",
-            "  set -e",
-            "  {add_jar_manifest_entry} --source {injar} --manifest-entry 'Target-Label:{target_label}' --output {outjar}".format(
-                add_jar_manifest_entry = ctx.attr._add_jar_manifest_entry.files_to_run.executable.path,
-                injar = injar.path,
-                target_label = ctx.label,
-                outjar = outjar.path,
-            ),
-            "fi",
-        ]),
         mnemonic = "StampJarManifest",
         progress_message = "Stamping the manifest of %s" % ctx.label,
-        tools = [ctx.attr._add_jar_manifest_entry.files_to_run],
+    )
+
+    compilejar = ctx.actions.declare_file("header_" + injar.basename, sibling = injar)
+    args = ctx.actions.args()
+    args.add_all(["--source", outjar, "--output", compilejar])
+    args.add_all(["--remove-entry", "Class-Path"])
+    ctx.actions.run(
+        executable = ctx.executable._add_jar_manifest_entry,
+        arguments = [args],
+        inputs = [outjar],
+        outputs = [compilejar],
+    )
+
+    compilejar = ctx.actions.declare_file("header_" + injar.basename, sibling = injar)
+    args = ctx.actions.args()
+    args.add_all(["--source", outjar, "--output", compilejar])
+    args.add_all(["--remove-entry", "Class-Path"])
+    ctx.actions.run(
+        executable = ctx.executable._add_jar_manifest_entry,
+        arguments = [args],
+        inputs = [outjar],
+        outputs = [compilejar],
     )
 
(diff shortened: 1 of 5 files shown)


cheister:
I feel like just deleting the entire manifest.mf file will have other follow on problems for the jars that might set things like `Multi-Release` to `true` or other jar meta info.

What about just deleting the `Class-Path` attributed from the manifest.mf file if it exists?  Seems like we could even adapt `AddJarManifestEntry` to delete the entry if it's there?

shs96c:
I'd not thought about the case where we'd be dealing with a multi-release jar. Good catch.

I've reworked the PR to use `AddJarManifestEntry` and made some changes to facilitate that. I've reworked the new test to use the `AddJarManifestEntry` but left it otherwise unchanged from the original version.

cheister:
Looks good.  I wonder if we should just combine the part where we add the `Target-Label` with the part that removes the `Class-Path` attribute since we'll now be modifying the jar always.

The other condition we check for in `Target-Label` case is if the the jar is signed then we don't modify the manifest because it will make the signature invalid.  We probably want that for the `Class-Path` attribute as well so seems like we should just combine the two?

shs96c:
The `compilejar` is only used for compilation, and the runtime checks aren't done on it. As long as it contains the right class files in the right place, everything should be fine. 

I may modify `AddJarManifestEntry` to handle the checks for us, so we can unify the logic into one place.

private/rules/jvm_import.bzl
```diff
@@ -15,27 +15,38 @@ def _jvm_import_impl(ctx):
...
+    args.add_all(["--source", outjar, "--output", compilejar])
+    args.add_all(["--remove-entry", "Class-Path"])
+    ctx.actions.run(
+        executable = ctx.executable._add_jar_manifest_entry,
+        arguments = [args],
+        inputs = [outjar],
+        outputs = [compilejar],
+    )
+
+    compilejar = ctx.actions.declare_file("header_" + injar.basename, sibling = injar)
```
cheister, line 41: I'm guessing this wasn't meant to be duplicated?

private/rules/jvm_import.bzl
```diff
@@ -15,27 +15,38 @@ def _jvm_import_impl(ctx):
 
     injar = ctx.files.jars[0]
     outjar = ctx.actions.declare_file("processed_" + injar.basename, sibling = injar)
-    ctx.actions.run_shell(
-        inputs = [injar] + ctx.files._add_jar_manifest_entry,
+    args = ctx.actions.args()
+    args.add_all(["--source", injar, "--output", outjar])
+    args.add_all(["--manifest-entry", "Target-Label:{target_label}".format(target_label = ctx.label)])
```
cheister, line 20: Can this be combined with the call below so we only call add_jar_manifest_entry once?

private/rules/jvm_import.bzl
```diff
@@ -15,27 +15,38 @@ def _jvm_import_impl(ctx):
 
     injar = ctx.files.jars[0]
     outjar = ctx.actions.declare_file("processed_" + injar.basename, sibling = injar)
-    ctx.actions.run_shell(
-        inputs = [injar] + ctx.files._add_jar_manifest_entry,
+    args = ctx.actions.args()
```
jin, line 18: Do you mind writing comments to explain why this is needed, especially for Bazel 4.0.0? (forcing the javac flag is another problem, but let's handle it separately)

private/rules/jvm_import.bzl
shs96c: Ooops. Fixed :)

private/rules/jvm_import.bzl
shs96c: Desirable as that may be, the stamped jar and the compile jar are used for different purposes. The output jar will be supplied as a transitive dependency and is included in the list of runtime jars.

The compile jar is just used by javac for compilation. Normally, bazel uses `ijar` to generate this, and that strips out the bodies of methods, leaving only the ABI in place. For this case, we could remove everything that wasn't the class files, and reduce the manifest down to indicate whether or not this is a multi-release jar. In previous reviews, you've asked for the manifest to be left intact, which is done, other than stripping the `Class-Path`, since that's the thing that causes bazel to the `bad path element` warnings this PR was designed to remove :)

private/rules/jvm_import.bzl
shs96c: Done

private/rules/jvm_import.bzl
```diff
@@ -15,27 +15,33 @@ def _jvm_import_impl(ctx):
...
+    args.add_all(["--source", outjar, "--output", compilejar])
+    # We need to remove the `Class-Path` entry since bazel 4 forces `javac`
+    # to run `-Xlint:path` no matter what other flags are passed. Bazel
+    # manages the classpath for us, so the `Class-Path` manifest entry isn't
+    # needed. Worse, if it's there and the jars listed in it aren't found,
+    # the lint check will emit a `bad path element` warning. We get quiet and
+    # correct builds if we remove the `Class-Path` manifest entry entirely.
+    args.add_all(["--remove-entry", "Class-Path"])
+    ctx.actions.run(
```
jin, line 40: Could you also add a mnemonic and progress message?

jin (approved):

PR #519, opened 2021-02-10
Make read_coordinates public

shs96c:
It's useful for other tools to be able to extract maven coordinates
from tags (for example, Selenium's Java Module generation) Making this
function public makes this possible.


cheister (approved):

jin:
Should this function be moved out of the private package, then? I'd like our Starlark public API to be well-defined, at least through package location.

shs96c:
@jin would you like me to expose the method from `defs.bzl`?

jin:
Yep, just so that folks don't load symbols from `@rules_jvm_external//private`.

jin (approved):

Issue #517, opened 2021-02-03
PSA: Bintray and JCenter are being shut down on May 1st


Bencodes:
Jcenter is being shut down on May 1st and the usages in this repo will need to be switched over to alternatives before May 1st

https://jfrog.com/blog/into-the-sunset-bintray-jcenter-gocenter-and-chartcenter/

cheister:
Fixed with https://github.com/bazelbuild/rules_jvm_external/pull/523

Issue #375, opened 2020-02-13
//tests/integration:UnsafeSharedCacheTest failed to build on Ubuntu


meteorcloudy:
Since https://buildkite.com/bazel/rules-jvm-external/builds/1199, this test is failing to build with the following error message:
```
Opening zip "external/unsafe_shared_cache/v1/https/repo1.maven.org/maven2/com/google/guava/guava/27.0-jre/guava-27.0-jre.jar": open(): No such file or directory
Unable to open Zip file external/unsafe_shared_cache/v1/https/repo1.maven.org/maven2/com/google/guava/guava/27.0-jre/guava-27.0-jre.jar: No such file or directory
```

I did some investigation and found that it will only fail to build when remote cache is enabled. 
Because it was green on the same commit before build 1199, the issue looks like caused by infra change. 

Another point worth notice is `external/unsafe_shared_cache/v1/https/repo1.maven.org/maven2/com/google/guava/guava/27.0-jre/guava-27.0-jre.jar` is a symlink pointing to another location. The file at the destination does exist.

/cc @jin @philwo

meteorcloudy:
The symlink points to a file under `/tmp/custom_coursier_cache`, it looks like the sandbox prevent ijar from accessing that file?

meteorcloudy:
@philwo The test failure of rules_jvm_external is failing due to this issue again.

philwo:
The sandbox mounts a fresh, empty tmpfs on /tmp, which means that the /tmp from the host system is shadowed and files in there cannot be accessed. If this is what's happening (and it sounds like that's the problem here), then this is working as intended.

But the strange thing is, this has been the case on Bazel CI since it exists, so I wonder why it only now fails (and why it sometimes works!?).

meteorcloudy:
I guess it happened to work for some reason and the result gets cached, so it keeps passing until the cache got cleaned or some Bazel change forced it to rerun.

jin:
>  so I wonder why it only now fails (and why it sometimes works!?).

I added support for custom `COURSIER_CACHE` locations in https://github.com/bazelbuild/rules_jvm_external/commit/7badfbe1947ee53b393947251d1fcafdab714b60. 

If /tmp is used for tmpfs, is there an on-disk writable location I can use for the tests?

philwo:
[...]

PR #523, opened 2021-02-16
Remove references to jcenter.bintray.com which is being shutdown on May 1, 2021

cheister:
Fix for https://github.com/bazelbuild/rules_jvm_external/issues/517


jin (approved):

cheister:
@jin Looks like `//tests/com/jvm/external:UnsafeSharedCacheTest` is failing on OSX.  I'm not sure if https://github.com/bazelbuild/rules_jvm_external/issues/375 is now a problem for OSX as well and if `//tests/com/jvm/external:UnsafeSharedCacheTest"` should be commented out for OSX in `.bazelci/presubmit.yml` as well?

jin:
The failure is 

```
exec ${PAGER:-/usr/bin/less} "$0" || exit 1
Executing tests from //tests/com/jvm/external:UnsafeSharedCacheTest
-----------------------------------------------------------------------------
JUnit4 Test Runner
.E
Time: 0.031
There was 1 failure:
1) test_jarsOnClassPath_areInTheSharedCache(com.jvm.external.UnsafeSharedCacheTest)
java.lang.AssertionError: 
Expected: a string containing "/tmp/custom_coursier_cache"
     but: was "file:/private/var/tmp/_bazel_buildkite/0f1ff3967c5ac2afc60797312589b600/execroot/rules_jvm_external/bazel-out/darwin-fastbuild/bin/external/maven/v1/https/repo1.maven.org/maven2/org/hamcrest/hamcrest/2.1/processed_hamcrest-2.1.jar"
	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:18)
	at org.junit.Assert.assertThat(Assert.java:964)
	at org.junit.Assert.assertThat(Assert.java:930)
	at com.jvm.external.UnsafeSharedCacheTest.test_jarsOnClassPath_areInTheSharedCache(UnsafeSharedCacheTest.java:35)

FAILURES!!!
Tests run: 1,  Failures: 1


BazelTestRunner exiting with a return value of 1
JVM shutdown hooks (if any) will run now.
The JVM will exit once they complete.

-- JVM shutdown starting at 2021-02-16 07:45:59 --
```

which is different from the linked issue. I think this has something to do with a recent mac upgrade on CI. Looking.

jin:
Could you try changing this line:

https://github.com/bazelbuild/rules_jvm_external/blob/367eb9ae5be8acd5a069b84c4672a2ed1d0bda82/.bazelci/presubmit.yml#L79

as well as UnsafeSharedCacheTest itself to use /Users/buildkite/coursier_cache, instead of the /tmp dir? I'm guessing that something stateful happened in the CI upgrade that caused something odd with /tmp.

cheister:
Ok, sorted it out the UnsafeSharedCacheTest was only working before because of a typo in the test looking for `jcentray.bintray.com` instead of `jcenter.bintray.com`, this filtered out the hamcrest dependencies.  Since only the `guava` dependency is included in https://github.com/bazelbuild/rules_jvm_external/blob/master/tests/integration/BUILD#L66, that is the only one we should look for in the test with the coursier_cache path.

jin:
Yikes, good catch.

PR #526, opened 2021-02-17
Enable maven_install to use the Starlark version of aar_import to facilitate migrating to the Starlark Android rules

ahumesky:
See also https://github.com/bazelbuild/bazel/issues/5354


coursier.bzl
```diff
@@ -1048,6 +1064,8 @@ pinned_coursier_fetch = repository_rule(
         "jetify_include_list": attr.string_list(doc = "List of artifacts that need to be jetified in `groupId:artifactId` format. By default all artifacts are jetified if `jetify` is set to True.", default = JETIFY_INCLUDE_LIST_JETIFY_ALL),
         "additional_netrc_lines": attr.string_list(doc = "Additional lines prepended to the netrc file used by `http_file` (with `maven_install_json` only).", default = []),
         "fail_if_repin_required": attr.bool(doc = "Whether to fail the build if the maven_artifact inputs have changed but the lock file has not been repinned.", default = False),
+        "starlark_aar_import": attr.bool(default = False, doc = "Whether to use the native or Starlark version of aar_import"),
```
jin, line 1067: How about "use_starlark_android_rules"?

coursier.bzl
```diff
@@ -1048,6 +1064,8 @@ pinned_coursier_fetch = repository_rule(
...
+        "aar_import_bzl_label": attr.string(default = DEFAULT_AAR_IMPORT_LABEL, doc = "The label (as a string) to use to import aar_import from"),
```
jin, line 1068: I don't think we need an attribute for this, right? This can be an implementation detail of the other attribute within the function bodies of coursier_fetch and pinned_coursier_fetch.

ahumesky: So the idea here is that the generated build file needs to have the label to the bzl file for aar_import, and the sample in the rules_android readme gives an http_archive with the name "build_bazel_rules_android", but repos could have a different name (e.g. just "rules_android"). Also, the label within the rules_android repo may differ between versions of the android rules. The current code requires "//android:rules.bzl", but the pre-alpha code requires "//rules:rules.bzl". So this provides some flexibility during migrations.

coursier.bzl
ahumesky: Sounds good

coursier.bzl
```diff
@@ -119,6 +127,12 @@ def _relativize_and_symlink_file(repository_ctx, absolute_path):
         repository_ctx.symlink(absolute_path, repository_ctx.path(artifact_relative_path))
     return artifact_relative_path
 
+def _get_aar_import_statement_or_empty_str(repository_ctx):
```
jin, line 130: Is there a way to prevent folks from using this attribute to inject arbitrary strings other than what it should be (a load stmt)? Maybe check for rules_android, aar_import, and newlines?

jin (approved):

coursier.bzl
```diff
@@ -30,10 +30,18 @@ package(default_visibility = ["//visibility:{visibility}"])
 
 load("@rules_jvm_external//private/rules:jvm_import.bzl", "jvm_import")
 load("@rules_jvm_external//private/rules:jetifier.bzl", "jetify_aar_import", "jetify_jvm_import")
```
jin, line 32: Oh wait, should the aar_import in jetifier.bzl be handled the same way?

coursier.bzl
ahumesky: How about using the Label() constructor to try to parse the label. If it's not a valid label, an error should be raised

jin: Yep, sounds like a reasonable guardrail.

coursier.bzl
ahumesky: Ah, yes, thank you for pointing that out. It's a little trickier but yes, that will need to be handled too. I'll add some more changes to take care of that.

coursier.bzl
ahumesky: Done.

coursier.bzl
ahumesky: Added support for injecting the aar_import rule to jetifier_aar_import, PTAL

jin (approved):

PR #527, opened 2021-02-26
Make documentation `bzl_library` publicly visible

shs96c:
This allows anyone who depends on `rules_jvm_external` and imports
"something" from these rules into their own rules to be able to
generate stardoc of their own functions.

I've filed https://github.com/bazelbuild/stardoc/issues/93 to track
this issue. Once it's fixed, we can close this.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -53,6 +53,9 @@ bzl_library(
         "//third_party/bazel_json/lib:json_parser.bzl",
     ],
     visibility = [
+        # This library is only visible to allow others who depend on
+        # `rules_jvm_external` to be able to document their code using
+        # stardoc.
         "//visibility:public",
     ],
 )
--- a/BUILD
+++ b/BUILD
@@ -52,6 +52,9 @@ bzl_library(
         "//settings:stamp_manifest.bzl",
         "//third_party/bazel_json/lib:json_parser.bzl",
     ],
+    visibility = [
+        "//visibility:public",
+    ],
 )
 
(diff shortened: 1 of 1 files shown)


jin:
I'm not comfortable for exposing this target because it's meant for RJE-internal docs. If you really need a file, could you create a temporary target with the minimum set of files you need, and expose that instead?

shs96c:
The way that stardoc works it that anything that transitively depends upon `rules_jvm_external` needs to include our `bzl` files in a `bzl_library`. Either we need to expose all those files publicly using something like `exports_files` or we make this rule visible. Exporting the files ties those rules even more intimately to the internals of this ruleset, which is undesirable.

Without this change, any rule that depends upon `rules_jvm_external` (such as [Selenium's own rules](https://github.com/SeleniumHQ/selenium/blob/trunk/java/private/export.bzl#L1-L8)) cannot be documented using stardoc. That's why I filed https://github.com/bazelbuild/stardoc/issues/93 and that's why I'm making this change.

shs96c:
Also, to create the temporary target, one would need to patch `rules_jvm_external` since there are build files in our tree, preventing us from just "reaching in and taking what we need"

jin:
Looks like there's a plan to support bazelbuild/stardoc#93, but work needs to be done to get there. For this PR, could you at least document that the bzl_library is public for stardoc and stardoc only?

jin (approved):

PR #528, opened 2021-02-26
Run buildifier over the repo. No logical changes

shs96c:


cheister (approved):
In buildifier we trust

PR #530, opened 2021-03-04
Fix broken test //tests/integration:version_interval_deps_test.

zhengwei143:
The flip of --incompatible_genquery_use_graphless_query causes genquery output to be sorted lexicographically, which causes this diff test to break.


jin:
Thanks for the fix! It looks like this hasn't been enabled on a release binary, so maybe pass in the graphless query flag to the project bazelrc or CI configuration?

meteorcloudy:
> so maybe pass in the graphless query flag to the project bazelrc or CI configuration?

Does --incompatible_genquery_use_graphless_query exist in 4.0? If so, we can do that. Otherwise, we probably need to disable the test for now, because it breaks with either Bazel@HEAD or Bazel@latest_release

jin:
Alternate solution that's independent of the flag or bazel version: https://github.com/bazelbuild/rules_jvm_external/pull/531

Issue #529, opened 2021-03-03
rules_jvm_external broken in Bazel downstream due to the flip of --incompatible_genquery_use_graphless_query


meteorcloudy:
https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/1939#457fc168-72ac-4485-985e-4c4a1e1919bb
```
Executing tests from //tests/integration:version_interval_deps_test
-----------------------------------------------------------------------------
1,2c1,4
< @version_interval_testing//:io_grpc_grpc_netty_shaded
< @version_interval_testing//:v1/https/repo1.maven.org/maven2/io/grpc/grpc-netty-shaded/1.29.0/grpc-netty-shaded-1.29.0.jar
---
> @version_interval_testing//:com_google_android_annotations
> @version_interval_testing//:com_google_code_gson_gson
> @version_interval_testing//:com_google_errorprone_error_prone_annotations
> @version_interval_testing//:io_grpc_grpc_api
4c6
< @version_interval_testing//:v1/https/repo1.maven.org/maven2/io/grpc/grpc-core/1.29.0/grpc-core-1.29.0.jar
---
> @version_interval_testing//:io_grpc_grpc_netty_shaded
6,9c8,9
< @version_interval_testing//:io_grpc_grpc_api
< @version_interval_testing//:com_google_errorprone_error_prone_annotations
< @version_interval_testing//:com_google_code_gson_gson
< @version_interval_testing//:com_google_android_annotations
---
> @version_interval_testing//:v1/https/repo1.maven.org/maven2/io/grpc/grpc-core/1.29.0/grpc-core-1.29.0.jar
> @version_interval_testing//:v1/https/repo1.maven.org/maven2/io/grpc/grpc-netty-shaded/1.29.0/grpc-netty-shaded-1.29.0.jar
FAIL: files "tests/integration/version-interval-deps.golden.unix" and "tests/integration/version_interval_deps" differ
```

https://buildkite.com/bazel/bazel-auto-sheriff-face-with-cowboy-hat/builds/447
Caused by https://github.com/bazelbuild/bazel/commit/0f1e3032d513ccf26b27803382cec2c758eb5799

Related issue: https://github.com/bazelbuild/bazel/issues/12464

zhengwei143:
The flip to use graphless query causes the output to be sorted in lexicographical order (from dependency order previously), which is what I assume is causing the issue here when running the diff. Let me look into a quick fix for the test.

[...]

PR #531, opened 2021-03-05
Sort genquery outputs to be compatible with --incompatible_genquery_use_graphless_query

jin:
This presorts the outputs of genquery before testing them, making them independent of graphless query or the bazel version under test.

Fixes: #529


meteorcloudy (approved):
Ah, thanks!

Issue #522, opened 2021-02-11
Is there a way to use a local maven repo


capezzbr:
Basically, I'd like to replicate what is possible on Gradle via

```groovy
repositories {
    mavenLocal()
    maven {
        // All of React Native (JS, Obj-C sources, Android binaries) is installed from npm
        url "$rootDir/../node_modules/react-native/android"
    }
}
```

So far I've tried

```
maven_install(
    artifacts = [
        "com.facebook.react:react-native:0.63.3",
    ],
    repositories = [
        "file:///Users/bruno/repo/node_modules/react-native/android",
    ],
)
```

and 

```
android_library(
    name = "main_application",
    srcs = [
        "MainActivity.java",
        "MainApplication.java",
    ],
    manifest = "AndroidManifest.xml",
    resource_files = glob(["res/**"]),
    deps = [
        "@maven//:com_facebook_react_react_native",
    ],
)
```

but it doesn't seem to be working, and I am getting this error

```
(14:46:46) ERROR: [...] depends on @maven//:com_facebook_react_react_native in repository @maven which failed to fetch. no such package '@maven//': Only artifacts downloaded over http(s) are supported: com.facebook.react:react-native:0.63.3
```

cheister:
I'm not sure coursier has a way to add local maven repositories for fetching artifacts.  They added a `m2local` option with https://github.com/coursier/coursier/issues/667 but that does not allow arbitrary directories to be maven repositories.

capezzbr:
I see, thanks for the input. This is what I have so far.

I've copied the local maven repo content under `~/.m2/repository`, so that I have this directory structure:

![image](https://user-images.githubusercontent.com/3534630/108112159-8cf9bd00-704a-11eb-845f-aed01fb35ca3.png)

and then added the dependency as

```
load("@rules_jvm_external//:specs.bzl", "maven")

maven_install(
    artifacts = [
        "com.facebook.react:react-native:0.63.3",
    ],
    use_unsafe_shared_cache = True,
    repositories = [
        maven.repository(
            "file:///Users/bruno/.m2/repository",
        ),
        ...
    ],
)
```

which errors out as 

```
(11:27:18) ERROR: An error occurred during the fetch of repository 'maven':
   Traceback (most recent call last):
[...]

Issue #305, opened 2019-12-04
Can't add maven local under `maven_install`'s repository


ThomasCJY:
I'm trying to add maven local repository under  `maven_install` but seems like it doesn't support this currently.

What I did is 
```
maven_install(
    name = "mvn",
    artifacts = [
        # Add all Android third-party deps here:
    ],
    repositories = [
        "https://maven.google.com",
        "file://localpath",
    ],
    version_conflict_policy = "pinned",
)
```
The file:// path will not be recognized

jin:
The repository url is passed to Coursier as-is. If Coursier doesn't support this feature, we don't the ability to do so as well.

utikeev:
@jin 
It seems that you explicitly forbid protocols other than `http(s)`:
https://github.com/bazelbuild/rules_jvm_external/blob/master/coursier.bzl#L657

joprice:
I'd like to be able to have some pattern for consuming local dependencies for migration purposes: I have sbt-plugins that have a core library that needs to be used from a genrule.

benjaminRomano:
Is this something we are comfortable with adding support for? @jin

PR #532, opened 2021-03-10
Add support for m2Local repository

tokudu:
This PR addresses the following open issues:
https://github.com/bazelbuild/rules_jvm_external/issues/522
https://github.com/bazelbuild/rules_jvm_external/issues/305

Coursier supports passing `m2Local` in the list of repositories, but jvm external can't properly handle the dependency tree output, since it returns pointers to the original files within `~/.m2/repository` directory. 

Sample usage (you probably have to add `fail_on_missing_checksum`, since mavel local is likely missing checksums:
```
    maven_install(
        name = "maven",
        artifacts = [ 
             ...
        ],
        repositories = [
            "m2Local"
        ],
        fail_on_missing_checksum = True
    )
```

### DIFF
--- a/coursier.bzl
+++ b/coursier.bzl
@@ -105,7 +105,7 @@ def _normalize_to_unix_path(path):
 # Relativize an absolute path to an artifact in coursier's default cache location.
 # After relativizing, also symlink the path into the workspace's output base.
 # Then return the relative path for further processing
-def _relativize_and_symlink_file(repository_ctx, absolute_path):
+def _relativize_and_symlink_file_in_coursier_cache(repository_ctx, absolute_path):
     # The path manipulation from here on out assumes *nix paths, not Windows.
     # for artifact_absolute_path in artifact_absolute_paths:
     #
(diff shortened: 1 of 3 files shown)


jin:
Thanks for working on this! Nice to see that it actually works with Coursier support. Could you please add some tests, especially integration tests? I would be more comfortable accepting this with its code path tested. You may need to modify .bazelci/presubmit.yml to run some setup script to pre-populate a local .m2 repository for those integration tests as well.

tokudu:
Will do. This needs to be cleaned up a bit to support the pinned maven_install.json path. Still doing some more testing.

tokudu:
Ok both pinned and unpinned cases should be working now. Will follow up next week with some integration tests

kamildoleglo:
Hi! Are there any updates on this? ☺️

tokudu:
I haven't had the time to add the integration tests as it ended up being pretty convoluted, but we have been using for a while in our internal fork and it works well.

tokudu:
You are welcome to contribute and we can team up to get this over the finish line.

smocherla-brex:
@tokudu I'd like to take a stab at getting this merged, if that's ok with you?

tokudu:
Absolutely!

jin:
#695 has been merged. Thanks all!

tjbanghart:
Hey yall glad to see this! I haven't been able to get this to work for my purposes though and could use a reality check.

I have a gradle project I am publishing to my maven local repo via [Maven Publish Plugin](https://docs.gradle.org/current/userguide/publishing_maven.html#publishing_maven:install) 

In my bazel project I am pulling in maven dependencies with the following helper:

```
def _maven_pinned_nexus_install(name, artifacts = [], excluded_artifacts = []):
    maven_install(
        name = name,
        artifacts = artifacts,
        repositories = [
            "https://nexusrepo.<my_proxy>.com/repository/maven-public/",
            "m2Local"
        ],
        fail_if_repin_required = True,
        fail_on_missing_checksum = False,
        fetch_sources = False,
        maven_install_json = "//internal/repositories:%s_install.json" % name,
        strict_visibility = False,
        version_conflict_policy = "pinned",
        excluded_artifacts = excluded_artifacts,
    )

```

While repining I get the following error:
```
invalid label '/usr/local/tjbanghart/.m2/<path_to_jar>' in element 0 of attribute 'jars' in 'jvm_import' rule: invalid target name '/usr/local/tjbanghart/.m2/<path_to_jar>': target names may not start with '/'
```
Is `m2Local` only compatible with certain types of artifacts? It seems like the `file://` target prefix  is missing? Any ideas are welcome. Thank you 🙇

smocherla-brex:
@tjbanghart It would seem that m2Local absolute path isn't converted to a relative path and symlinked as it should be [here](https://github.com/bazelbuild/rules_jvm_external/blob/master/coursier.bzl#L1014). I can't personally reproduce it myself, but if you have a simple repro, I could take a look.

If you could also perhaps share the output of this 
```
bazel query @unpinned_maven//... --output=build
```
that might also be more clarifying as to what's going wrong.

PR #533, opened 2021-03-11
Add support for custom java install for coursier

cristi-:
We are using this to avoid java versions mismatch between environments (where some machines may have jdk1.8 or jdk14 instead of jdk11), which causes pinning to generate different results.

A rough example of how we use this with a `repository_rule`
```
JAVA_EXEC_TEMPLATE = """
#!/usr/bin/env bash
exec %s "$@"
"""

BUILD_FILE_CONTENTS = """
exports_files(["java"])
"""

def _jdkenv(repository_ctx):
    cwd = repository_ctx.execute(["pwd"]).stdout.strip()
    java_bin = download_jdk11()
    repository_ctx.file(
        "java",
        content = JAVA_EXEC_TEMPLATE % java_bin,
        executable = True,
    )

    repository_ctx.file("BUILD", content = BUILD_FILE_CONTENTS)
jdkenv = repository_rule(
    implementation = _jdkenv,
    environ = ["JAVA_HOME"],
    doc = """Repository rule for detecting jdk11:
@repo//:java -- Java binary
""",
```
And in `maven_install`:
```
load("@rules_jvm_external//:defs.bzl", "maven_install")
load("//tools/build:jdk.bzl", "jdkenv")

jdkenv(
    name = "jdk11",
)
maven_install(
        name = "jvm_es_deps",
        ....
        java_binary = "@jdk11//:java",
)   
```

### DIFF
--- a/coursier.bzl
+++ b/coursier.bzl
@@ -142,7 +142,10 @@ def _generate_java_jar_command(repository_ctx, jar_path):
     coursier_opts = repository_ctx.os.environ.get("COURSIER_OPTS", "")
     coursier_opts = coursier_opts.split(" ") if len(coursier_opts) > 0 else []
 
-    if java_home != None:
+    if repository_ctx.attr.java_binary != None:
+        java = repository_ctx.path(repository_ctx.attr.java_binary)
+        cmd = [java, "-noverify", "-jar"] + coursier_opts + _get_java_proxy_args(repository_ctx) + [jar_path]
+    elif java_home != None:
         # https://github.com/coursier/coursier/blob/master/doc/FORMER-README.md#how-can-the-launcher-be-run-on-windows-or-manually-with-the-java-program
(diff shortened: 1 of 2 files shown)


shs96c:
Is there any reason not to use the `JAVA_BASE` environment variable? That should mean your build and coursier would share the same JDK.... https://docs.bazel.build/versions/master/be/make-variables.html#custom_variables

cristi-:
> Is there any reason not to use the `JAVA_BASE` environment variable? That should mean your build and coursier would share the same JDK.... https://docs.bazel.build/versions/master/be/make-variables.html#custom_variables

I was looking at https://github.com/bazelbuild/rules_jvm_external/issues/445#issuecomment-678446422 and I thought it would be impossible to access the path as provided by JAVA_BASE. Do you think there is a way to get to provide the path to coursier ?

PR #534, opened 2021-03-11
Add a `deploy_env` attribute to `java_export`

shs96c:
This allows `java_export` to be used to generate things such as
plugins for Apache Spark or jar files for use in JEE servers by
excluding provided dependencies from the generated jars. This is
analagous to maven's `provided` scope, and maps to the `deploy_env`
attribute of `java_binary`.

The difference between `java_export` and `java_binary`'s `deploy_env`
is that `java_export` allows anything that provides a `JavaInfo` as an
input to `deploy_env`. This is because the expected use-cases are for
things that are normally provided as regular imports from a maven
dependency, rather than a custom binary.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -1,3 +1,18 @@
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+# API Reference
+
+- [Basic functions](#basic-functions)
+  - [javadoc](#javadoc)
+  - [java_export](#java_export)
+  - [maven_install](#maven_install)
+- [Maven specification functions](#maven-specification-functions)
+  - [maven.repository](#mavenrepository)
+  - [maven.artifact](#mavenartifact)
+  - [maven.exclusion](#mavenexclusion)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
 # Basic functions
 
 These are the basic functions to get started.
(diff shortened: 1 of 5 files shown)


docs/api.md
```diff
@@ -1,18 +1,3 @@
-<!-- START doctoc generated TOC please keep comment here to allow auto update -->
```
jin, line 1: Was this ran with https://github.com/bazelbuild/rules_jvm_external/blob/master/scripts/generate_docs.sh?

tests/integration/java_export/check-deploy-env.sh
```diff
@@ -0,0 +1,22 @@
...
+  source "$(grep -sm1 "^$f " "$0.exe.runfiles_manifest" | cut -f2- -d' ')" 2>/dev/null || \
+  { echo>&2 "ERROR: cannot find $f"; exit 1; }; f=; set -e
+# --- end runfiles.bash initialization v2 ---
+
+set -euox pipefail
+
+classes_file=$(rlocation rules_jvm_external/tests/integration/java_export/classes.txt)
+
+if grep -q Dependency.class "$classes_file"; then
+  exit 1
```
jin, line 17: echo an error message here?

tests/integration/java_export/check-deploy-env.sh
```diff
@@ -0,0 +1,22 @@
...
+fi
+
+if ! grep -q Main.class "$classes_file"; then
+  exit 2
```
jin, line 21: Why are the exit codes different?

jin (approved):

tests/integration/java_export/check-deploy-env.sh
shs96c: Done

tests/integration/java_export/check-deploy-env.sh
shs96c: To differentiate why the test failed. Echoing a message instead.

docs/api.md
shs96c: It is now :) Good catch. Thank you!

Issue #536, opened 2021-03-15
rules_jvm_external CI failing with Bazel@HEAD


c-parsons:
rules_jvm_external CI failing with Bazel@HEAD as of our CI run on yesterday, Sun, Mar 14.

Example failure:
https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/1954#ee6bc81f-60ea-4720-8e2e-5c0444f885be

Log snippet:
```
1) initializationError(com.google.android.apps.common.testing.suite.AndroidDeviceTestSuite)
[...]

PR #537, opened 2021-03-16
Set RBE machine type with KVM for instrumentation tests

jin:
Fixes https://github.com/bazelbuild/rules_jvm_external/issues/536

Same fix applied to android/testing-samples in https://github.com/android/testing-samples/pull/327


philwo (approved):
Thank you!

Issue #541, opened 2021-03-25
Failure when HTTP_PROXY doesn't have a scheme


alexeagle:
my env had `HTTP_PROXY=localhost:8888`

and then subsequent builds fail with 
```
ERROR: Error fetching repository: Traceback (most recent call last):
	File "/private/var/tmp/_bazel_alex.eagle/03c50bd9f989d25b88a1be300d53820c/external/rules_jvm_external/coursier.bzl", line 513, column 24, in _pinned_coursier_fetch_impl
		_add_outdated_files(repository_ctx, artifacts, repositories)
	File "/private/var/tmp/_bazel_alex.eagle/03c50bd9f989d25b88a1be300d53820c/external/rules_jvm_external/coursier.bzl", line 331, column 58, in _add_outdated_files
		"{proxy_opts}": " ".join(_get_java_proxy_args(repository_ctx)),
	File "/private/var/tmp/_bazel_alex.eagle/03c50bd9f989d25b88a1be300d53820c/external/rules_jvm_external/coursier.bzl", line 167, column 31, in _get_java_proxy_args
		return get_java_proxy_args(http_proxy, https_proxy, no_proxy)
	File "/private/var/tmp/_bazel_alex.eagle/03c50bd9f989d25b88a1be300d53820c/external/rules_jvm_external/private/proxy.bzl", line 39, column 37, in get_java_proxy_args
		proxy_user = _get_proxy_user(http_proxy)
	File "/private/var/tmp/_bazel_alex.eagle/03c50bd9f989d25b88a1be300d53820c/external/rules_jvm_external/private/proxy.bzl", line 2, column 33, in _get_proxy_user
		netloc = url.split("://", 1)[1]
Error: index out of range (index is 1, but sequence has 1 elements)
```

All this split logic needs to handle the case of a URI with no scheme https://github.com/bazelbuild/rules_jvm_external/blob/master/private/proxy.bzl#L2

PR #542, opened 2021-03-30
Fix proxy parsing when HTTP_PROXY doesn't have a scheme (#541)

cheister:
Fix for HTTP_PROXY parsing when url doesn't have a scheme https://github.com/bazelbuild/rules_jvm_external/issues/541

### DIFF
--- a/private/proxy.bzl
+++ b/private/proxy.bzl
@@ -1,30 +1,35 @@
+def _strip_url_scheme(url):
+    return url.split("://", 1)[1] if "://" in url else url
+
 def _get_proxy_user(url):
-    netloc = url.split("://", 1)[1]
-    if "@" in netloc:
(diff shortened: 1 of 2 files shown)


private/proxy.bzl
```diff
@@ -1,30 +1,32 @@
...
+    no_scheme_url = url.split("://", 1)[1] if "://" in url else url
+    netloc = no_scheme_url.split("@")[-1]
     if ":" in netloc:
         return netloc.split(":")[0]
     else:
         return netloc
 
 def _get_proxy_port(url):
-    netloc = url.split("://", 1)[1].split("/")[0].split("@")[-1]
+    no_scheme_url = url.split("://", 1)[1] if "://" in url else url
```
jin, line 28: nit: extract this line into a function?

jin (approved):

PR #543, opened 2021-03-31
Fix stamp_enabled flag so the jar isn't modified if the flag is set

cheister:
After the changes for https://github.com/bazelbuild/rules_jvm_external/pull/518 the `--@rules_jvm_external//settings:stamp_manifest` flag only kept the stamp out of the outjar and not the compile jar.  This update means the flag will always keep the jar from getting stamped if the flag is set.

Follow up question:  Do we want to keep this flag anymore?


shs96c (approved):
Good catch. Thank you! LGTM

PR #552, opened 2021-04-26
Don't use symlinks in maven_publish DefaultInfo

sitaktif:
Symlinks in DefaultInfo is not supported by remote build execution at
the moment (see https://github.com/bazelbuild/bazel/issues/13355).

Removing use of symlinks as a workaround for the time being since the
issue above is not trivial to fix.

---

The following command (assuming `bazel-rbe` is a wrapper to run bazel with remote build execution)...

```
bazel-rbe run //my/package:mytarget.publish --define "maven_repo=..." --remote_download_outputs=toplevel
```

...only works once this patch is applied. Before then, it would fail to find the artifact when being run.

There is currently a workaround consisting in running with `--remote_download_outputs=all` but it causes many unnecessary downloads.

### DIFF
--- a/private/rules/maven_publish.bzl
+++ b/private/rules/maven_publish.bzl
@@ -11,7 +11,7 @@ MavenPublishInfo = provider(
 _TEMPLATE = """#!/usr/bin/env bash
 
 echo "Uploading {coordinates} to {maven_repo}"
-./uploader {maven_repo} {gpg_sign} {user} {password} {coordinates} pom.xml artifact.jar source.jar doc.jar
+{uploader} {maven_repo} {gpg_sign} {user} {password} {coordinates} {pom} {artifact_jar} {source_jar} {javadoc}
 """
 
 def _maven_publish_impl(ctx):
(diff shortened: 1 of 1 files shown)


private/rules/maven_publish.bzl
```diff
@@ -39,13 +44,12 @@ def _maven_publish_impl(ctx):
...
             runfiles = ctx.runfiles(
-                symlinks = {
-                    "artifact.jar": ctx.file.artifact_jar,
-                    "doc.jar": ctx.file.javadocs,
-                    "pom.xml": ctx.file.pom,
-                    "source.jar": ctx.file.source_jar,
-                    "uploader": ctx.executable._uploader,
-                },
+                files = [
+                    ctx.file.pom,
```
shs96c, line 48: Nit: can we please alphabetise this list?

shs96c (changes requested):

shs96c (approved):
LGTM. Thank you!

sitaktif:
Thanks!

PR #563, opened 2021-05-24
Don't export via deps, only via exports

illicitonion:
https://github.com/bazelbuild/rules_jvm_external/pull/557 replaced using
`JavaInfo.transitive_exports` by instead using
`HasMavenDeps.transitive_exports` but it changed how this was computed.

The JavaInfo traversal code:
https://github.com/bazelbuild/bazel/blob/5a647bb234ba9817f51fdb2b4b5973a0602f51a4/src/main/java/com/google/devtools/build/lib/rules/java/JavaCommon.java#L341-L357
only follows the `exports` attribute, and merges together each of the
transitive exports it finds that way.

The replacement starlark code was traversing each of `deps`, `exports`,
and `runtime_deps`, looking for anything which may have been exported
_to_ those targets, which exports significantly more targets.

This adds a unit test to prevent future regressions.

### DIFF
--- a/private/rules/has_maven_deps.bzl
+++ b/private/rules/has_maven_deps.bzl
@@ -98,7 +98,7 @@ _gathered = provider(
     ],
 )
 
-def _extract_from(gathered, maven_info, dep):
+def _extract_from(gathered, maven_info, dep, include_transitive_exports):
     java_info = dep[JavaInfo] if dep and JavaInfo in dep else None
 
     gathered.all_infos.append(maven_info)
@@ -108,6 +108,8 @@ def _extract_from(gathered, maven_info, dep):
             gathered.dep_infos.append(dep[JavaInfo])
         else:
             gathered.artifact_infos.append(dep[JavaInfo])
+            if include_transitive_exports:
+                gathered.transitive_exports.append(maven_info.transitive_exports)
 
 def _has_maven_deps_impl(target, ctx):
     if not JavaInfo in target:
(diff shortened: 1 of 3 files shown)


comius:
> /cc @comius in case the change in behaviour was intended

No, it was unintentional. Thanks for the fix and the test!

illicitonion:
Sorry for the delay - I've been out on PTO :)

> > /cc @comius in case the change in behaviour was intended
> 
> No, it was unintentional; I assume you have a case that is broken/different now.

I do indeed!

> Didn't the old code here https://github.com/bazelbuild/rules_jvm_external/pull/557/files#diff-4c35294998684c53cec071ff25a96119a3545b55579a7a21233c83870ed37c6cL56-L58 also collect transitive_exports from deps and runtime_deps? Because artifact_infos are gathered over _ASPECT_ATTRS?

I have spent quite a while staring at this code, and I _think_ I now understand the key difference. `JavaInfo` itself does filtering for us to only follow `exports` attributes when computing `transitive_exports` - so `transitive_exports` is arguably a misleading name - it's really `transitive_exports_via_my_exports` - target X only has something appear in its `transitive_exports` if the route from X to that target is in the `exports` attribute of X.

Your change had the effect of removing this "only via the exports of X" filtering for `maven_project_jar` targets - before we consulted its target's `target[JavaInfo].transitive_exports` which performed the "only via my exports" filtering, and now we don't because we're reading its `target[MavenInfo].transitive_exports` which includes `transitive_exports` from its target's `deps` attribute.

I've just pushed a commit which I think makes this more clear - when traversing the tree to gather, we only accumulate the `transitive_exports` if the attribute we went via to get to the target we're accumulating from was itself `exports`. I also added a test on the `transitive_runtime_deps` the `maven_project_jar` sees, as this is the root cause of my failure in a private repo.

shs96c (approved):
Sorry it took so long to review. Thank you @illicitonion!

Issue #472, opened 2020-10-20
Use built-in Bazel JSON parser


gmishkin:
With our maven_install.json it takes about 10-20 seconds to parse the JSON. Bazel just added a JSON parser so if this package switched to it, that bottleneck would go away.

jin:
This is not yet released in a bazel version, but would be a nice feature for someone to pick up on :)

gmishkin:
It's released!

PR #566, opened 2021-06-04
Replace bazel_json with json.decode for #472

JaredNeil:
For our 32k line JSON lock file evaluation of `_pinned_coursier_fetch_impl` went from 14000ms to 250ms.
There are some downsides to this change:

- Requires Bazel >= 4.0.0
- Loses the nice error message if the lock file is malformed

If we don't want to require 4.0.0, we could keep the old parser around and only use this if it's available. I couldn't figure out how to check if the `json` module is available or a simple way to check the bazel version (the one I know about seemed quite [complicated](https://github.com/bazelbuild/rules_nodejs/blob/stable/internal/common/check_bazel_version.bzl)).

### DIFF
--- a/coursier.bzl
+++ b/coursier.bzl
@@ -11,7 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 
-load("//third_party/bazel_json/lib:json_parser.bzl", "json_parse")
+load("//private/rules:json.bzl", "json_decode")
 load("//private/rules:jetifier.bzl", "jetify_artifact_dependencies", "jetify_maven_coord")
 load("//:specs.bzl", "maven", "parse", "utils")
 load("//:private/proxy.bzl", "get_java_proxy_args")
@@ -206,7 +206,7 @@ def _compute_dependency_tree_signature(artifacts):
 def compute_dependency_inputs_signature(artifacts):
     artifact_inputs = []
     for artifact in artifacts:
-        parsed = json_parse(artifact)
+        parsed = json_decode(artifact)
 
         # Sort the keys to provide a stable order
         keys = sorted(parsed.keys())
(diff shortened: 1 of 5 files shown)


coursier.bzl
```diff
@@ -346,37 +345,29 @@ def _pinned_coursier_fetch_impl(repository_ctx):
 
     _windows_check(repository_ctx)
 
+    print(native.bazel_version)
```
JaredNeil, line 348: Oops

shs96c:
I'll wait for @jin to weigh in on which versions of bazel we need to support before reviewing.

jin:
This is excellent!

Our current supported Bazel versions goes back to 1.x, and maybe even before. We don't have data, but it would be great to maintain that. Do you think we can conditionally use the JSON module using an attribute on maven_install? I don't know if functions are lazily loaded from bzl files, and if so, we could abstract the json decode logic in different bzl files and load them based on a new attribute.

The attribute could also be automatically set based on the bazel version, but that's another problem that we can handle later.

shs96c:
@JaredNeil you can access the current bazel version via `native.bazel_version` in macros run from a `WORKSPACE` file. If you use this, then it's possible to create a `.bzl` file that exports either the current `json_parse` or the new native JSON decode method and give them both the same API.

JaredNeil:
>you can access the current bazel version via native.bazel_version in macros run from a WORKSPACE file.

Yeah, that's how [`rules_nodejs`](https://github.com/bazelbuild/rules_nodejs/blob/stable/internal/common/check_bazel_version.bzl) (which was copied from [tensorflow](https://github.com/tensorflow/tensorflow/blob/5541ef4fbba56cf8930198373162dd3119e6ee70/tensorflow/workspace.bzl#L44)) does it. I just wasn't sure if that's the easiest/best way to do it. Should we copy that as well? ~It seems like something that should probably be in Skylib~ (it [totally is](https://github.com/bazelbuild/bazel-skylib/blob/main/docs/versions_doc.md), I'll use that).

>exports either the current json_parse or the new native JSON decode method and give them both the same API.

Should the exported API just not support `fail_on_invalid` at all?
I guess if they pass `fail_on_invalid = False` we could always use the Starlark parser, and only use the native one if it's available and `fail_on_invalid = True`. But then you don't get the speedups.

JaredNeil:
>Do you think we can conditionally use the JSON module using an attribute on maven_install? I don't know if functions are lazily loaded from bzl files, and if so, we could abstract the json decode logic in different bzl files and load them based on a new attribute.

I don't think so. How could that handle multiple `maven_install`s with different options?
We could pass a dynamically chosen parse function around, but it would have to be passed in through all of the helper functions, which is kinda messy.

You can export a different function based on the bazel version, so that might actually be the easier way to go. The downside is we have to do one of these three things:

1. Implement our own version checking
2. Copy in the version checking code from somewhere else, like `rules_nodejs` or `skylib`
3. Add a user workspace dependency on skylib (`rules_jvm_external` currently doesn't have any of these)

I'm partial to number 3, since the `http_archive(...); load(..., "X_dependencies"); X_dependencies()` pattern is [pretty](https://github.com/atlassian/bazel-tools/blob/master/multirun/README.md#setup-and-usage-via-bazel
) [well](https://github.com/bazelbuild/bazel-skylib#workspace-file
) [established](https://github.com/bazelbuild/rules_scala#getting-started
) for [other](https://bazelbuild.github.io/rules_rust/#setup
) [rulesets](https://github.com/bazelbuild/rules_docker#setup), and I think most workspaces probably already load skylib. But it would be new for this one, which would be a more significant change for end users.

Since we already have a pattern of vendoring code in `third_party`, going with option 2, and pulling in the [one file from `skylib`](https://github.com/bazelbuild/bazel-skylib/blob/main/lib/versions.bzl) would probably be the easiest option.

@jin, do you have a preference?

PR #569, opened 2021-06-11
Remove build_tools_version and api_level from example workspace files

ahumesky:
Remove `build_tools_version` and `api_level` from workspace file in examples to let `android_sdk_repository` automatically select the latest installed version. See https://github.com/bazelbuild/bazel/issues/13409


shs96c:
Closing in favour of #571 which appears to be identical and reviewed.

PR #579, opened 2021-07-18
Print a warning when there is a duplicate artifact in the artifact list

cheister:
Print a warning when an artifact appears in the list more than once. For example for the following maven_install
```
maven_install(
    name = "maven",
    artifacts = [
        "com.fasterxml.jackson.core:jackson-annotations:2.10.1",
        "com.fasterxml.jackson.core:jackson-annotations:2.12.1",
        "com.fasterxml.jackson.core:jackson-annotations:2.10.1",
        "com.fasterxml.jackson.core:jackson-annotations:2.11.2",
        "com.github.jnr:jffi:1.3.3",
        maven.artifact(
            group = "com.github.jnr",
            artifact = "jffi",
            version = "1.3.3",
            classifier = "native",
        ),
        maven.artifact(
            group = "com.github.jnr",
            artifact = "jffi",
            version = "1.3.2",
            classifier = "native",
        ),
    ],
    repositories = [
        "https://repo.maven.apache.org/maven2/",
    ],
)
```

print the following

```
>  bazel run @maven//:pin         
DEBUG: /private/var/tmp/_bazel_cheister/60fd9b2f4bf06a7f576b88fee484f3ad/external/rules_jvm_external/coursier.bzl:598:18: com.fasterxml.jackson.core:jackson-annotations is in the artifact list more than once. Please remove duplicate artifacts from the artifact list so you do not get unexpected artifact versions
DEBUG: /private/var/tmp/_bazel_cheister/60fd9b2f4bf06a7f576b88fee484f3ad/external/rules_jvm_external/coursier.bzl:598:18: com.fasterxml.jackson.core:jackson-annotations is in the artifact list more than once. Please remove duplicate artifacts from the artifact list so you do not get unexpected artifact versions
DEBUG: /private/var/tmp/_bazel_cheister/60fd9b2f4bf06a7f576b88fee484f3ad/external/rules_jvm_external/coursier.bzl:598:18: com.fasterxml.jackson.core:jackson-annotations is in the artifact list more than once. Please remove duplicate artifacts from the artifact list so you do not get unexpected artifact versions
DEBUG: /private/var/tmp/_bazel_cheister/60fd9b2f4bf06a7f576b88fee484f3ad/external/rules_jvm_external/coursier.bzl:598:18: com.github.jnr:jffi:native is in the artifact list more than once. Please remove duplicate artifacts from the artifact list so you do not get unexpected artifact versions
INFO: Analyzed target @maven//:pin (18 packages loaded, 67 targets configured).
INFO: Found 1 target...
Target @maven//:pin up-to-date:
  bazel-bin/external/maven/pin
INFO: Elapsed time: 6.296s, Critical Path: 0.00s
INFO: 1 process: 1 internal.
INFO: Build completed successfully, 1 total action
INFO: Build completed successfully, 1 total action
```

### DIFF
--- a/coursier.bzl
+++ b/coursier.bzl
@@ -372,6 +372,8 @@ def _pinned_coursier_fetch_impl(repository_ctx):
     for artifact in repository_ctx.attr.artifacts:
         artifacts.append(json_parse(artifact))
 
+    _check_artifacts_are_unique(artifacts, repository_ctx.attr.duplicate_version_warning)
+
     # Read Coursier state from maven_install.json.
     repository_ctx.symlink(
         repository_ctx.path(repository_ctx.attr.maven_install_json),
@@ -606,6 +608,31 @@ def _deduplicate_artifacts(dep_tree):
     dep_tree.update({"dependencies": deduped_artifacts.values() + null_artifacts})
     return dep_tree
 
+def _check_artifacts_are_unique(artifacts, duplicate_version_warning):
+    if duplicate_version_warning == "none":
+        return
+    seen_artifacts = {}
+    duplicate_artifacts = {}
+    for artifact in artifacts:
+        artifact_coordinate = artifact["group"] + ":" + artifact["artifact"] + (":%s" % artifact["classifier"] if artifact.get("classifier") != None else "")
+        if artifact_coordinate in seen_artifacts:
+            if artifact_coordinate in duplicate_artifacts:
+                duplicate_artifacts[artifact_coordinate].append(artifact["version"])
+            else:
+                duplicate_artifacts[artifact_coordinate] = [artifact["version"]]
+        else:
+            seen_artifacts[artifact_coordinate] = artifact["version"]
+
+    if duplicate_artifacts:
+        msg_parts = ["Found duplicate artifact versions"]
+        for duplicate in duplicate_artifacts:
+            msg_parts.append("    {} has multiple versions {}".format(duplicate, ", ".join([seen_artifacts[duplicate]] + duplicate_artifacts[duplicate])))
+        msg_parts.append("Please remove duplicate artifacts from the artifact list so you do not get unexpected artifact versions")
+        if duplicate_version_warning == "error":
+            fail("\n".join(msg_parts))
+        else:
+            print("\n".join(msg_parts))
+
 # Get the path to the cache directory containing Coursier-downloaded artifacts.
 #
 # This method is public for testing.
@@ -778,6 +805,8 @@ def _coursier_fetch_impl(repository_ctx):
     artifacts = []
     for artifact in repository_ctx.attr.artifacts:
         artifacts.append(json_parse(artifact))
+
+    _check_artifacts_are_unique(artifacts, repository_ctx.attr.duplicate_version_warning)
 
     excluded_artifacts = []
     for artifact in repository_ctx.attr.excluded_artifacts:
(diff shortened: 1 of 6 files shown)


jin (approved):

coursier.bzl
```diff
@@ -588,6 +590,16 @@ def _deduplicate_artifacts(dep_tree):
     dep_tree.update({"dependencies": deduped_artifacts.values() + null_artifacts})
     return dep_tree
 
+def _check_artifacts_are_unique(artifacts):
```
jin, line 593: nit improvements, feel free to change in a follow-up.

* Only print warning messages for each set of duplicated artifacts once. I don't think there's a need for 3 identical messages for a set of 3 duplicated artifacts.
* Show the full set of duplicated artifacts including versions. e.g. if com.foo.bar has v1, v2 and v3 in the list, print info about v1, v2 and v3 as well.

cheister:
The updated warning message looks like 
```
DEBUG: /Users/cheister/Development/rules_jvm_external/coursier.bzl:634:18: Found duplicate artifact versions
    com.fasterxml.jackson.core:jackson-annotations has multiple versions 2.10.1, 2.12.1, 2.10.1, 2.11.2
    com.github.jnr:jffi:native has multiple versions 1.3.3, 1.3.2
Please remove duplicate artifacts from the artifact list so you do not get unexpected artifact versions
```
I also added a flag so you can disable the message or fail the build when there are duplicates.

For testing I made the `tests/bazel_run_tests.sh` file which just runs shell commands and checks their output.  It's pretty basic but makes testing bazel run output much easier in CI.  If it looks good we should probably move all of the bazel run tests in the presubmit.yml file into `tests/bazel_run_tests.sh`

jin:
Nice!

Have you tried making`tests/bazel_run_tests.sh` a `sh_test`? If it's too complicated (e.g. calling bazel from bazel and setting up mock environments), let's do it in a follow-up, but generally we should run these checks as bazel tests.

jin (approved):

cheister:
> Have you tried making`tests/bazel_run_tests.sh` a `sh_test`? If it's too complicated (e.g. calling bazel from bazel and setting up mock environments), let's do it in a follow-up, but generally we should run these checks as bazel tests.

Yeah, I looked at https://github.com/bazelbuild/bazel/tree/master/src/test/shell and it's quite a bit of work to have Bazel test `bazel run`.   I figured since we already had all of the bazel run commands in `.bazelci/presubmit.yml`  to "integration" test the run rules, this was just a small extension so we can verify the output from those commands.  The downside is they don't get picked up by a `bazel test` target, but that was already true because they were only run `.bazelci/presubmit.yml`

lukaciko:
@cheister the copy is a bit confusing when the imported versions are the same:
> rules_jvm_external/coursier.bzl:660:18: Found duplicate artifact versions
    com.google.dagger:dagger has multiple versions 2.36, 2.36, 2.36

Not sure I see an issue with duplicate imports of a dependency with the same version. Actually if we follow the official [Dagger Bazel instructions](https://github.com/google/dagger#bazel) for vanilla Dagger & Hilt we end up importing [some artifacts](https://github.com/google/dagger/blob/b5990a0641a7860b760aa9055b90a99d06186af6/workspace_defs.bzl#L22-L44) twice.

cheister:
> Dagger Bazel instructions

I guess the default warning could be relaxed to only warn when the versions are different.  It still strikes me as an odd case that you have the same dependency in the list multiple times, but it's really only a problem if the versions are different.

PR #584, opened 2021-08-12
Remove android_instrumentation_test on RBE

comius:
Android Emulator on RBE basically works accidentally on their platform and it could stop working any time.

I'm removing the test because it is failing (see https://buildkite.com/bazel/rules-jvm-external-examples/builds/1960#4cef1180-9796-4dff-ae98-222391d9bc5a)


jin (approved):

jin:
Thanks!

PR #589, opened 2021-08-15
Use argsfile for java if we are using java 9 or higher to avoid command line length limits

cheister:
Should fix https://github.com/bazelbuild/rules_jvm_external/issues/549 if they are using Java 9 or higher

### DIFF
--- a/coursier.bzl
+++ b/coursier.bzl
@@ -17,6 +17,7 @@ load("//:specs.bzl", "maven", "parse", "utils")
 load("//:private/proxy.bzl", "get_java_proxy_args")
 load("//:private/dependency_tree_parser.bzl", "JETIFY_INCLUDE_LIST_JETIFY_ALL", "parser")
 load("//:private/coursier_utilities.bzl", "SUPPORTED_PACKAGING_TYPES", "escape")
+load("//:private/java_utilities.bzl", "parse_java_version")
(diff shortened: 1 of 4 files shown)


coursier.bzl
```diff
@@ -741,6 +746,22 @@ def make_coursier_dep_tree(
         # https://github.com/coursier/coursier/blob/1cbbf39b88ee88944a8d892789680cdb15be4714/modules/paths/src/main/java/coursier/paths/CoursierPaths.java#L29-L56
         environment = {"COURSIER_CACHE": str(repository_ctx.path(coursier_cache_location))}
 
+    # If we are using Java 9 or higher we can use an argsfile to avoid command line length limits
+    java_path = _java_path(repository_ctx)
+    if java_path:
+        exec_result = _execute(repository_ctx, [java_path, "-version"])
+        if (exec_result.return_code != 0):
+            fail("Error while running java -version: " + exec_result.stderr)
+        if parse_java_version(exec_result.stdout + exec_result.stderr) > 8:
```
jin, line 755: Is the output of java -version on both stdout and stderr?

cheister: It looks like it goes to stderr in the two cases I tried but I wasn't sure if that would be the case for all jvm versions.

PR #591, opened 2021-08-23
support for javac options

suman-ganta:
Added support for passing javac options while generating javadocs.

javac options need to be passed to javadoc compiler. This ensures that all the options passed while compiling sources are also passed to javadoc.

Ex: `--add-exports java.base/jdk.internal.misc=ALL-UNNAMED` is needed to be passed if source uses `jdk.internal.misc` package with the jdks that support modules.

### DIFF
--- a/private/tools/java/rules/jvm/external/javadoc/JavadocJarMaker.java
+++ b/private/tools/java/rules/jvm/external/javadoc/JavadocJarMaker.java
@@ -59,10 +59,11 @@ public class JavadocJarMaker {
     Set<Path> sourceJars = new HashSet<>();
     Path out = null;
     Set<Path> classpath = new HashSet<>();
+    List<String> options = new ArrayList<>();
 
     for (int i = 0; i < args.length; i++) {
       String flag = args[i];
-      String next = args[++i];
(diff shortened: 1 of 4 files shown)


private/tools/java/rules/jvm/external/javadoc/JavadocJarMaker.java
```diff
@@ -59,10 +59,11 @@ public static void main(String[] args) throws IOException {
     Set<Path> sourceJars = new HashSet<>();
     Path out = null;
     Set<Path> classpath = new HashSet<>();
+    List<String> options = new ArrayList<>();
 
     for (int i = 0; i < args.length; i++) {
       String flag = args[i];
-      String next = args[++i];
+      String next = i + 1 < args.length ? args[++i] : null;
```
shs96c, line 66: I'm guessing this is to handle the case where we may have single flags? If that's true, we should avoid reading `next` until we know that `flag` implies that there's a second parameter. That means that we should only be reading `next` if `flag` is one of `--cp`, `--in`, or `--out`

shs96c (changes requested):
The change makes sense. We need to make the argument processing a little more robust, though.

PR #593, opened 2021-08-25
Update Android build_tools_version and api_level in examples

Bencodes:
PRs are currently red as a result of this.

### DIFF
--- a/examples/android_instrumentation_test/WORKSPACE
+++ b/examples/android_instrumentation_test/WORKSPACE
@@ -3,7 +3,6 @@ load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 android_sdk_repository(
     name = "androidsdk",
     api_level = 28,
-    build_tools_version = "30.0.3",
 )
 
 android_ndk_repository(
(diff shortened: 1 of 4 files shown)


jin:
Thanks for the PR! Could you remove the build_tools_version line altogether and rely on the autodiscovery mechanism in the rule?

examples/android_instrumentation_test/WORKSPACE
```diff
@@ -3,7 +3,7 @@ load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 android_sdk_repository(
     name = "androidsdk",
     api_level = 28,
-    build_tools_version = "28.0.2",
+    build_tools_version = "30.0.3",
```
jin, line 6: We can remove this version pinning altogether.

jin (changes requested):

examples/android_instrumentation_test/WORKSPACE
Bencodes: Updated

jin (approved):
Thanks, and apologies for a long delay

PR #597, opened 2021-09-15
Fix jq not formatting large (> 4 MB) in linux

HackToHell:
Redid pin.sh as this fails on linux when the json file for maven dependencies becomes very large. Switches to using a intermediary file.

### DIFF
--- a/coursier.bzl
+++ b/coursier.bzl
@@ -43,7 +43,6 @@ load("%s", "aar_import")
 """
 
 _BUILD_PIN = """
-exports_files(['unsorted_deps.json'])
 genrule(
     name = "jq-binary",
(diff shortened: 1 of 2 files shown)


coursier.bzl
```diff
@@ -43,6 +43,7 @@ load("%s", "aar_import")
 """
 
 _BUILD_PIN = """
+exports_files(['unsorted_deps.json'])
```
jin, line 46: Is this line required, since it's not referenced outside of the package?

HackToHell: Good catch, doesn't seem to be required. Have updated the PR.

jin (approved):

jin:
thanks!

PR #599, opened 2021-09-22
Remove build_tools_version pin from example repos.

jin:
The CI removed v28, and may change again, so let's just rely on the
autodiscovery mechanism in android_sdk_repository to use the right
build_tools_version.


jin:
Closing this in favour of https://github.com/bazelbuild/rules_jvm_external/pull/593, which came earlier

PR #609, opened 2021-10-18
Add artifact URL to jvm_import tags

danielmachlab:
This change adds an artifact's download URL into the `tags` of `jvm_import` in the external repository BUILD file, generated by `maven_install`. Previously, the `tags` only included `maven_coordinate`, but they will now also include `maven_url`, storing the artifact's URL. This information is intended to be consumed by an aspect that is traversing a build graph for dependencies with the purpose of OSS identification and license compliance.

As an example of the generated `tags`: when importing `com.google.code.findbugs:jsr305:3.0.2` with `maven_install`, the generated BUILD file containing `jvm_import` will contain the following tags (where the `maven_url` is dependent on the values provided to `maven_install`'s `repository` attribute): 
```
tags = [
        "maven_coordinates=com.google.code.findbugs:jsr305:3.0.2",
        "maven_url=https://jcenter.bintray.com/com/google/code/findbugs/jsr305/3.0.2/jsr305-3.0.2.jar",
],
```

This change will be used immediately by the VMware/rules_oss_audit project—soon to be published on github.

**Testing Done:**
This change adds an analysis test to verify the existence of the `maven_url` and `maven_coordinate` tags in `jvm_import`. The analysis test consumes an aspect which propagates the tags with the `TagsInfo` provider.

The test suite, containing the single analysis test, can be run with:
```
% bazel test tests/unit/jvm_import:jvm_import_tests
```

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -415,6 +415,16 @@ maven_install(
 )
 
 maven_install(
+    name = "jvm_import_test",
+    artifacts = [
+        "com.google.code.findbugs:jsr305:3.0.2",
+    ],
+    repositories = [
+        "https://jcenter.bintray.com/",
+    ],
+)
+
+maven_install(
     name = "starlark_aar_import_with_sources_test",
     artifacts = [
         "androidx.work:work-runtime:2.6.0",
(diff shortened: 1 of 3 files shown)


jin:
Thanks! Could you please accompany this PR with an analysis test that verifies the existence of this URL and the artifact coordinates? 

https://docs.bazel.build/versions/main/skylark/testing.html should have the necessary APIs for a test like that.

danielmachlab:
Sounds good! Thanks for letting me know about the test failures and taking a look at this PR 👍

PR #624, opened 2021-11-22
Use bazel's native json decoder where possible

shs96c:
And fall back to the slower, Starlark-based one when we either need to
survive invalid json files, or when we're using a version of bazel
that doesn't have the built-in decoder.

### DIFF
--- a/coursier.bzl
+++ b/coursier.bzl
@@ -11,7 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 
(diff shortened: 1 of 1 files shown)


shs96c:
This is a simplification of #566 and should get us the same performance boost.

jin (approved):
Nice :)

coursier.bzl
```diff
@@ -82,6 +82,11 @@ sh_binary(
 )
 """
 
+def json_parse(json_string, fail_on_invalid = True):
+    if getattr(native, "json_decode", None) and not fail_on_invalid:
```
jin, line 86: Could you comment with the minimum bazel version json_decode is available in?

shs96c: Done.

Issue #630, opened 2021-12-11
examples/spring_boot references vulnerable log4j version


philwo:
Hi,

I'm going through all of bazelbuild's code and found that [examples/spring_boot/maven_install.json](https://github.com/bazelbuild/rules_jvm_external/blob/fa73b1a8e4846cee88240d0019b8f80d39feb1c3/examples/spring_boot/maven_install.json#L319) references a vulnerable version of log4j.

I know that it's just example code, but I think it would still be good to update this to the latest version. Unfortunately I don't know how to regenerate the JSON file with newer versions. Could you please look at this?

Thanks,
Philipp

philwo:
We should also update [regression_testing_install.json](https://github.com/bazelbuild/rules_jvm_external/blob/73a1bc73b1758732de9f7a1ed3b1b349405f7f18/tests/custom_maven_install/regression_testing_install.json#L2934).

jin:
Sent the PR. For the record, the steps are to:

1) Update the maven_install definition in the WORKSPACE file with the new version, and using `version_conflict_policy = "pinned"` to tell Coursier to use the explicit versions specified.
2) Run `bazel run @unpinned_maven//:pin`. This is documented here: https://github.com/bazelbuild/rules_jvm_external#updating-maven_installjson

philwo:
Thanks!

PR #631, opened 2021-12-11
Update vulnerable 2.x log4j version to 2.15.0.

jin:
Fixes https://github.com/bazelbuild/rules_jvm_external/issues/630

The other log4j versions used are 1.x, which are not affected by this particular vulnerability affecting 2.x before 2.15.0.


examples/spring_boot/WORKSPACE
```diff
@@ -7,6 +7,8 @@ load("@rules_jvm_external//:defs.bzl", "maven_install")
 
 maven_install(
     artifacts = [
+        "org.apache.logging.log4j:log4j-api:2.15.0",
```
philwo, line 10: Should we add a comment why these versions are explicitly pinned? Otherwise I fear this will never be removed again, even when we would no longer need it.

philwo (approved):

cheister (approved):

examples/spring_boot/WORKSPACE
cheister: 👍 Maybe add `# log4j deps are added only for https://github.com/bazelbuild/rules_jvm_external/issues/630`

jin: Done

PR #644, opened 2022-01-04
Allow overriding rules_jvm_external_deps_install.json

dmivankov:
This can be useful for setting up alternative urls for internal dependencies of rules_jvm_external.
Some potential use-cases
- having slightly different version of those dependencies
- having an extra proxy hosting that could be faster / more local
- having an extra proxy hosting that is persisted / archived for extra reproducibility
- just having a uniform proxy for all / most external dependencies

### DIFF
--- a/repositories.bzl
+++ b/repositories.bzl
@@ -5,13 +5,13 @@ _DEFAULT_REPOSITORIES = [
     "https://maven.google.com",
 ]
 
-def rules_jvm_external_deps(repositories = _DEFAULT_REPOSITORIES):
+def rules_jvm_external_deps(repositories = _DEFAULT_REPOSITORIES, maven_install_json = "@rules_jvm_external//:rules_jvm_external_deps_install.json"):
     maven_install(
         name = "rules_jvm_external_deps",
         artifacts = [
             "com.google.cloud:google-cloud-storage:1.113.4",
         ],
-        maven_install_json = "@rules_jvm_external//:rules_jvm_external_deps_install.json",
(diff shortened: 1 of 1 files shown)


shs96c:
Maybe I'm missing something, because I'm not sure how this PR addresses the problems you're outlining in a reliable way.

If you modify the list of dependencies, the builds will fail because `fail_if_repin_required` is set to `True`. If we need to update the lock file, please let us know, and we can do that.

If you need to regenerate the lock file yourself, then just running `REPIN=1 bazel run @unpinned_rules_jvm_external_deps//:pin` will regenerate it. In the rare cases I've needed to do this, I use the `patches` and `patch_args` parameters of `http_archive` to apply my changes to `rules_jvm_external`

Because we're using a lock file, the downloads are all going through bazel's own downloader, so things like the proxy to use should be configured at that level.

Could you please expand a little more about how this change addresses these problems?

dmivankov:
Yes, the lock file may need to be updated on rules_jvm_external updates.
This PR is basically to have less patches, but if that's not very useful/generic it's ok to not merge it too :+1: 

For bazel downloader proxy that could be interesting to investigate, it would allow overriding/persisting urls on the proxy, but not change contents (`--override-repository=` could be used there though).
With lock patching approach `bazel query` can be used to list all external urls, and check if all of them are proxied (via an artifactory instance for example).

dmivankov:
Probably yes, too ad hoc and on the other hand so small that it's not hard to keep that patch out of mainstream

Issue #147, opened 2019-05-17
Consider an optional switch to turn all `deps` into `exports`, allowing access to all transitive classes just be depending on the top level artifact


jin:
This was motivated by a conversation with @simonstewart for better developer ergonomics and gradual adoption, due to the behavior replicating Maven and Gradle.

The general idea is to depend on just the top level dependency, and get all the classes in the transitive classpath for free. This was previously implemented by @davidsantiago in https://github.com/jin/rules_maven/pull/9, but we removed it in https://github.com/jin/rules_maven/pull/16 with the concerns stated in the first comment and @dkelmer's comment in https://github.com/jin/rules_maven/pull/16#pullrequestreview-210447108

We should consider the tradeoffs between ergonomics and the growth of classpath complexity. This can be an opt-in feature with major disclaimers on its implications.

shs96c:
As an opt-in feature, this is great. There are plenty of libraries that people depend on via Maven that are designed to bring in others transitively (for example, Selenium's `selenium-java` package, which gives access to a whole heap of magic) 

Ideally, this could be configured at the repository level (if I'm pulling from Maven Central, I want the default maven behaviour, but if I'm pulling from a local maven repo maybe something more disciplined makes sense) At the artifact level would end up being super-noisy in the common case.

uri-canva:
I think this is required for scala support: the way the types work means that all libraries need to be compiled with symbols from transitive dependencies available during compilation.

jin:
Should we have this as an opt-in feature per artifact? That is,

```python
maven_install(
    artifacts = [
        maven.artifact("group.id", "artifact", "1.2.3", export_transitive_compile_deps = True),
    ],
)
```

instead of applying this across the entire graph. So, for the following example,

https://github.com/bazelbuild/rules_jvm_external/blob/9e8ab911f102b436f7a692af7901c1e1dd80f8f0/regression_testing_install.json#L63-L84

we would place

https://github.com/bazelbuild/rules_jvm_external/blob/9e8ab911f102b436f7a692af7901c1e1dd80f8f0/regression_testing_install.json#L65-L68

into `exports` instead of `deps` of the generated `aar_import` target.

jin:
@shs96c what do you think? I think this is a more controllable approach than a `maven_install` wide implementation.

shs96c:
[...]

PR #645, opened 2022-01-04
Allow pinning transitive artifacts via new pinned_artifacts attribute

dmivankov:
Using artifacts= + version_conflict_policy=pinned is possible to set versions
of transitive artifacts but that affects visibility, required policy change and
makes artifacts= bigger than it needs to be.

### DIFF
--- a/coursier.bzl
+++ b/coursier.bzl
@@ -674,6 +674,7 @@ def make_coursier_dep_tree(
         repository_ctx,
         artifacts,
         excluded_artifacts,
+        pinned_artifacts,
         repositories,
         version_conflict_policy,
         fail_on_missing_checksum,
@@ -706,6 +707,8 @@ def make_coursier_dep_tree(
         for coord in artifact_coordinates:
             # Undo any `,classifier=` suffix from `utils.artifact_coordinate`.
             cmd.extend(["--force-version", coord.split(",classifier=")[0]])
+    for a in pinned_artifacts:
+        cmd.extend(["--force-version", utils.artifact_coordinate(a).split(",classifier=")[0]])
     cmd.extend(["--artifact-type", ",".join(SUPPORTED_PACKAGING_TYPES + ["src", "doc"])])
     cmd.append("--verbose" if _is_verbose(repository_ctx) else "--quiet")
     cmd.append("--no-default")
@@ -837,6 +840,10 @@ def _coursier_fetch_impl(repository_ctx):
     for artifact in repository_ctx.attr.excluded_artifacts:
         excluded_artifacts.append(json_parse(artifact))
 
+    pinned_artifacts = []
+    for artifact in repository_ctx.attr.pinned_artifacts:
+        pinned_artifacts.append(json_parse(artifact))
+
     # Once coursier finishes a fetch, it generates a tree of artifacts and their
     # transitive dependencies in a JSON file. We use that as the source of truth
     # to generate the repository's BUILD file.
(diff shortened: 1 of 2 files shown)


shs96c:
I thought coursier supported exact version pinning using the `[version]` [maven syntax](https://docs.oracle.com/middleware/1212/core/MAVEN/maven_version.htm#MAVEN402). If so, you can get the equivalent of `pinned_artifacts` using coordinates like `com.google.guava:guava:[31.0.1-jre]`.

Please let me know if I am mistaken, and I'll do a proper review of this PR.

dmivankov:
It's possible to use 
```python
maven_install(
    artifacts = ["com.google.guava:guava:31.0.1-jre"],
    version_conflict_policy = "pinned",
)
```
But the downside is
- conflict policy applies to all artifacts
- artifact items become visible even if not necessarily directly used by project

With new attribute one can do something like this
```python
maven_install(
    artifacts = ["com.google.guava:guava:31.0.1-jre", ...].
    version_conflict_policy = "pinned",  # or maybe use the default one
    pinned_artifacts = [
        # 4.1.67.Final has CVE https://github.com/advisories/GHSA-grg4-wf29-r9vv
        "io.netty:netty-codec-http2:4.1.68.Final",
        "io.netty:netty-codec-http:4.1.68.Final",
        "io.netty:netty-codec-socks:4.1.68.Final",
        "io.netty:netty-codec:4.1.68.Final",
        "io.netty:netty-handler-proxy:4.1.68.Final",
        "io.netty:netty-transport-native-epoll:jar:linux-x86_64:4.1.68.Final",
        "io.netty:netty-transport-native-kqueue:jar:osx-x86_64:4.1.68.Final",
        "io.netty:netty-transport-native-unix-common:4.1.68.Final",
    ],  
)
```
transitive dependency version is bumped, but directly used dependencies vs transitive dependency versions lists are still separated

jin:
I don't see how the new attribute is necessarily simpler than putting the new list of pinned_artifacts in artifacts, and using pinned version_conflict_policy?

dmivankov:
Mainly two reasons:
- esthetic separation of direct dependencies + not providing public visibility for indirect artifacts
- makes it simpler to write a check for directly unused dependencies (which don't need their version to be manually maintained/pinned generally, except for when they do and this is unblocked by this PR). Script along such lines
```sh
bazel query 'kind(jvm_import, @maven//... - deps(//..., 1)) intersect attr(visibility, public, @maven//...)'
```

jin:
> esthetic separation of direct dependencies + not providing public visibility for indirect artifacts

I don't think adding a new API for this makes it necessarily simpler. A new attribute is an attribute that we'd have to support ~forever. For aesthetics, it could be a project-side concatenation of lists.

re: public visibility for indirect artifacts is needed, given https://github.com/bazelbuild/rules_jvm_external/issues/147#issuecomment-493496912

> makes it simpler to write a check for directly unused dependencies (which don't need their version to be manually maintained/pinned generally, except for when they do and this is unblocked by this PR). Script along such lines

Why can't this be an rdeps query?

In general, I'm not too comfortable with adding another core attribute like this, especially if it doesn't bring significant benefit over the existing mechanisms.

dmivankov:
main thing that isn't possible without this PR is having `version_conflict_policy != "pinned"` and setting specific single dependency version. But maybe that's a rare case as often either `pinned` is used or `default` picks same or newer (iirc) version of all specified dependencies (downgrade pinning should be more rare)

will close then

PR #646, opened 2022-01-04
Allow setting :scope via maven coords too

dmivankov:
Default scope is "compile" but it could be useful to depend on other scopes like "tests"

### DIFF
--- a/private/rules/pom_file.bzl
+++ b/private/rules/pom_file.bzl
@@ -13,7 +13,7 @@ def _format_dep(unpacked):
 
 def _unpack_coordinates(coords):
     """Takes a maven coordinate and unpacks it into a struct with fields
-    `groupId`, `artifactId`, `version`, `type_`, `scope`
+    `groupId`, `artifactId`, `version`, `type`, `scope`
     where type and scope are optional.
 
     Assumes following maven coordinate syntax:
@@ -32,7 +32,7 @@ def _unpack_coordinates(coords):
     return struct(
         groupId = parts.get(0),
         artifactId = parts.get(1),
-        type_ = parts.get(2),
+        type = parts.get(2),
         scope = parts.get(3),
         version = version,
     )
@@ -49,7 +49,7 @@ def _pom_file_impl(ctx):
         "{groupId}": coordinates.groupId,
         "{artifactId}": coordinates.artifactId,
         "{version}": coordinates.version,
-        "{type}": coordinates.type_,
+        "{type}": coordinates.type,
         "{scope}": coordinates.scope,
     }
 
--- a/private/rules/pom_file.bzl
+++ b/private/rules/pom_file.bzl
@@ -13,7 +13,7 @@ def _format_dep(unpacked):
 
 def _unpack_coordinates(coords):
     """Takes a maven coordinate and unpacks it into a struct with fields
-    groupId, artifactId, version, type_, scope
+    `groupId`, `artifactId`, `version`, `type_`, `scope`
     where type and scope are optional.
 
     Assumes following maven coordinate syntax:
--- a/private/rules/pom_file.bzl
+++ b/private/rules/pom_file.bzl
@@ -2,13 +2,13 @@ load(":has_maven_deps.bzl", "MavenInfo", "has_maven_deps")
 
 def _format_dep(unpacked):
     return "".join([
- "        <dependency>\n",
- "            <groupId>%s</groupId>\n" % unpacked.groupId,
- "            <artifactId>%s</artifactId>\n" % unpacked.artifactId,
- "            <version>%s</version>\n" % unpacked.version,
-("            <type>%s</type>\n" % unpacked.type) if unpacked.type and unpacked.type != "jar" else "",
-("            <scope>%s</scope>\n" % unpacked.scope) if unpacked.scope and unpacked.scope != "compile" else "",
- "        </dependency>",
+        "        <dependency>\n",
+        "            <groupId>%s</groupId>\n" % unpacked.groupId,
+        "            <artifactId>%s</artifactId>\n" % unpacked.artifactId,
+        "            <version>%s</version>\n" % unpacked.version,
+        ("            <type>%s</type>\n" % unpacked.type) if unpacked.type and unpacked.type != "jar" else "",
+        ("            <scope>%s</scope>\n" % unpacked.scope) if unpacked.scope and unpacked.scope != "compile" else "",
+        "        </dependency>",
     ])
 
 def _unpack_coordinates(coords):
--- a/private/rules/pom_file.bzl
+++ b/private/rules/pom_file.bzl
@@ -1,44 +1,41 @@
 load(":has_maven_deps.bzl", "MavenInfo", "has_maven_deps")
 
-_PLAIN_DEP = """        <dependency>
-            <groupId>{0}</groupId>
-            <artifactId>{1}</artifactId>
-            <version>{2}</version>
-        </dependency>"""
+def _format_dep(unpacked):
+    return "".join([
+ "        <dependency>\n",
+ "            <groupId>%s</groupId>\n" % unpacked.groupId,
+ "            <artifactId>%s</artifactId>\n" % unpacked.artifactId,
+ "            <version>%s</version>\n" % unpacked.version,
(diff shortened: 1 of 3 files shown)


private/rules/pom_file.bzl
```diff
@@ -13,19 +13,30 @@ _TYPED_DEP = """        <dependency>
...
     if len(parts) == 3:
-        return (parts[0], parts[1], parts[2], "jar")
+        return (parts[0], parts[1], parts[2], "jar", "compile")
     if len(parts) == 4:
         # Assume a buildr coordinate: groupId:artifactId:type:version
-        return (parts[0], parts[1], parts[3], parts[2])
+        return (parts[0], parts[1], parts[3], parts[2], "compile")
+    if len(parts) == 5:
+        # Assume a buildr coordinate: groupId:artifactId:type:scope:version
+        return (parts[0], parts[1], parts[4], parts[2], parts[3])
```
shs96c, line 39: The comment says we're returning `group:artifact:type:scope:version`, but the actual tuple returned is `group:artifact:scope:version:type`. Either the comment or the code isn't quite right, and in this case, I think it's the code.

shs96c (changes requested):
Looks great once the comment is addressed. Thank you for the PR!

private/rules/pom_file.bzl
dmivankov: Yes, this ordering is fragile, `explode_coordinates`, `_*_DEP` format strings and `_pom_file_impl` use of exploded coordinates all needs to be in sync, maybe it's time to simplify that by introducing a struct instead of tuple or something like that.

As for the comment, it's kind of correct, refers to input string format ([buildr](https://buildr.apache.org/artifacts.html) flavor), existing branches also change order by putting version into 3rd item of the tuple, while in coordinate string it's always the last part, if present.

dmivankov: tried to rewrite this part now

private/rules/pom_file.bzl
```diff
@@ -1,33 +1,41 @@
...
-            <artifactId>{1}</artifactId>
-            <version>{2}</version>
-        </dependency>"""
+def _format_dep(unpacked):
+    return "".join([
+ "        <dependency>\n",
+ "            <groupId>%s</groupId>\n" % unpacked.groupId,
+ "            <artifactId>%s</artifactId>\n" % unpacked.artifactId,
+ "            <version>%s</version>\n" % unpacked.version,
+("            <type>%s</type>\n" % unpacked.type) if unpacked.type and unpacked.type != "jar" else "",
```
shs96c, line 9: Does this work? I would expect: `"<type>%s</type>\n" % unpacked.type if unpacked.type and unpacked.type != "jar" else ""` That is, the closing bracket seems like it's in the wrong place (and it's confusing even if it's the right place :) )

private/rules/pom_file.bzl
```diff
@@ -1,33 +1,41 @@
 load(":has_maven_deps.bzl", "MavenInfo", "has_maven_deps")
 
-_PLAIN_DEP = """        <dependency>
-            <groupId>{0}</groupId>
-            <artifactId>{1}</artifactId>
-            <version>{2}</version>
-        </dependency>"""
+def _format_dep(unpacked):
+    return "".join([
+ "        <dependency>\n",
```
shs96c, line 5: Indentation goes wrong here. Have you run `buildifier -r -lint fix  .` on this PR? If not, could you please do so?

private/rules/pom_file.bzl
```diff
@@ -1,33 +1,41 @@
 load(":has_maven_deps.bzl", "MavenInfo", "has_maven_deps")
 
-_PLAIN_DEP = """        <dependency>
-            <groupId>{0}</groupId>
-            <artifactId>{1}</artifactId>
-            <version>{2}</version>
-        </dependency>"""
+def _format_dep(unpacked):
```
shs96c, line 3: I know why this always creates the longer version of the dependency declaration (because it's consistent and that's always nice), but it reads terribly in the generated POM files, which is why we had the two alternatives before.

Perhaps this method could conditionally add the `type` and `scope` fields if they need to be added?

We can address this in a follow up PR, but it'd be nice to do it now if you can.

private/rules/pom_file.bzl
```diff
@@ -1,33 +1,41 @@
...
 
-_TYPED_DEP = """        <dependency>
-            <groupId>{0}</groupId>
-            <artifactId>{1}</artifactId>
-            <version>{2}</version>
-            <type>{3}</type>
-        </dependency>"""
+def _unpack_coordinates(coords):
+    """Takes a maven coordinate and unpacks it into a struct with fields
+    groupId, artifactId, version, type_, scope
```
shs96c, line 16: `type_`? A quick exploration suggests you'll be fine with `type` on it's own:

```starlark
foo = struct(type = "bar")
print("Foo is: %s and dir is %s" % (foo, dir(foo)))
```

prints `Foo is: struct(type = "bar") and dir is ["to_json", "to_proto", "type"]`

shs96c: You may want to mark those fields as code (eg. `groupId` rather than just groupId) so it renders more legibly in the generated docs.

private/rules/pom_file.bzl
```diff
@@ -1,33 +1,41 @@
...
-    if len(parts) == 4:
-        # Assume a buildr coordinate: groupId:artifactId:type:version
-        return (parts[0], parts[1], parts[3], parts[2])
+    nparts = len(parts)
+    if nparts < 3 or nparts > 5:
+        fail("Unparsed: %s" % coords)
 
-    fail("Unparsed: %s" % coords)
+    version = parts[-1]
+    parts = dict(enumerate(parts[:-1]))
```
shs96c, line 31: Clever

private/rules/pom_file.bzl
shs96c: Also, the struct has a `type_` field, and you're using `type` here, so the code won't work. I'm not sure how this managed to sneak past the CI, but that's my problem to solve :)

shs96c (changes requested):
The only major thing is that attempting to build a pom file will now fail because of the `type`/`type_` problem.

private/rules/pom_file.bzl
dmivankov: used `type_` just because vim syntax highlighted it as a keyword, but maybe it works, let's try

private/rules/pom_file.bzl
dmivankov: that's actually what it tries to do, maybe in a little sneaky way by having
```
       ("            <type>%s</type>\n" % unpacked.type) if unpacked.type and unpacked.type != "jar" else "",
       ("            <scope>%s</scope>\n" % unpacked.scope) if unpacked.scope and unpacked.scope != "compile" else "",
```
so lines should be absent when they aren't needed, but there could be a cleaner way to achieve that

private/rules/pom_file.bzl
dmivankov: running on whole repo creates big diff, so applying it only to this file for now

dmivankov:
trying to fix test failures with https://github.com/bazelbuild/rules_jvm_external/pull/657

shs96c (commented):
Looks good, but this fails:

```shell
cd examples/java-export
bazel build //:example-export-pom
```

The output is:

```shell
ERROR: /src/rules_jvm_external/examples/java-export/BUILD:10:12: in pom_file rule //:example-export-pom: 
	File "/tmp/_bazel_shs/76cdb8bc63b14507f5041d8c30cccb52/external/rules_jvm_external/private/rules/pom_file.bzl", line 63, column 32, in _pom_file_impl
		ctx.actions.expand_template(
Error in expand_template: got dict<string, NoneType> for 'substitutions', want dict<string, string>
ERROR: Analysis of target '//:example-export-pom' failed; build aborted: Analysis of target '//:example-export-pom' failed
```

Apparently, our test suite coverage isn't great for building poms, and I apologise for that. I think that if `bazel build //...` passes in `examples/java-export` we're good to go.

dmivankov:
Thank you for testing, will try to make an update

shs96c (approved):

PR #647, opened 2022-01-04
Propagate bazel tags= in java_export sub-targets

dmivankov:
One use-case is using tags=["manual"] to exclude building of these targets by default.

### DIFF
--- a/private/rules/java_export.bzl
+++ b/private/rules/java_export.bzl
@@ -61,7 +61,7 @@ def java_export(
         and so may contain any valid parameter for that rule.
     """
 
-    tags = tags + ["maven_coordinates=%s" % maven_coordinates]
+    maven_coordinates_tags = ["maven_coordinates=%s" % maven_coordinates]
     lib_name = "%s-lib" % name
 
     javadocopts = kwargs.pop("javadocopts", [])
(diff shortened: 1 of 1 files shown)


private/rules/java_export.bzl
```diff
@@ -61,15 +61,14 @@ def java_export(
         and so may contain any valid parameter for that rule.
     """
 
-    tags = tags + ["maven_coordinates=%s" % maven_coordinates]
```
shs96c, line 64: Previously, both the `java_library` and the `maven_project_jar` had the maven coordinates added in their tag. With this change, only the former includes it. Please add the coordinates tag back to `maven_project_jar` too

shs96c (changes requested):
Looks great, but there's an unexpected change in the existing functionality. Once that's fixed, I'll be happy to merge this into the tree. Thank you for your PR!

private/rules/java_export.bzl
dmivankov: Will do

dmivankov: Should be fixed now

jin (approved):

shs96c (approved):

shs96c:
Thank you for the PR! Merging now :)

PR #648, opened 2022-01-04
Expose list of all artifacts, including transitive

dmivankov:
And add support for :type and :classifier in defs.bzl:artifact() by internally
using strip_packaging_and_classifier_and_version. To keep support for groupId:artifactId
update strip_packaging_and_classifier_and_version implementation for that case.

To be used like
```python
    load("@maven//:defs.bzl", "maven_artifacts")
```

and for example
```python
    load("@rules_jvm_external//:defs.bzl", "artifact")
    load("@rules_jvm_external//:specs.bzl", "parse")

    all_maven_targets = [artifact(a) for a in maven_artifacts if not parse.parse_maven_coordinate(a).get("classifier", "compile") in ["sources", "native"]]

    java_export(
        name = "maven-transitive-deps-lib",
        maven_coordinates = "com.example.deps:%s:%s" % (name, "0.1"),
        runtime_deps = all_maven_targets,
    )
```
which makes "maven-transitive-deps-lib-pom" target available that
can be useful for other tools like security scanners and/or dependabot

### DIFF
--- a/coursier.bzl
+++ b/coursier.bzl
@@ -471,11 +471,13 @@ def _pinned_coursier_fetch_impl(repository_ctx):
         "load(\"@bazel_tools//tools/build_defs/repo:utils.bzl\", \"maybe\")",
         "def pinned_maven_install():",
     ]
+    maven_artifacts = []
     netrc_entries = {}
 
     for artifact in dep_tree["dependencies"]:
         if artifact.get("url") != None:
             http_file_repository_name = escape(artifact["coord"])
+            maven_artifacts.extend([artifact["coord"]])
             http_files.extend([
                 "    http_file(",
                 "        name = \"%s\"," % http_file_repository_name,
@@ -498,6 +500,8 @@ def _pinned_coursier_fetch_impl(repository_ctx):
             http_files.append("    )")
 
     http_files.extend(_get_jq_http_files())
+
+    http_files.extend(["maven_artifacts = [\n%s\n]" % (",\n".join(["    \"%s\"" % artifact for artifact in maven_artifacts]))])
 
     repository_ctx.file("defs.bzl", "\n".join(http_files), executable = False)
     repository_ctx.file(
--- a/private/coursier_utilities.bzl
+++ b/private/coursier_utilities.bzl
@@ -46,6 +46,10 @@ def strip_packaging_and_classifier(coord):
     return ":".join(coordinates)
 
 def strip_packaging_and_classifier_and_version(coord):
+    coordinates = coord.split(":")
+    # Support for simplified versionless groupId:artifactId coordinate format
+    if len(coordinates) == 2:
+        return ":".join(coordinates)
     return ":".join(strip_packaging_and_classifier(coord).split(":")[:-1])
 
 # TODO: Should these methods be combined with _parse_maven_coordinate_string in specs.
--- a/private/rules/artifact.bzl
+++ b/private/rules/artifact.bzl
@@ -12,6 +12,14 @@ def maven_artifact(a):
 def _escape(string):
     return string.replace(".", "_").replace("-", "_").replace(":", "_")
 
+# inverse of parse_maven_coordinate
 def _make_artifact_str(artifact_obj):
-    # TODO: add support for optional type, classifier and version parts in artifact_obj case
-    return artifact_obj["group"] + ":" + artifact_obj["artifact"]
+    # produce either simplified g:a or standard g:a:[p:[c:]]v Maven coordinate string
+    coord = [artifact_obj["group"], artifact_obj["artifact"]]
+    if "version" in artifact_obj:
+        if "packaging" in artifact_obj:
+           coord.extend([artifact_obj["packaging"]])
+           if "classifier" in artifact_obj:
+             coord.extend([artifact_obj["classifier"]])
+        coord.extend([artifact_obj["version"]])
+    return ":".join(coord)
--- a/private/rules/artifact.bzl
+++ b/private/rules/artifact.bzl
@@ -1,9 +1,10 @@
 load("//:specs.bzl", "parse")
 load("//private:constants.bzl", "DEFAULT_REPOSITORY_NAME")
+load("//private:coursier_utilities.bzl", "strip_packaging_and_classifier_and_version")
 
 def artifact(a, repository_name = DEFAULT_REPOSITORY_NAME):
-    artifact_obj = _parse_artifact_str(a) if type(a) == "string" else a
-    return "@%s//:%s" % (repository_name, _escape(artifact_obj["group"] + ":" + artifact_obj["artifact"]))
+    artifact_str = _make_artifact_str(a) if type(a) != "string" else a
+    return "@%s//:%s" % (repository_name, _escape(strip_packaging_and_classifier_and_version(artifact_str)))
 
 def maven_artifact(a):
     return artifact(a, repository_name = DEFAULT_REPOSITORY_NAME)
@@ -11,9 +12,6 @@ def maven_artifact(a):
 def _escape(string):
     return string.replace(".", "_").replace("-", "_").replace(":", "_")
 
-def _parse_artifact_str(artifact_str):
-    pieces = artifact_str.split(":")
-    if len(pieces) == 2:
-        return {"group": pieces[0], "artifact": pieces[1]}
-    else:
-        return parse.parse_maven_coordinate(artifact_str)
+def _make_artifact_str(artifact_obj):
+    # TODO: add support for optional type, classifier and version parts in artifact_obj case
+    return artifact_obj["group"] + ":" + artifact_obj["artifact"]
--- a/README.md
+++ b/README.md
@@ -452,6 +452,9 @@ To use it, add the load statement to the top of your BUILD file:
 load("@rules_jvm_external//:defs.bzl", "artifact")
 ```
 
+Full `group:artifact:[packaging:[classifier:]]version` maven coordinates are also
+supported and translate to corresponding versionless target.
+
 Note that usage of this macro makes BUILD file refactoring with tools like
 `buildozer` more difficult, because the macro hides the actual target label at
 the syntax level.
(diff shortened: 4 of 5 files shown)


coursier.bzl
```diff
@@ -469,11 +469,14 @@ def _pinned_coursier_fetch_impl(repository_ctx):
         "load(\"@bazel_tools//tools/build_defs/repo:utils.bzl\", \"maybe\")",
         "def pinned_maven_install():",
     ]
+    maven_artifacts = []
     netrc_entries = {}
 
     for artifact in dep_tree["dependencies"]:
         if artifact.get("url") != None:
             http_file_repository_name = escape(artifact["coord"])
+            if not get_classifier(artifact["coord"]) == "sources":
```
shs96c, line 478: This check seems awfully fragile. I think it'll allow javadoc jars and platform binaries to be included, which probably isn't your intent. Perhaps an allow list of classifiers would be a better approach?

shs96c (changes requested):
I'm not sure what the utility of exposing every single maven dependency is. For code scanning, it seems like it would be more targetted to override the macros generating deployable artifacts to walk the dependencies (possibly using the `has_maven_deps` aspect?) Out of process tools, such as dependabot, can already do something like `bazel query @maven//:all` to discover all the maven artifacts.

I'm reviewing even though I don't understand the _why_ of this PR.

dmivankov:
For github dependabot script execution isn't allowed but it has support for checked in `pom.xml` files (and it doesn't see transitive dependencies there as different projects may potentially resolve them differently), this file can be maintained by having `java_export` for transitive dependencies and `diff_test` between checked in file and java_export pom

Second case where it can be used is writing tools that depend on both dependencies metadata and jars, with dependencies info exported into starlark one can write such tools in bazel. One example is being able to list all `(classname, maven dependency)` pairs.

coursier.bzl
dmivankov: Yes, maybe it's be better to expose more information (classifier + label?) and let consumers decide what to do with it, will look into that.

shs96c: I think you also want to reject javadoc jars as well since `rules_jvm_external` makes it easy to download those too.

shs96c (changes requested):
One thing which I think our users will bump into quickly, then this is good to go. Thank you for the PR!

dmivankov:
completely forgot to update this, let me maybe try exposing richer artifact information to avoid hardcoding the handling of sources and javadocs

coursier.bzl
```diff
@@ -497,6 +499,17 @@ def _pinned_coursier_fetch_impl(repository_ctx):
 
     http_files.extend(_get_jq_http_files())
 
+    http_files.extend([
+        "maven_artifacts = [%s]" % (",".join([
```
shs96c, line 503: This structure doesn't match the expected usage. If you tried to add these `maven_artifacts` to a target's `runtime_deps`, the build will fail, as the deps are expected to be labels and not structs.

I think there's a couple of ways of handling this:

1. Document this clearly in the README.md and explain how people could use it.
2. Use the maven coordinates directly, once they've been converted back from the dict they are to a string, and use the `artifact` macro to ensure they're loaded properly.

The first option is a lot less work, but I think the second one makes the output more usable.

shs96c (changes requested):
I can see why this would be a useful feature. Thank you for explaining.

coursier.bzl
dmivankov: one slight complication with coords is that `artifact` macro only works with `"groupId:artifactId"` and that there's strip classifier logic, need some work to extend `artifact` macro

for example for `com.github.jnr:jffi:jar:native:1.3.1`, `com.github.jnr:jffi:jar:sources:1.3.1` and `com.github.jnr:jffi:1.3.1` there are following targets
```
@maven//:com_github_jnr_jffi_native_1_3_1 // alias
@maven//:com_github_jnr_jffi_native // actual import
@maven//:com_github_jnr_jffi_jar_sources_1_3_1_extension
@maven//:com_github_jnr_jffi_jar_native_1_3_1_extension
@maven//:com_github_jnr_jffi_1_3_1_extension
@maven//:com_github_jnr_jffi_1_3_1 // alias
@maven//:com_github_jnr_jffi // import
```
so `artifact("com.github.jnr:jffi:jar:native:1.3.1")` needs to also strip `jar` classifier it seems

shs96c: The group and artifact IDs are the first two fields in the maven coordinates, which makes life a bit easier. I think you'd need to gather together all the imports, transform, and then sort, removing duplicates, and you'd be okay.

dmivankov: Currenty I can put `com.github.jnr:jffi:jar:native:1.3.1` and `com.github.jnr:jffi:1.3.1` as artifacts in `maven_install` and then refer to them as `@maven//:com_github_jnr_jffi_native` and `@maven//:com_github_jnr_jffi`.

But there seems to be no ready-made macro yet that'd transform string `"com.github.jnr:jffi:jar:native:1.3.1"` to label `@maven//:com_github_jnr_jffi_native`

```python
load("@rules_jvm_external//:defs.bzl", "artifact")
load("@rules_jvm_external//:specs.bzl", "parse")

s = "com.github.jnr:jffi:jar:native:1.3.1"
artifact(s) ==> @maven//:com_github_jnr_jffi # without _native part
parse.parse_maven_coordinate(s) ==> { group: "com.github.jnr", artifact: "jffi", classifier: "native", packaging: "jar", version: "1.3.1" }
artifact(parse.parse_maven_coordinate(s)) ==> still @maven//:com_github_jnr_jffi
```
I could decypher and replicate dict -> label conversion (aka `escape(strip_packaging_and_classifier_and_version(s))`), but maybe instead `artifact` macro should be extended to handle all potential keys (it already can accept dict, but decides to only use group and artifact)

dmivankov: Sounds quite doable, but maybe there's an even better way?

private/dependency_tree_parser.bzl
```diff
@@ -194,8 +194,7 @@ def _generate_imports(repository_ctx, dep_tree, explicit_artifacts, neverlink_ar
             for dep in artifact["directDependencies"]:
                 if get_packaging(dep) == "json":
                     continue
-                stripped_dep = strip_packaging_and_classifier_and_version(dep)
```
dmivankov, line 197: wasn't used

private/rules/artifact.bzl
```diff
@@ -1,19 +1,22 @@
...
 
-def _parse_artifact_str(artifact_str):
-    pieces = artifact_str.split(":")
-    if len(pieces) == 2:
-        return {"group": pieces[0], "artifact": pieces[1]}
-    else:
-        return parse.parse_maven_coordinate(artifact_str)
+def _ensure_version(a):
+    # fake :version part is due to get_simple_coord not supporting unversioned coords
+    return a if len(a.split(":")) > 2 else (a + ":fake_version")
```
dmivankov, line 17: ugly hack, want to reuse `get_simple_coord` logic with support for type and classifier, but it doesn't support unversioned artifacts, also only works on strings

coursier.bzl
dmivankov: Update PR with support for full string coords in `artifact` macro

private/rules/artifact.bzl
```diff
@@ -1,19 +1,22 @@
...
+
+def _make_artifact_str(artifact_obj):
+    # TODO: add support for optional type, classifier and version parts
```
dmivankov, line 20: gets a little complicated with dictionary case as there are multi-level optional fields so if `packaging` is absent but `classifier` is present would need to apply some default value to packaging, this is a new use-case and can be added later, for now dict version still only handles `group + artifact` only

dmivankov: calling `fail()` in cases where `packaging` is absent but `classifier` is present could be an option too

private/coursier_utilities.bzl
```diff
@@ -27,7 +27,7 @@ SUPPORTED_PACKAGING_TYPES = [
     "scala-jar",
 ]
 
-def strip_packaging_and_classifier(coord):
+def get_versioned_coord(coord):
```
cheister, line 30: Why these name changes?  Feels like the new names are not as clear.

private/rules/artifact.bzl
cheister: I don't understand why you need all of this?  Is it so you can keep the one call to `get_simple_coord` ?  Why can't this logic be put in `get_simple_coord` instead?

dmivankov: it can, but it may cause breaking changes to other use sites of that function

private/coursier_utilities.bzl
dmivankov: can revert renaming. New names describe the purpose and applicability (maybe even better would be `get_versioned_label_name` and `get_simple_label_name`? Didn't do that initially because the jump to labels requires an extra `escape` on top and there are places where `escape` isn't applied) and it makes it more clear why they can be reused to parse coord strings into labels, while old ones describe implementation (and applicability but for cases where we have purpose in terms of implementation / action on coord parts rather than in terms of ways to address one and the same artifact)

dmivankov: reverted the rename. The only place without escape is around checking against jetify_include_list which could be shifted to operate on labels rather than coords, and it may add support for classifier to jetify_include_list, but that can be a separate PR if needed at all

private/rules/artifact.bzl
dmivankov: did a quick look, looks like this should be safe so moved there now

dmivankov:
squashed and updated to remove renames and expose just the coords, making sure artifact macro now can handle coords with classifiers. Diff got smaller now

private/coursier_utilities.bzl
```diff
@@ -46,6 +46,10 @@ def strip_packaging_and_classifier(coord):
     return ":".join(coordinates)
 
 def strip_packaging_and_classifier_and_version(coord):
+    coordinates = coord.split(":")
```
cheister, line 49: Do you mind adding a test to `tests/unit/coursier_utilities_test.bzl` for this?

private/coursier_utilities.bzl
cheister: thanks, the diff is easier to follow without the renames.

private/coursier_utilities.bzl
dmivankov: added a test case now

cheister (approved):

jin (dismissed):
Given that this is a user-accessible feature, please document it in the README (or nobody will know it exists)

shs96c (approved):

shs96c (approved):
Thank you for being so patient with this PR, and thank you for documenting the new feature, @dmivankov. LGTM! :)

PR #649, opened 2022-01-04
Add new strict_visibility_value= attribute to allow exposing strict dependencies to some scopes

dmivankov:
Note that using package default_visibility doesn't always work so explicit visiblity attributes are needed
https://github.com/bazelbuild/bazel/issues/13681

Somewhat related to https://github.com/bazelbuild/rules_jvm_external/pull/648

In combination this for example allows to create a class_name -> maven artifact resolver for
transitive dependency classes while still keeping strict visibility (making transitive dependencies
visible only to the resolve cli tool)

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -1,18 +1,3 @@
-<!-- START doctoc generated TOC please keep comment here to allow auto update -->
-<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
-# API Reference
-
-- [Basic functions](#basic-functions)
-  - [javadoc](#javadoc)
-  - [java_export](#java_export)
-  - [maven_install](#maven_install)
-- [Maven specification functions](#maven-specification-functions)
-  - [maven.repository](#mavenrepository)
-  - [maven.artifact](#mavenartifact)
-  - [maven.exclusion](#mavenexclusion)
-
-<!-- END doctoc generated TOC please keep comment here to allow auto update -->
-
 # Basic functions
 
 These are the basic functions to get started.
--- a/docs/api.md
+++ b/docs/api.md
@@ -1,3 +1,18 @@
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+# API Reference
+
+- [Basic functions](#basic-functions)
+  - [javadoc](#javadoc)
+  - [java_export](#java_export)
+  - [maven_install](#maven_install)
+- [Maven specification functions](#maven-specification-functions)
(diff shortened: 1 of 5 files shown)


private/rules/maven_install.bzl
```diff
@@ -56,6 +57,8 @@ def maven_install(
       strict_visibility: Controls visibility of transitive dependencies. If `True`, transitive dependencies
         are private and invisible to user's rules. If `False`, transitive dependencies are public and
         visible to user's rules.
+      strict_visibility_value: Allows changing transitive dependencies strict visibility scope from private
+        to specifies scopes list.
```
shs96c, line 61: Nit: "specific" or "specified"

shs96c (changes requested):
That's an amazing Bazel bug. Thanks for spotting this and creating this PR. Could you also please run `./scripts/generate_docs.sh` to make sure that the new flag is properly documented?

dmivankov:
Tests failed, will have a look

docs/api.md
```diff
@@ -1,18 +1,3 @@
-<!-- START doctoc generated TOC please keep comment here to allow auto update -->
```
shs96c, line 1: Looks like you don't have doctoc installed: https://github.com/bazelbuild/rules_jvm_external#generating-documentation

shs96c (approved):
Thank you!

docs/api.md
dmivankov: oops, didn't notice this in the diff

dmivankov:
funny, cla check thinks @shs96c hasn't signed it even being a codeowner

shs96c:
@dmivankov could you please remove my commit from the history? The CLA bot isn't happy since it came from the GitHub UI and doesn't have my proper email address. I'll merge the PR when it gives me the green light.

dmivankov:
done, squashed and reathored commits related to toc

Kernald:
Because of another Bazel bug (https://github.com/bazelbuild/bazel/issues/13553), accessing the jar files directly was quite useful - see https://github.com/Bencodes/bazel_issue_13553/pull/1. This was possible because of the `default_visibility` at the `package` level, which is now commented out. Is there any way to still access the jar files directly?

dmivankov:
@Kernald jar files should still be accessible, do you have a more detailed repro case?
genrule converting downloaded files (like `@javax_inject_javax_inject_1//file:file`) to `@maven//...` jars should have same visibility as before. You may also try using downloaded jars directly (`bazel query --output build @maven//....jar`)

Kernald:
Such a query doesn't seem to yield anything, but pinning seems to fix the visibility issue. Which seems to point to a slightly different behaviour between 4.3 and 4.4.2 without pinning, but similar behaviour with pinning (and pinning is a totally acceptable solution as far as I'm concerned, I should have done that from the start anyway).

PR #652, opened 2022-01-10
pin.sh: also sort dependencies by coordinate

dhalperi:
Sorting the overall list of maven libraries by coordinate gives a stable order
with locality and makes it easy to understand/review changes.

When I wrote bazelbuild/rules_jvm_external#611, I did not believe this sort was
necessary; I think because something else was already producing dependencies in
sorted order. Maybe an upgrade to coursier or something else has changed this.

Either way, explicitly sorting by .coord is idempotent at worst, and it
guarantees the output will be sorted.

Also ran the following repinning commands, not all of which changed pinned files:

```
bazel run @unpinned_maven//:pin
bazel run @unpinned_policy_pinned_testing//:pin
bazel run @unpinned_maven_install_in_custom_location//:pin
bazel run @unpinned_json_artifacts_testing//:pin
bazel run @unpinned_unsafe_shared_cache_with_pinning//:pin
bazel run @unpinned_manifest_stamp_testing//:pin
bazel run @unpinned_regression_testing//:pin
```

### DIFF
--- a/tests/custom_maven_install/unsafe_shared_cache_with_pinning_install.json
+++ b/tests/custom_maven_install/unsafe_shared_cache_with_pinning_install.json
@@ -6,17 +6,6 @@
         "conflict_resolution": {},
         "dependencies": [
             {
-                "coord": "com.google.code.findbugs:jsr305:jar:sources:3.0.2",
-                "dependencies": [],
-                "directDependencies": [],
-                "file": "v1/https/repo1.maven.org/maven2/com/google/code/findbugs/jsr305/3.0.2/jsr305-3.0.2-sources.jar",
-                "mirror_urls": [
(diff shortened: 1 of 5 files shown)


tests/custom_maven_install/unsafe_shared_cache_with_pinning_install.json
```diff
@@ -203,10 +198,15 @@
                 "url": "https://repo1.maven.org/maven2/org/codehaus/mojo/animal-sniffer-annotations/1.17/animal-sniffer-annotations-1.17.jar"
             },
             {
-                "coord": "com.google.guava:listenablefuture:jar:sources:9999.0-empty-to-avoid-conflict-with-guava",
```
dhalperi, line 206: Example of fix: this was for some reason last instead of near the other `com.google.guava` dependencies.

jin (approved):
thank you!

PR #654, opened 2022-01-14
Bzlmod proof of concept

shs96c:
This is an exploration of whether `bzlmod` could be used to replace the existing `rules_jvm_external` workspace rules. 

TL;DR: not yet, since there's no way to express complex types such as how we handle exclusions in artifacts, but other than that things are looking hopeful.

### DIFF
--- a/examples/bzlmod/.bazelrc.example
+++ b/examples/bzlmod/.bazelrc.example
@@ -1,3 +1,2 @@
 common --experimental_enable_bzlmod
-
-build --javacopt="--release 11"
+build --java_language_version=11
--- a/examples/bzlmod/java/src/com/github/rules_jvm_external/examples/bzlmod/BUILD
+++ b/examples/bzlmod/java/src/com/github/rules_jvm_external/examples/bzlmod/BUILD
@@ -6,6 +6,5 @@ java_binary(
     deps = [
         artifact("com.google.guava:guava"),
         artifact("org.seleniumhq.selenium:selenium-api"),
-        artifact("org.seleniumhq.selenium:selenium-chrome-driver"),
     ],
 )
--- a/examples/bzlmod/java/src/com/github/rules_jvm_external/examples/bzlmod/Main.java
+++ b/examples/bzlmod/java/src/com/github/rules_jvm_external/examples/bzlmod/Main.java
@@ -22,6 +22,6 @@ public class Main {
         // Make sure we can import something from the selenium-api jar
         // We don't want to fire up a full driver, because we might not
         // have access to the binaries.
-        WebDriver = null;
+        WebDriver driver = null;
     }
 }


MODULE.bazel
```diff
@@ -0,0 +1,33 @@
+module(
+    name = "rules_jvm_external",
+    version = "1.0"
+)
+
+bazel_dep(name = "bazel_skylib", version = "1.1.1")
+bazel_dep(name = "rules_java", version = "5.0.0")
```
jin, line 7: What are we using rules_java for? IIUC the Starlark rewrite uses "builtins" to manage migration now, so no need for rules_java loads.

private/extensions/http_file.bzl
```diff
@@ -0,0 +1,34 @@
+load("@bazel_tools//tools/build_defs/repo:http.bzl", bazel_http_file = "http_file")
+
+http_download = tag_class(
```
jin, line 3: Seems like this should be part of skylib or something

private/extensions/maven.bzl
```diff
@@ -0,0 +1,80 @@
...
+            default = ["https://repo1.maven.org/maven2"],
+        ),
+    }
+)
+
+def _maven_impl(mctx):
+    # Gather the artifacts
+    repos = {}
+
+    for mod in mctx.modules:
```
jin, line 26: This is exciting :) have you tried making this work in a transitive project environment?

marcguilera:
Is this actively being worked on? Would be cool to see :)

yuanweixin:
I am curious what is the timeline/deadline for migrating to bzlmod, if one exist? I peeked at the other language rules but didn't spot code or pull requests like this one.

shs96c:
This is being actively worked on. It fell down my list or priorities, but it's back there now :)

shs96c:
This PR depends on some of the other ones I've filed in the repo, but the `bzl` files are ready for review.

Wyverald (approved):
I had a brief look and the usages of Bzlmod stuff seem pretty good to me. I didn't look at the precise business logic but that's kind of outside my purview anyway. Ship it :)

private/extensions/maven.bzl
```diff
@@ -0,0 +1,348 @@
...
+            for repository in parse.parse_repository_spec_list(install.repositories):
+                repo_string = _json.write_repository_spec(repository)
+                if repo_string not in existing_repos:
+                    existing_repos.append(repo_string)
+            repo["repositories"] = existing_repos
+
+            if install.lock_file:
+                lock_file = repo.get("lock_file")
+                if lock_file and lock_file != install.lock_file:
+                    fail("There can only be one lock file. Values were %s and %s" % (lock_file, install.lock_file))
```
jin, line 222: suggestion: "There can be only one lock file for the repo %s"

private/extensions/maven.bzl
```diff
@@ -0,0 +1,348 @@
...
+
+            existing_artifacts.append(to_add)
+            repo["artifacts"] = existing_artifacts
+            repos[artifact.name] = repo
+
+        for install in mod.tags.install:
+            repo = repos.get(install.name, {})
+
+            artifacts = repo.get("artifacts", [])
+            repo["artifacts"] = artifacts + install.artifacts
```
jin, line 210: Does this mean that if module A depends on module B, and both A and B has `maven.install` that use the same `install.name`, their artifacts will be flattened into the same list?

jin (approved):

PR #657, opened 2022-01-17
Don't use jcenter.bintray.com in test, it is sort of EOL

dmivankov:
Replace with https://repo1.maven.org/maven2

Test can time out fetching from jcenter like in
https://buildkite.com/bazel/rules-jvm-external/builds/2461#c70711ec-e59f-48e9-b925-7d755a014070

It seems that in the end it is supposed to be maintained as read-only but atm times out
https://jfrog.com/blog/into-the-sunset-bintray-jcenter-gocenter-and-chartcenter/

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -420,7 +420,7 @@ maven_install(
         "com.google.code.findbugs:jsr305:3.0.2",
     ],
     repositories = [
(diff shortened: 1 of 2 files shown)


jin (approved):

jin:
thank you!

PR #659, opened 2022-01-19
Use bazel's native json decode method

shs96c:
This means that `rules_jvm_external` is only compatible with
Bazel 4 and later. That shipped on 2021-01-21, and so should
be widely adopted already. Older versions of this ruleset
will continue to work for repos using older versions of Bazel
so it's not removing any existing functionality, and the new
Bazel 5 release should have already shipped.
    
The main reason for doing this is that the native json
decoding is significantly faster than the starlark-based one.
    
The following options were explored as alternatives:
    
1. Making the `json_parse` method check the bazel version and
   optionally use the `json` module if it was present. This
   doesn't work because of
   https://github.com/bazelbuild/bazel/issues/12735
2. Create a `json` repository rule that creates a repo that
   exposes a `json.bzl` file that delegates to the native
   `json` module if present, or the starlark one if not
   (as determined by looking at the major version number of
   bazel). This doesn't work because `maven_install` is used
   by RJE's own `repositories.bzl` file, so there's no chance
   to load the repository rule before it's needed.


coursier.bzl
```diff
@@ -132,7 +122,9 @@ def _execute(repository_ctx, cmd, timeout = 600, environment = {}, progress_mess
 # The representation of a Windows path when read from the parsed Coursier JSON
 # is delimited by 4 back slashes. Replace them with 1 forward slash.
 def _normalize_to_unix_path(path):
-    return path.replace("\\\\", "/")
+    if path == None:
```
jin, line 125: Why this change? Seems error prone to have `None` paths being passed around.

coursier.bzl
```diff
@@ -227,7 +219,7 @@ def _compute_dependency_tree_signature(artifacts):
         if artifact["file"] != None:
             artifact_group.extend([
                 artifact["sha256"],
-                artifact["file"],
+                artifact["file"].replace("\\", "/"),  # Make sure we represent files in a stable way cross-platform
```
jin, line 222: why not use `_normalize_to_unix_path`? here and below

jin (approved):
nice! some nits.

coursier.bzl
shs96c: Just being paranoid. I'll back it out.

coursier.bzl
shs96c: Done

PR #666, opened 2022-02-09
Use buffered IO when writing the output jar.

mauriciogg:
This improves io performance for large jars
Running this code on my machine with a compressed 70MB zip with ~40K
entries now takes ~5s from ~25s. For the same uncompressed jar goes from ~20s to a subsecond operation.

### DIFF
--- a/private/tools/java/rules/jvm/external/jar/AddJarManifestEntry.java
+++ b/private/tools/java/rules/jvm/external/jar/AddJarManifestEntry.java
@@ -5,6 +5,9 @@ import rules.jvm.external.ByteStreams;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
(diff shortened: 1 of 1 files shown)


shs96c:
The test failures look related to this change. Could you please make sure that `bazel test tests/com/...` passes for you?

mauriciogg:
> The test failures look related to this change. Could you please make sure that `bazel test tests/com/...` passes for you?

The failing tests would have been hard to fix using the FileSystem API, since the manifest has to be the first (not sure why java needs it there) entry and FileSystem hides the zip impl details and operates at a higher level. Turns out that simply wrapping the JarOutputStream around a buffered stream makes the performance good enough so I ended up just going that route. I updated the PR to reflect this change.

shs96c (approved):
LGTM. Thank you for the patch, and the nice performance boost! :)

PR #673, opened 2022-03-17
Allow compile jars to strip illegal automatic module names

shs96c:
There are plenty of jars out there that have an `Automatic-Module-Name`
that isn't legal. Normally, this doesn't matter too much, but when
compiling code with a modern compiler bazel will use a module path,
and that cares _a lot_, causing builds to fail.

To rectify this, we can pass an additional flag when creating the
compile jar that will strip illegal automatic module names from the
generated manifests.

### DIFF
--- a/private/tools/java/rules/jvm/external/jar/AddJarManifestEntry.java
+++ b/private/tools/java/rules/jvm/external/jar/AddJarManifestEntry.java
@@ -2,8 +2,8 @@ package rules.jvm.external.jar;
 
 import rules.jvm.external.ByteStreams;
 
-import java.io.IOException;
 import java.io.BufferedOutputStream;
+import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.nio.file.Files;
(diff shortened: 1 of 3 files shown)


cheister:
Do you mind adding an example error to the description when this isn't fixed? I'm curious what the problems actually look like.

private/tools/java/rules/jvm/external/jar/AddJarManifestEntry.java
```diff
@@ -165,12 +236,94 @@ public static void addEntryToManifest(
...
+          List<String> toRemove,
+          boolean makeSafe) {
     manifest.getMainAttributes().put(new Attributes.Name("Created-By"), "AddJarManifestEntry");
     toAdd.forEach(manifestEntry -> {
       String[] manifestEntryParts = manifestEntry.split(":", 2);
       manifest.getMainAttributes().put(new Attributes.Name(manifestEntryParts[0]), manifestEntryParts[1]);
     });
     toRemove.forEach(name -> manifest.getMainAttributes().remove(new Attributes.Name(name)));
+
+    if (!makeSafe) {
```
cheister, line 252: Why not `if (makeSafe) {
  checkAutomaticModuleName(jar, manifest);
}` ?

jin (approved):

PR #674, opened 2022-03-24
java_export to inherit testonly on its sub-targets

tsawada:
I was trying to use `maven_artifacts` via `java_export` as suggested in #648, and noticed that -pom targets fail to build when they include testonly targets as deps.

I am not sure if this is a good way to resolve the issue (whether making pom target as testonly makes sense) though.

### DIFF
--- a/private/rules/java_export.bzl
+++ b/private/rules/java_export.bzl
@@ -10,6 +10,7 @@ def java_export(
         pom_template = None,
         visibility = None,
         tags = [],
+        testonly = False,
         **kwargs):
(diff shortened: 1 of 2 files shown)


tsawada:
cc @dmivankov

dmivankov:
Makes sense to me: `testonly`was already being passed (if it is set) from `java_export` via `kwargs` to `java_library`, this PR propagates it to other targets under in `maven_export`/`java_export`

shs96c (approved):
This makes sense to me. Thank you for the PR!

PR #677, opened 2022-04-04
Upgrade protobuf version to 3.20.0

Wyverald:
We were using a very old version of protobuf (3.8.0-rc1) and it was in turn using a zlib version that's no longer there.


Wyverald:
ping -- we can't re-enable "rules_jvm_external - examples" on CI until this is merged.

cheister (approved):

Issue #582, opened 2021-07-30
`override_targets` can create duplicate dependencies


jdai8:
[Guice multibindings](https://github.com/google/guice/wiki/Multibindings) was added to Guice core in 4.2. Our repo only uses Guice 4.2+, so we'd like to point multibindings to guice:
```
override_targets = {
  "com.google.inject.extensions:guice-multibindings": "@maven//:com_google_inject_guice"
}
```

For dependencies which depend on both guice and guice multibindings, for example [nifty-core](https://repo1.maven.org/maven2/com/facebook/nifty/nifty-core/0.23.0/nifty-core-0.23.0.pom), this results in a duplicate dependency in the generated rule:
```
jvm_import(
	name = "com_facebook_nifty_nifty_core",
	jars = [...],
	srcjar = "...",
	deps = [
		":io_airlift_log",
		":io_airlift_units",
		":com_google_code_findbugs_annotations",
		"@maven//:com_google_inject_guice",
		":io_netty_netty",
		":com_google_guava_guava",
		":javax_inject_javax_inject",
		":com_google_inject_guice",
	],
	tags = ["maven_coordinates=com.facebook.nifty:nifty-core:0.23.0"],
)
```

which fails with an error that looks like:
```
Label '@maven//:com_google_inject_guice' is duplicated in the 'deps' attribute of rule 'com_facebook_nifty_nifty_core'
```

jdai8:
One solution might be to still generate the "overriden" target:
```
jvm_import(
	name = "com_facebook_nifty_nifty_core",
	jars = [...],
	srcjar = "...",
	deps = [
		":io_airlift_log",
		":io_airlift_units",
		":com_google_code_findbugs_annotations",
		":com_google_inject_extensions_guice_multibindings",
		":io_netty_netty",
		":com_google_guava_guava",
		":javax_inject_javax_inject",
		":com_google_inject_guice",
	],
	tags = ["maven_coordinates=com.facebook.nifty:nifty-core:0.23.0"],
)
```

And take advantage of this alias:
```
alias(
	name = "com_google_inject_extensions_guice_multibindings",
	actual = "@maven//:com_google_inject_guice",
	visibility = ["//visibility:public"],)
```

That way we don't have to worry about parsing targets/repositories/etc. in an effort to dedupe.

PR #678, opened 2022-04-05
Use alias instead of override destination.

samshadwell:
Resolves https://github.com/bazelbuild/rules_jvm_external/issues/582

## Problem

That issue is caused by overrides to fully-qualified labels being duplicative with their non-overridden counterparts. For example, consider [ndbench-core](https://mvnrepository.com/artifact/com.netflix.ndbench/ndbench-core/0.7.2), which takes a dependency on both `com.codahale.metrics:metrics-core` and `io.dropwizard.metrics:metrics-core`. Those two libraries expose many of the same fully-qualified classes, and we'd like to only have one version of each class on the classpath.

To accomplish this, we could use an override like:
```
override_targets = {
    "com.codahale.metrics:metrics-core": "@regression_testing//:io_dropwizard_metrics_metrics_core",
},
```
so that the "old" codahale metrics-core is replaced with the "new" dropwizard one.

Currently, this results in generated deps like:
```
...
jvm_import(
...
	deps = [
                ...
		"@regression_testing//:io_dropwizard_metrics_metrics_core",
		...
		":io_dropwizard_metrics_metrics_core",
		...
	],
...
```

Analysis of this fails with errors like:
```
ERROR: /private/var/tmp/_bazel_sam_shadwell/7db21a72dc367049bac9ba2882a68834/external/regression_testing/BUILD:2469:11: Label '@regression_testing//:io_dropwizard_metrics_metrics_core' is duplicated in the 'deps' attribute of rule 'com_netflix_ndbench_ndbench_core'
```

## Solution

As suggested by @jdai8, the solution is to instead reference the generated alias in deps (instead of eagerly resolving the alias as is done now).

So now we generate deps like:
```
alias(
	name = "com_codahale_metrics_metrics_core",
	actual = "@regression_testing//:io_dropwizard_metrics_metrics_core",
	visibility = ["//visibility:public"],)
...
jvm_import(
...
	deps = [
                ...
		":com_codahale_metrics_metrics_core",
		...
		":io_dropwizard_metrics_metrics_core",
		...
	],
...
```


alexeagle:
Ping @shs96c - it looks reasonable to me. (I'm working with Sam)

samshadwell:
Welp, we're going on 2 years now and no word from anyone.

While I still think this would be a positive change, I'm no longer working on Bazel-related things. Closing this PR to clean up my personal PR dashboard.

shs96c:
I can't believe I missed this. My apologies, @samshadwell

shs96c:
Let me see if this applies on the current tip-of-tree.

PR #681, opened 2022-04-18
Stop hard-coding `-source 8 -target 8`

cushon:
And instead compile with whatever the default `--tool_java_language_version` is. Explicitly specifying a language level for individual targets should be unecessary.

I noticed this debugging https://github.com/bazelbuild/bazel/issues/15272


cheister:
Should `--tool_java_language_version` be added to `.bazelrc` as well to make sure we're building Java 8 jars?

cushon:
Why not just use the `--tool_java_language_version`? Is there anything about these tools that wouldn't be compatible with newer language levels?

shs96c:
We should be able to land this safely enough. Let's see what happens.

PR #685, opened 2022-04-25
WIP: proof of concept for license splicing 

aiuto:
@danielmachlab This is proof of concept of the license lookup indirection.

I am not happy that we can't get a fallback repository.  But, it could be done in setup,
with the requirement that you call setup.

### DIFF
--- a/license_demo/BUILD
+++ b/license_demo/BUILD
@@ -1,0 +1,9 @@
+
+
+java_library(
+    name = "my_lib",
+    srcs = ["MyLib.java"],
+    deps = [
+        "@maven//:junit_junit",
+    ],
+)
--- a/license_demo/MyLib.java
+++ b/license_demo/MyLib.java
@@ -1,0 +1,10 @@
+/**
+ * 
+ */
+final class MyLib {
+
+  public MyLib() {
+  }
+
+  int foo() { return 1; }
+}
--- a/license_demo/WORKSPACE
+++ b/license_demo/WORKSPACE
@@ -1,0 +1,61 @@
+# Copyright 2022 The Bazel Authors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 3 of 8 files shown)


aiuto:
This is quite annoying.
- The primary setup method `rules_jvm_external_deps()` calls `maven_install` to create the repo `rules_jvm_external_deps`
- so maven_install can't load the the classifier repository unless that was created
- but the place you want to create it is in `rules_jvm_external_deps()`
- then the `setup` method depends on the repo `rules_jvm_external_deps`

Sigh...

danielmachlab:
@aiuto this workflow looks good, especially considering the restrictions we have. It all works locally for me. What should the next steps be? It might be useful to get a third opinion on the direction we are heading in while finishing up the prototype `lookup_license` function. I can make a PR tomorrow or Friday into this PR to fill in the `lookup_license` function.

aiuto:
I don't know. I don't really like that you need a third setup function for this repository. I want to get the @license_checker repository injected by one of the first rules, so that all the examples don't have to change.  My thought is that I could point to rules_jvm_external for the checker as a default, and add an optional rule that rewrites the dependency reader to point to the other repository.

But... you could certainly use what I have here to test the license injection at your scale.

jin:
I'm quite confused about what's happening here. Is there an issue/context for this FR?

danielmachlab:
Hi @jin we are trying to find a good way to allow users of rules_jvm_external to apply rules_license `license`s to the artifacts that they import with `maven_install`. This is coming out of the new rules_license SIG (see monthly meeting notes [here](https://docs.google.com/document/d/14jAqgGfLHwrJzBR7A0-ndLyYxUC1rg3vf0Zdtc4zKd0/edit#heading=h.pni07mn1rdsf)). [Here](https://docs.google.com/document/d/1XnDEQArn_hAIiExSHPgo8v2UZWP3eRf0OtUldLuHso8/edit#heading=h.wa7ly5ouolxq) is a document which describes our problem and our original approach for this change. The initial approach didn't work out since we can't pass a macro to a repository rule.

danielmachlab:
Hi @aiuto, circling back to this. In this approach, why couldn't we make it so it is backward compatible where users not using a license classifier can avoid adding `use_default_license_classifier()` to their `WORKSPACE`?

I'm thinking that the optimal case would be for a user to:
1. create a file defining `lookup_license`
2. load `set_license_classifier` from rje, which accepts the path of the `lookup_license` file.

Issue #686, opened 2022-04-26
Using netty-tcnative-boringssl-static as a dependency generates internal dependency loop


tjoneslo:
If I try to include `"io.netty:netty-tcnative-boringssl-static:2.0.49-Final"` or later version into my maven install it generates a dependency loop in the internal setup:

```
ERROR: /private/var/tmp/_bazel_tjoneslo/ee798b775802fcb379cfe763e38aeaf8/external/maven/BUILD:290:11: in jvm_import rule @maven//:io_netty_netty_tcnative_boringssl_static_linux_aarch_64: cycle in dependency graph:
    //:server (63684d86ff47f33398646ba0e3860d9ff1fcad99873cee4f2aec407f500adbdd)
    @maven//:io_servicetalk_servicetalk_http_netty (63684d86ff47f33398646ba0e3860d9ff1fcad99873cee4f2aec407f500adbdd)
    @maven//:io_servicetalk_servicetalk_dns_discovery_netty (63684d86ff47f33398646ba0e3860d9ff1fcad99873cee4f2aec407f500adbdd)
    @maven//:io_servicetalk_servicetalk_transport_netty (63684d86ff47f33398646ba0e3860d9ff1fcad99873cee4f2aec407f500adbdd)
    @maven//:io_servicetalk_servicetalk_transport_netty_internal (63684d86ff47f33398646ba0e3860d9ff1fcad99873cee4f2aec407f500adbdd)
    @maven//:io_netty_netty_tcnative_boringssl_static_osx_aarch_64 (63684d86ff47f33398646ba0e3860d9ff1fcad99873cee4f2aec407f500adbdd)
.-> @maven//:io_netty_netty_tcnative_boringssl_static_linux_aarch_64 (63684d86ff47f33398646ba0e3860d9ff1fcad99873cee4f2aec407f500adbdd)
|   @maven//:io_netty_netty_tcnative_boringssl_static_windows_x86_64 (63684d86ff47f33398646ba0e3860d9ff1fcad99873cee4f2aec407f500adbdd)
`-- @maven//:io_netty_netty_tcnative_boringssl_static_linux_aarch_64 (63684d86ff47f33398646ba0e3860d9ff1fcad99873cee4f2aec407f500adbdd)
ERROR: Analysis of target '//:server' failed; build aborted
```

[...]

PR #687, opened 2022-04-28
Fix the Boringssl static dependency loop

tjoneslo:
Addresses [issue](https://github.com/bazelbuild/rules_jvm_external/issues/686) 

Check for the specific failure case seen by the netty-tcnative-boringssl-static library by identifying the specific problem as seen and skipping the dependency loop generation in the `BUILD` file generation. This may be too fragile, but does address the problem.


shs96c (approved):
Thank you @tjoneslo! Working code is far better than a perfect solution in this case. I appreciate the effort digging into this and fixing it :)

PR #688, opened 2022-04-28
Factor repositories into artifact cache 

illicitonion:
Which repositories we consult can affect both whether a query is
successful, and its output (e.g. `url` and `mirror_urls` may be set
differently). Accordingly, we should factor it into the key we use to
determine whether lockfiles are stale.


shs96c (approved):
Nice catch. LGTM. Thank you!

PR #692, opened 2022-04-28
WIP: License lookup - a better version

aiuto:
This is a backwards compatible approach.  It is so ugly it is delicious.

We start with a fallback license_lookup, implemented in rules_jvm_external.  If you do nothing else you get that, so no license information is added.

If you want to override that with your own method:
- implement //:license_classifier.bzl%lookup_license in a repository of your choosing. It could be a local one.
- load that repository before rules_jvm_external.
- load rules_jvm_external with a patch_cmd that points to your classifier repo.

```python
local_repository(
  name = "my_compliance",
  path = "compliance",
)

http_archive(
    name = "rules_jvm_external",
    urls = ["file:///tmp/rje.tar"],
    patch_cmds = [
      "./use_license_classifier.sh my_compliance",
    ]
)
```

Note that the patch command is actually implemented in rules_jvm_external, so you don't have to write your own.


license_demo/WORKSPACE
```diff
@@ -0,0 +1,60 @@
...
+  name = "my_compliance",
+  path = "compliance",
+)
+
+
+# Point to parent for testing. You would use http_archive in real life.
+http_archive(
+    name = "rules_jvm_external",
+    # path = "..",
+    urls = ["file:///tmp/rje.tar"],
```
danielmachlab, line 40: How did you generate this `.tar`? I ended up just using `git_repository()` workspace rule pointing at this branch as a work-around

danielmachlab (commented):
Mmmm, yes, delicious! This definitely works. As you pointed out it is cumbersome, but it is backwards compatible. Sorry for the delayed review, I was OOO most of last week. 

For a new user, what they need to do is:
1. define a new repo with `lookup_license`
2.  add the `patch_cmd` to the `http_archive` or `git_repository` call

This is not too bad, but could be a little complex to onboard some. Thats okay.

When comparing this with #685, I left [a comment](https://github.com/bazelbuild/rules_jvm_external/pull/685#issuecomment-1122736307) on #685

license_demo/WORKSPACE
aiuto: Tar from the command line. This is just to get something to demo, not what we would really do.

PR #695, opened 2022-05-01
Add support for m2Local repository

smocherla-brex:
Piggybacking off #532, fixing some merge conflicts and adding the necessary integration tests to ensure the feature is in a state to merge. Opened a separate PR as I don't have the permissions to push changes to that PR. Happy to reconcile in that PR if that's ideal.

I've added a test in `tests/run_bazel_tests.sh` to build and publish a target within the repo `//tests/integration/kt_jvm_export:test.publish` to the local maven repo, and have a maven_install for this testcase that refers to maven coordinates for that rule `com.example:kt.1.0.0`, and validate that the local artifact is sourced in that testcase. I can add additional testcases if required.

It appears to be running successfully on all the 3 OS'es from the links I checked. I had to update a bunch of `maven_install.json` files for the examples, since the builds determined the hashes were out of date.


smocherla-brex:
@jin Could you take a look and let me know if the integration tests here suffice? Thanks

coursier.bzl
```diff
@@ -487,6 +502,10 @@ def _pinned_coursier_fetch_impl(repository_ctx):
                 # contains the mirror_urls field.
                 http_files.append("        urls = [\"%s\"]," % artifact["url"])
             http_files.append("    )")
+        elif is_maven_local_path(artifact["file"]):
+            # This file comes from maven local, so instead of creating an http_file for it, we simply
+            # simlink the file from the mavel local directory to file within the repository rule workspace
```
jin, line 507: ..symlink the file from the maven..

tests/bazel_run_tests.sh
```diff
@@ -15,6 +15,19 @@ function test_duplicate_version_warning() {
   expect_log "Successfully pinned resolved artifacts"
 }
 
+function test_m2local_testing_found_local_artifact() {
```
jin, line 18: can you also add a couple of more tests:

1) do a full build on the jar target (`@m2local_testing//:com_example_kt`) instead of running the `pin` executable
2) manually create the m2local directory instead of relying on java_export, so we have a test that doesn't depend on java_export's semantics

coursier.bzl
smocherla-brex: Fixed, thanks for pointing out

tests/bazel_run_tests.sh
smocherla-brex: Added 1 test using the `@m2local_testing//:com_example_kt` build target.

Also added another test, that builds the kt library, and copies binaries from bazel-bin and populate it into m2local, and then build the testing target `@m2local_testing//:com_example_kt)` (after a bazel clean) to validate that it is fetched from m2local. Checksums don't exist in this testcase, so I created a WORKSPACE repository that has `fail_on_missing_checksums = False`.

jin (approved):

jin:
@cheister @shs96c this is a pretty fundamental FR, mind taking a look too?

smocherla-brex:
@cheister @shs96c Could I have your eyes on this PR? Thanks

jin (approved):

PR #705, opened 2022-06-02
Build using Bazel 5.1.1

Bencodes:


shs96c (approved):
CI passes, and we should have done this a while ago. Thank you for the update!

PR #716, opened 2022-07-05
Add index of packages in jar files when pinning

illicitonion:
This allows other tooling, e.g. gazelle, to make use of this
information. This can be useful for dependency inference, dependency
analysis, and suggesting missing dependencies.

This makes sense to do at pin time because the lifecycle of this
information is closely tied to the artifacts being fetched/pinned.
Generating the information is cheap enough that this shouldn't
significantly slow down pins, but is expensive enough that other tooling
may want to avoid doing so on the fly at runtime, so is useful to cache.


coursier.bzl
```diff
@@ -119,6 +119,35 @@ def _execute(repository_ctx, cmd, timeout = 600, environment = {}, progress_mess
...
+def _execute_with_argsfile(
+        repository_ctx,
+        tool,
+        tool_name,
+        progress_message,
+        error_description,
+        files_to_inspect):
+    # Avoid argument limits by putting list of files to hash into a file
+    repository_ctx.file(
+        "{}_argsfile".format(tool_name),
```
shs96c, line 131: This implies that each tool can only ever be used once....

private/tools/java/rules/jvm/external/jar/BUILD
```diff
@@ -10,6 +10,23 @@ java_binary(
     ],
 )
 
+java_library(
+    name = "ListPackages-lib",
+    srcs = ["ListPackages.java"],
+    deps = [
+        "@maven//:com_google_code_gson_gson",
```
shs96c, line 17: You don't want to depend on the default `@maven` namespace here, since it means that all users of `rules_jvm_external` will need to add gson to their dependencies. Use `@rules_jvm_external_deps` which is created in `//:repositories.bzl`

private/tools/java/rules/jvm/external/jar/ListPackages.java
```diff
@@ -0,0 +1,84 @@
...
+  public static void main(String[] args) throws IOException {
+    if (args.length != 2 || !"--argsfile".equals(args[0])) {
+      System.err.printf("Required args: --argsfile /path/to/argsfile%n");
+      System.exit(1);
+    }
+
+    SortedMap<String, SortedSet<String>> index = new TreeMap<>();
+
+    Files.lines(Paths.get(args[1]))
+        .parallel()
```
shs96c, line 33: Since this is doing disk IO, does being in parallel make much difference?

private/tools/java/rules/jvm/external/jar/ListPackages.java
```diff
@@ -0,0 +1,84 @@
...
+            throw new RuntimeException(e);
+          }
+        });
+
+    System.out.println(new Gson().toJson(index));
+  }
+
+  public static SortedSet<String> process(Path path) throws IOException {
+    SortedSet<String> packages = new TreeSet<>();
+    try (InputStream fis = Files.newInputStream(path);
```
shs96c, line 48: Wrap `fis` in a `BufferedInputStream`

private/tools/java/rules/jvm/external/jar/ListPackages.java
```diff
@@ -0,0 +1,84 @@
...
+    SortedMap<String, SortedSet<String>> index = new TreeMap<>();
+
+    Files.lines(Paths.get(args[1]))
+        .parallel()
+        .forEach(path -> {
+          try {
+            SortedSet<String> packages = process(Paths.get(path));
+            index.put(path, packages);
+          } catch (IOException e) {
+            throw new RuntimeException(e);
```
shs96c, line 39: `UncheckedIOException` is now available and better communicates the problem.

coursier.bzl
illicitonion: Once per repository, yeah. Added a comment :)

private/tools/java/rules/jvm/external/jar/BUILD
illicitonion: Fixed - thanks!

private/tools/java/rules/jvm/external/jar/ListPackages.java
illicitonion: It does! After adding the `BufferedInputStream`, this provided an improvement for a large sample repo I have from in the 5-6 second range down to in the 1-2 second range. (Before adding the `BufferedInputStream` it was even bigger; 12-13 seconds -> 1-3 seconds)

That said, I've rewritten how the parallel collection happens, because parallel writes to a non-concurrent-safe collection were causing the CI failures - this now does a `collect` rather than writes from threads.

private/tools/java/rules/jvm/external/jar/ListPackages.java
```diff
@@ -0,0 +1,84 @@
...
+    }
+    int skip = 0;
+    // As per https://docs.oracle.com/en/java/javase/13/docs/specs/jar/jar.html
+    if (parts.length > 3 && "META-INF".equals(parts[0]) && "versions".equals(parts[1]) && isNumericVersion(parts[2])) {
+      skip = 3;
+    }
+
+    // -1 for the class name, -skip for the skipped META-INF prefix.
+    int limit = parts.length - 1 - skip;
+    return Arrays.stream(parts).skip(skip).limit(limit).collect(Collectors.joining("."));
```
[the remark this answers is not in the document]
illicitonion: This one is the path components of the package, so they shouldn't be sorted. Everything else (lists of packages, dicts from jar to packages) is using sorted sets/maps, so should be appropriately sorted.

private/tools/java/rules/jvm/external/jar/ListPackages.java
```diff
@@ -0,0 +1,91 @@
...
+        })
+        .collect(Collectors.toMap(
+            entry -> entry.getKey(),
+            entry -> entry.getValue(),
+            (left, right) -> {
+              throw new RuntimeException("Duplicate keys detected but not expected");
+            },
+            TreeMap::new
+        ));
+    System.out.println(new Gson().toJson(index));
```
shs96c, line 50: Please remove :)

shs96c (approved):
One small nit, then it's good to go

private/tools/java/rules/jvm/external/jar/ListPackages.java
illicitonion: This is the output of the tool!

shs96c: Well that's okay then :)

shs96c (approved):
CI builds are passing, and the changes to the lock files seem fine. LGTM

PR #717, opened 2022-07-05
Update rules_kotlin version

Wyverald:


Wyverald:
This fails with Bazel 4.x -- raised it with rules_kotlin: https://github.com/bazelbuild/rules_kotlin/pull/782#issuecomment-1175070199

PR #724, opened 2022-08-02
Add workspace_prefix_for_pinned_deps

tgeng:
When artifacts are pinned, `maven_install` would create individual
external repos for each dependency. The name of such an external repo is
simply the maven coordinate with all punctuations replaced by `_`. This
can cause name conflicts for code bases that are already using
`http_archive` to download thousands of dependencies through some other
mechanism but would like to migrate to use `maven_install`.

This PR tries to fix this by specifying a prefix for intermediate
external repos so that name collision becomes manageable.


tgeng:
Hi @cheister , can you take a look at this change? Thanks!

tgeng:
Hi @shs96c can you take a look at this?

shs96c:
Apologies for the slow review on this. Here now!

First comment: if you want a prefix, then I'd suggest defaulting to the repository rule's `name`. Secondly, if the versions pointed to are the same (down to the sha256), then it shouldn't matter which one is being used or where it's being pulled from. If this is the case, then it would be better to use a `maybe` to load the `http_file`, rather than a prefix.

Would that approach work for you?

tgeng:
> Apologies for the slow review on this. Here now!
> 
> First comment: if you want a prefix, then I'd suggest defaulting to the repository rule's `name`. Secondly, if the versions pointed to are the same (down to the sha256), then it shouldn't matter which one is being used or where it's being pulled from. If this is the case, then it would be better to use a `maybe` to load the `http_file`, rather than a prefix.
> 
> Would that approach work for you?

Hi, `maybe` would not solve our problem because in our current code base we are creating external repos with a different mechanism, which could result in artifacts that are of the same version but loaded from different sources (and hence has different SHA). My purpose of this PR is to allow easier adoption of `maven_install` in so that it will not interfere with existing code.

Defaulting the prefix to repo name would be a breaking change for existing users. Also, moving forward, with our use case (and typical use cases of others), we actually want to share the same intermediate artifacts across multiple pinned maven install. So it seems defaulting to `""` is still the desirable behavior for most use cases.

BTW, regardless of this change, it does seem that without a prefix, the same `http_file` could be registered multiple times with multiple `maven_install`. So it seems all `http_file` calls in `defs.bzl` should be wrapped inside `maybe`. I can also update this PR to include such change if desired.

What do you think?

tgeng:
gentile pint @shs96c

tgeng:
Any updates? @shs96c

PR #729, opened 2022-08-17
Enable pinned_coursier_fetch to use the Starlark version of aar_import

mauriciogg:
see https://github.com/bazelbuild/bazel/issues/5354

pinned_coursier_fetch was missed in original pr https://github.com/bazelbuild/rules_jvm_external/pull/526


shs96c (approved):
Good catch! Thank you the PR and fixing this!

PR #737, opened 2022-08-30
Create empty directories in merged zip files

shs96c:
Without this, some tools will not correctly unpack zip
files in some cases. This also restores behaviour from
rje where we would create these intermediate folders.

### DIFF
--- a/tests/com/jvm/external/jar/MergeJarsTest.java
+++ b/tests/com/jvm/external/jar/MergeJarsTest.java
@@ -20,12 +20,12 @@ import java.util.Set;
 import java.util.jar.Attributes;
 import java.util.jar.Manifest;
 import java.util.zip.ZipEntry;
+import java.util.zip.ZipFile;
 import java.util.zip.ZipInputStream;
 import java.util.zip.ZipOutputStream;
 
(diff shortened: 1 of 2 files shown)


tests/com/jvm/external/jar/MergeJarsTest.java
```diff
@@ -351,6 +351,35 @@ public void aMergedJarShouldHaveTheManifestAsTheFirstOrSecondEntry() throws IOEx
     }
   }
 
+  @Test
+  public void shouldAddMissingDirectories() throws IOException {
+    Path input = temp.newFile("example.jar").toPath();
+    createJar(input, ImmutableMap.of("foo/bar/baz.txt", "Hello, World!"));
```
jin, line 357: Can you add a test case with both `foo/bar/baz.txt` and `foo/bar/baz/qux.txt` to make sure the directory creation log don't get confused at `baz`?

jin (approved):
Thanks for the tests, as usual!

tests/com/jvm/external/jar/MergeJarsTest.java
shs96c: Done!

PR #738, opened 2022-08-30
Make DOS timestamp local time zone independent

### LINKED ISSUE #735: `AddJarManifestEntry` output differs based on local time zone
While investigating remote cache misses I found out that the hash of `header_<>.jar` files differed between what I had locally and what was on the remote agent, even though the JAR files from which these were generated was the same.

When comparing the JAR files using `zipinfo -v` I saw the modification times in the input JAR were:

```
file last modified on (DOS date/time):          2020 Jul 20 18:41:22
file last modified on (UT extra field modtime): 2020 Jul 21 03:41:22 local
file last modified on (UT extra field modtime): 2020 Jul 21 01:41:22 UTC
```

the local `header_` JAR has generated from the previous input has:

```
file last modified on (DOS date/time):          2020 Jul 21 03:41:22
file last modified on (UT extra field modtime): 2020 Jul 21 03:41:22 local
file last modified on (UT extra field modtime): 2020 Jul 21 01:41:22 UTC
```

and the remote `header_` JAR has:

```
file last modified on (DOS date/time):          2020 Jul 21 01:41:22
file last modified on (UT extra field modtime): 2020 Jul 21 03:41:22 local
file last modified on (UT extra field modtime): 2020 Jul 21 01:41:22 UTC
```

ZIP files encode the ancient MS-DOS time format, which represents "local" time and has no time zone information. An `extended timestamp` extra field (ID `0x5455`) can be used to encode the UTC time. You can ignore the lines ending with `local` in the examples above, that's not actually encoded, just the extra field printed based on my local time.

It looks like the MS-DOS time can differ based on the local time zone. I believe the issue happens here:

https://github.com/bazelbuild/rules_jvm_external/blob/8a9bcd940e441b9478ac82b8ac04ab85ed25758c/private/tools/java/rules/jvm/external/jar/AddJarManifestEntry.java#L209

(The unzipped contents of both JAR files are the same, including the modification time stamps. The MS-DOS time stamps are ignored by at least my `unzip` implementation.)

This issue is similar to e.g. https://github.com/bazelbuild/bazel/pull/10976.

lukaciko:
Adds tests replicating and fixes #735.

The `ZipEntry` APIs are pretty limited when it comes to querying and setting timestamps. The getters return the extended timestamp if that is present and MS-DOS timestamp otherwise, but there is no indication of which one was returned.

This change ensures that the manifest timestamp is always 2010-01-01, even if the system default changes during runtime. The rest of ZIP entries are copied over entirely so their modification timestamps (both extended and MS-DOS) remain the same. This also means that some other data is copied over which previously wasn't - flags, access time, creation time. I don't think that would be a problem, but I might be missing something.


tests/com/jvm/external/jar/AddJarManifestEntryTest.java
```diff
@@ -359,26 +391,50 @@ private Path createJarWithAutomaticModuleName(String name) throws IOException {
...
         zos.putNextEntry(ze);
         zos.write(entry.getValue().getBytes(UTF_8));
         zos.closeEntry();
       }
     }
   }
 
+  private void writeJarInTimezone(Path inputJar, Path outputJar, String timeZone) throws IOException {
+    final TimeZone previousTimeZone = TimeZone.getDefault();
+    TimeZone.setDefault(TimeZone.getTimeZone(timeZone));
```
lukaciko, line 430: Changing the global time zone doesn't feel nice. Happy to discuss alternative ways of testing for the issue.

shs96c (changes requested):
Thank you for the PR! This is really helpful. A couple of comments, but I hope it's not too much work.

private/tools/java/rules/jvm/external/jar/AddJarManifestEntry.java
```diff
@@ -38,11 +38,7 @@
  */
 public class AddJarManifestEntry {
 
-  private static final long DEFAULT_TIMESTAMP =
-      LocalDateTime.of(2010, 1, 1, 0, 0, 0)
-          .atZone(ZoneId.systemDefault())
-          .toInstant()
-          .toEpochMilli();
+  private static final LocalDateTime DEFAULT_TIMESTAMP = LocalDateTime.of(2010, 1, 1, 0, 0, 0);
```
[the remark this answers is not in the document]
lukaciko: @shs96c it makes sense to use the same timestamp, but I believe `StableZipEntry` has a similar problem to what this PR is solving. 

https://github.com/bazelbuild/rules_jvm_external/blob/96c12ccbc6094c0a362ee00d765210429fec2a92/private/tools/java/rules/jvm/external/zip/StableZipEntry.java#L44-L46

It took me a while to grok this. `ZipEntry.setTime()` / `setLastModifiedTime()` / `setTimeLocal` all set the modified time, which is internally represented by MS-DOS time and optionally the extended timestamp.

- `setTime()` will only set the MS-DOS time, adjusted for the default timezone (unless it's outside the MS-DOS time range)
- `setLastModifiedTime` will do the same, but also always set the extended timestamp
- `setTimeLocal` is not available before JDK 11

So the right thing to do for these "stable" entries is to make sure we're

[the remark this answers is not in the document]
shs96c: So it sounds like we should fix `StableZipEntry` in this PR

tests/com/jvm/external/jar/AddJarManifestEntryTest.java
shs96c: We have complete control over `AddJarManifestEntry`. We can make it a Proper Class rather than just a string of static methods, and then we can set the timezone via a setter in the tests, and just default it to something sensible in the default case.

If you're wary of doing that, I can prep a PR for it.

shs96c: Having said that, the output of your build should really be the same no matter which timezone you run it in.

private/tools/java/rules/jvm/external/jar/AddJarManifestEntry.java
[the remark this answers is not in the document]
lukaciko: Sure, happy to do that. I'll update the PR.

tests/com/jvm/external/jar/AddJarManifestEntryTest.java
lukaciko: The issue is that we don't have a way to inject the `TimeZone` or a provider for it into `ZipEntry`. It will use `TimeZone.getDefault()` under the hood.

Using a JUnit rule for setting and re-setting the time zone would be a bit better than what I have now, otherwise I'm out of ideas.

shs96c: I can prep a PR to provide some accessors and mutators. It's not a big change.

shs96c: PR ready: https://github.com/bazelbuild/rules_jvm_external/pull/744

The idea here is we can now add a setter for the `TimeZone` on `AddJarManifestEntry` for use in our tests. Feel free to either pull those changes into this PR, or wait until we merge it and rebase.

tests/com/jvm/external/jar/AddJarManifestEntryTest.java
```diff
@@ -359,26 +391,50 @@ private Path createJarWithAutomaticModuleName(String name) throws IOException {
...
+    AddJarManifestEntry.main(new String[]{
```
shs96c, line 431: Please use 
```java
AddJarManifestEntry adder = new AddJarManifestEntry()
adder.setTimeZone(timezone);
adder.addEntryToManifest(...)
```

We modified the class to make this possible without needing to reset the default timezone

private/tools/java/rules/jvm/external/zip/StableZipEntry.java
```diff
@@ -17,15 +17,15 @@
...
+import java.time.LocalDateTime;
+import java.time.ZoneId;
 import java.util.zip.ZipEntry;
 
 public class StableZipEntry extends ZipEntry {
 
   // File time is taken from the epoch (1970-01-01T00:00:00Z), but zip files
   // have a different epoch. Oof. Make something sensible up.
-  private static final long DOS_EPOCH = Instant.parse("1985-02-01T00:00:00.00Z").toEpochMilli();
+  private static final LocalDateTime DOS_EPOCH = LocalDateTime.of(1985,2,1,0,0);
```
shs96c, line 28: The idea with this was that it was UTC "1st February, 1985". I _think_ the constant we're looking for here is `long DOS_EPOCH = 473472000000` Using local timezones and timestamps will lead to the build being different depending on where you are in the world, and the date, when you run the build

private/tools/java/rules/jvm/external/zip/StableZipEntry.java
```diff
@@ -34,15 +34,16 @@ public class StableZipEntry extends ZipEntry {
   public StableZipEntry(String name) {
     super(name);
 
+    long timestamp = DOS_EPOCH.atZone(ZoneId.systemDefault()).toInstant().toEpochMilli();
```
shs96c, line 37: If we had the constant above, we'd not need this line.

shs96c (changes requested):

private/tools/java/rules/jvm/external/zip/StableZipEntry.java
lukaciko: I'm pretty sure that with the `ZipEntry` API we can't use a single timestamp, but need to get a timestamp that is adjusted with the system timezone so it gets "cancelled out" when we pass it into the `setTime()` method. The same approach is used in the Bazel repo here:

https://github.com/bazelbuild/bazel/blob/ef3f05852b3bf542903f091184da96b96354a13d/src/java_tools/buildjar/java/com/google/devtools/build/buildjar/jarhelper/JarHelper.java#L58-L62

and here:

https://github.com/bazelbuild/bazel/blob/ef3f05852b3bf542903f091184da96b96354a13d/src/tools/android/java/com/google/devtools/build/android/AarGeneratorAction.java#L73-L75

tests/com/jvm/external/jar/AddJarManifestEntryTest.java
lukaciko: I'm not sure how this would help. I can't pass or inject the time zone into `ZipEntry` class. Could you give me a more detailed example if you think that's the way to go?

lukaciko:
@shs96c sorry for taking some time with this PR. We've cherry picked it into our builds at Spotify and it has resolved the remote caching issues we had. Now we essentially get 100% remote cache hit rate on local builds when building a commit that was already built by the CI agent.

Thank you for the review. Please see my comments and let me know what you think.

tests/com/jvm/external/jar/AddJarManifestEntryTest.java
shs96c: We can add a `TimeZone` as an argument in the `StableZipEntry`. Given that this is only needed for testing, and given that we're running the tests using Bazel's test runner, the modification of the default timezone for the test should be okay

private/tools/java/rules/jvm/external/zip/StableZipEntry.java
shs96c: Thank you for the links. I also played around with the APIs and went and read the source of what the zip output stream is doing. I regret some life decisions, but this looks good to me. Thank you for clarifying :)

shs96c (approved):
LGTM. Thank you for your patience and thoughtful replies with this review!

PR #740, opened 2022-09-01
Pin rje to a hermetic Java 11 and use the `artifact` macro

shs96c:
We also enable strict dependencies, so we know what we
actually need to build, rather than picking up implicit
deps.


shs96c:
The JVM crashed in one of the test cases: I don't think that's the fault of this PR, since it passed in the other builds. Can we re-run that failing test, please?

jin (approved):

Issue #723, opened 2022-08-02
Can you limit a repository to only certain artifacts?


dhalperi:
We have one weird artifact that lives in its own repository. If we add its repository to the overall `repositories` field, then that shows up as a mirror URL for every artifact in the project, and Bazel queries it for every other artifact, and users get tons of 404 warnings.

Is there a way to provide repositories per-artifact while staying under the same connected external repository?

dhalperi:
Related (404 warnings): https://github.com/bazelbuild/rules_jvm_external/issues/349

Issue #349, opened 2019-12-31
ability to silence warnings when using multiple repositories


joprice:
When using multiple repositories, the assumption is that the repositories will be iterated in order and a dependency will be resolved from the first that succeeds. Failing to find a dependency in any given repository is not usually of interest if it is found in at least one. When dependency resolution fails, however, it is convenient to have this information to recreate each successive resolution attempt. Silencing these warnings in the positive case, if possible, would be helpful, but perhaps the call doesn't need to be made at all: I see multiple urls for "mirror_urls" in my pin file, where, for any given dependency, I expect only one to work. If the dependency was resolved once from a given mirror_url, could that information be stored, and that url be tried first?

An example of the error message I am referring to:

`WARNING: Download from ... failed: class com.google.devtools.build.lib.bazel.repository.downloader.UnrecoverableHttpException GET returned 404 Not Found`

jin:
Yes, this is pretty annoying. I'd expect this feature to be wired up to `repository_ctx.download`, and not a rules_jvm_external specific feature though.

Zetten:
Suppressing the warning should be handled in `repository_ctx.download`, I agree.

But some options to actually do some dependency _management_ are applicable here - essentially I think it's not a safe assumption that every artifact exists in every repository in a given maven_install target.

In my particular case I have a large set of artifacts from Maven Central, but also a few artifacts from another repository. I could create two maven_install targets, but it leads to a potential issue as the transitive deps of the second artifact set may not be pinned to the versions I request from Central, plus developers have to figure out which workspace contains which artifact.

Could rules_jvm_external avoid producing `mirror_urls` for entries in maven_install.json? This could be switched either globally or per-artifact, or best (and slowest) of all, based on testing each artifact in each repository and emitting only known-good URLs.

Perhaps the most elegant option would be to allow composing several maven_install targets into a single external workspace/maven_install.json - so we could specify repositories+artifacts in known-good combinations, but still benefit from a single `@maven//...` set of targets.

steeve:
How come the the old `aar_` and `maven_` repo rules didn't suffer from this problem?

jin:
> Could rules_jvm_external avoid producing mirror_urls for entries in maven_install.json? This could be switched either globally or per-artifact, or best (and slowest) of all, based on testing each artifact in each repository and emitting only known-good URLs.

[...]

PR #743, opened 2022-09-08
coursier: prefer the primary_url over mirrors

dhalperi:
This is for https://github.com/bazelbuild/rules_jvm_external/issues/349 and tangentially for https://github.com/bazelbuild/rules_jvm_external/issues/723.

Users are often confused by (expected) 404s when using multiple Maven repositories. These 404s can be safely ignored (Bazel's repository fetching logic gracefully tries all mirrors), but they are chatty and there is as of yet no way to silence them.

See also: https://github.com/bazelbuild/bazel/issues/13394

While we cannot solve the issue, we can reduce its impact by ordering the URLs: `http_file` is documented to try them in order:
https://bazel.build/rules/lib/repo/http#http_file-urls


coursier.bzl
```diff
@@ -1043,6 +1043,10 @@ def _coursier_fetch_impl(repository_ctx):
         primary_artifact_path = infer_artifact_path_from_primary_and_repos(primary_url, repository_urls)
 
         mirror_urls = [url + "/" + primary_artifact_path for url in repository_urls]
+        if primary_url in mirror_urls:
```
dhalperi, line 1046: It wasn't obvious to me that this is guaranteed to be the case, but I also didn't see how it couldn't be.

dhalperi (commented):
FYI, I am assuming there will be tests to update, and that will help me figure out a new test to write as well.

coursier.bzl
dhalperi: Ahh, auth makes it not true.

tests/custom_maven_install/regression_testing_install.json
```diff
@@ -2175,34 +2175,34 @@
                 "url": "https://repo1.maven.org/maven2/com/microsoft/azure/adal4j/1.6.0/adal4j-1.6.0.jar"
             },
             {
-                "coord": "com.nimbusds:lang-tag:1.6",
+                "coord": "com.nimbusds:lang-tag:1.7",
```
dhalperi, line 2178: FYI repinning just did this upgrade to 1.7. I can add another version set to fix if preferred.

jin (approved):

jin:
Nice!

PR #749, opened 2022-09-20
chore: upgrade to rules_js

alexeagle:
This is lazy and won't fetch any packages from npm when the loading phase visits scripts/BUILD


alexeagle:
Failing because rules_js requires Bazel 5. @jin do you have a policy for when r_jvm_e will drop support for Bazel 4?

jin:
We don't have concrete plans to drop backwards compat yet. Is this a necessary change?

shs96c:
We only include the JS rules so we can run `doctoc` when we update our documentation. I don't think we need to change the version of Bazel we're using to support that, since we so seldom run this part of the build.

Also, I'd prefer to hold on until Bazel 6 for an upgrade to `rules_js` since that's when it'll start working with remote builds.

alexeagle:
Not necessary at all, I was just driving by. Thanks!

shs96c:
It's appreciated :) I'm looking forward to picking up `rules_js` when we can!

PR #752, opened 2022-09-22
Move `outdated` deps to regular rje deps

shs96c:
This means that it's in the lock file (so no-one needs to re-resolve the dependencies), and also means that the tests which are failing as we attempt to move to the BCR pass.

As part of this work, we'll need another RJE release, so this also bumps the version in the module file.


shs96c:
The build that highlighted that this needed to be done: https://buildkite.com/bazel/bcr-presubmit/builds/513#01836498-147d-428d-b464-f04243cb3364

jin (approved):
thanks for the cleanup.

PR #754, opened 2022-09-22
Ensure module version used by example matches rje module version

shs96c:


shs96c:
The failing build which suggests this is necessary: https://buildkite.com/bazel/bcr-presubmit/builds/516#018364e6-39fe-4e27-b94d-7e5bbf605c92

jin (approved):

examples/bzlmod/MODULE.bazel
```diff
@@ -1,9 +1,9 @@
 module(
     name = "rules_jvm_external_examples_bzlmod",
-    version = "5.0.0",
+    version = "4.2.2",
```
jin, line 3: extract to constant?

shs96c: I didn't think a `MODULE.bazel` file could `load` an external file: I believe that they need to be entirely self-contained.

PR #757, opened 2022-09-22
Add rules license

davidkurkov:


shs96c:
I'm not sure of the motivation behind this PR. Could you please expand on why this is necessary and how it benefits users in the PR description?

In addition, you'll also need to modify the `MODULE.bazel` to ensure that `rules_jvm_external` continues to work with `bzlmod`.

aiuto:
This needs a design and conversation first.
The plan I am prototyping for splicing license and package information into imported code is substantially different.  Some background reading
[general requirments](https://docs.google.com/document/d/1oY53dQ0pOPEbEvIvQ3TvHcFKClkimlF9AtN89EPiVJU)
[brainstorming on the approach](https://docs.google.com/document/d/1PkSk1fHcLNYJX9fk8cEMKueMvrpt22QcVIwcTICCEIk)

Tl;DR; we don't want to implement this in each repository rule set. We want to be able to write a single repository transformer that injects license information into repos of any type. This will reduce the chance that each one does something slightly incompatible.

PR #761, opened 2022-09-30
Generate a stub so bzl_library works as expected

shs96c:


jin:
Could you give a bit more description / context behind this change?

shs96c:
Sure. When we freeze deps (as we do in [contrib_rules_jvm](https://github.com/bazel-contrib/rules_jvm/blob/main/tools/freeze-deps.py)), we keep a copy of the generated workspace in a zip. This is then unpacked and used as a dependency, thus avoiding the need to do a normal initialisation process of "first make rje available, then load its deps, then call `maven_install`, then load the generated workspace". This allows `maven_install` to be used for deps that take a regular "have a function for dependencies, and a function for setting up the ruleset" structure.

However, when we do that, stardoc will attempt to index through the "frozen" workspace to find the definitions from rje. This PR ensures that behaviour works as expected.

In other news, freezing deps is so useful, it's tempting to add it as a feature in rje.

jin:
Thanks for the clarification!

jin (approved):

PR #768, opened 2022-10-12
Add deploy_env to java_exports's javadoc class path

fmeum:
Javadoc may have to load classes only available on the deploy_env class path in order to generate the docs.


shs96c (approved):
Makes sense to me. Thank you for the PR! LGTM.

Issue #574, opened 2021-06-25
What is the intended workflow to pin dependencies in a repository requiring authentication?


hxtk:
In short, the issue is that configuring a repository with basic auth credentials and then pinning the dependency tree results in those basic auth credentials being embedded in `maven_install.json`, which is not a file intended to be edited manually. `netrc` authentication cannot be used until after the dependencies have already been pinned, so the intended sequence of operations to end up with a `maven_install.json` file which does not embed credentials and uses `netrc` for its authentication information is not clear to me.

The actual use case that I am trying to support is storing credentials in a way that allows me to commit all of my BAZEL files to version control without leaking those credentials or forcing all developers and service accounts to use the same credentials.

Below, I have detailed the steps that I personally took to work around this in case it helps to spot the point where I went wrong.

-----

I have a Maven repository which requires authentication. This is a supported use-case noted in the [usage section of the README](https://github.com/bazelbuild/rules_jvm_external#usage), and I was able to download the dependencies.

However, this had two issues: first of all, I did not want to hard-code credentials (especially not in a file that would be committed to version control) and second of all I wanted to pin my dependencies.

When I pinned the dependencies, I expected to be able to use the `netrc_lines` or a `netrc` file to store credentials by either substituting them from environment variables or reading them from a `.gitignore`'d netrc file that each developer could have on their own machine.

This didn't work, though, because pinning the dependencies embeds the exact URLs as built from the repository definitions, including the credentials. I used search-and-replace to remove any instances of the credentials from the generated `maven_install.json` file, but this caused the file to fail checksum. I removed the checksum and the result was something that worked, but throws warnings every time.

I don't think I'm using this feature as intended, but it is not clear to me how I am meant to use the `netrc` parameters to add credentials if I cannot use `netrc` for credentials until I have pinned the dependencies and the dependencies can only be pinned with their credentials already embedded.

Although it doesn't strictly match the issue title, I would also be happy to be informed of any other intended mechanism for allowing credentials to be set per-developer rather than embedding them in version-controlled files.

cheister:
If you don't put the credentials in the url and use a [properties file for coursier](https://get-coursier.io/docs/other-credentials.html#property-file) can you pin the artifacts without having the username and password in the `maven_install.json` file?

cheister:
[...]

PR #772, opened 2022-10-14
Strip username from pinned urls

brujoand:
As described in #574 there is a situation where we depend on a private repository which requires authentication, and that authentication is provided through coursier properties files. In this case it's not ideal that the username of the person who performed a pinning is present in both the file and url attributes stored in the pinned json file. This creates quite a bit of noise when reviewing changes.

This PR attempts to clean away this string without affecting other authentication flows. I'm fairly new to starlark and this project in particular so it's very possible that I've overlooked potential side effects. PTAL

### DIFF
--- a/coursier.bzl
+++ b/coursier.bzl
@@ -998,7 +998,7 @@ def _coursier_fetch_impl(repository_ctx):
             # This file comes from maven local, so handle it in two different ways depending if
             # dependency pinning is used:
             # a) If the repository is unpinned, we keep the file as is, but clear the url to skip it
-            # b) Otherwise, we clear the url and also simlink the file from the mavel local directory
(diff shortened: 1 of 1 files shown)


shs96c:
Does this work when using an unpinned maven repo that requires authentication?

brujoand:
I ran some more test and I updated my WORKSPACE so that I had
- A repo which was unpinned targeting "https://repo1.maven.org/maven2"
- A repo which was unpinned targeting a private repository
- A repo which was pinned targeting "https://repo1.maven.org/maven2"
- A repo which was pinned targeting a private repository

All repos could be fetched, even after a `bazel clean --expunge`.

shs96c (approved):
Thank you for checking those cases. Let's land this and see what happens :)

Thank you very much for both the PR and your patience!

PR #776, opened 2022-10-19
Mention that `java_binary` isn't supported in `java_export` `deploy_env`

fmeum:
Fixing this doesn't seem possible as `JavaRuntimeClasspathProvider` isn't visible from Java.

Also update the Bazel docs URLs.


shs96c (approved):
Thank you!

PR #777, opened 2022-10-19
Fail on `java_binary` in `deploy_env` of `java_export`

fmeum:
`java_export` silently ignores `java_binary` targets in `deploy_env` since their `JavaInfo` has an empty runtime classpath. Since this is the direct opposite of how the attribute of the same name on `java_binary` works, which only accepts `java_binary` targets, this is quite error-prone and warrants a strong warning/failure.

Follow up to https://github.com/bazelbuild/rules_jvm_external/pull/776

### DIFF
--- a/private/rules/maven_project_jar.bzl
+++ b/private/rules/maven_project_jar.bzl
@@ -34,6 +34,22 @@ def _maven_project_jar_impl(ctx):
         bin_jar,
     )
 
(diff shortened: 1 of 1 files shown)


fmeum:
If you are worried about backwards compatibility, this could also be made `print`.

shs96c (approved):
LGTM. Thank you!

PR #778, opened 2022-10-21
feature: add coursier --credentials from home netrc

pcj:
This is revisitation of https://github.com/bazelbuild/rules_jvm_external/pull/484.

### DIFF
--- a/coursier.bzl
+++ b/coursier.bzl
@@ -98,8 +98,6 @@ sh_binary(
 )
 """
 
-COURSIER_PROPERTIES_FILENAME = "credentials.properties"
-
 def _is_verbose(repository_ctx):
     return bool(repository_ctx.os.environ.get("RJE_VERBOSE"))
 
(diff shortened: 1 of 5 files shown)


coursier.bzl
```diff
@@ -798,6 +798,8 @@ def make_coursier_dep_tree(
         cmd.extend(["--repository", repository["repo_url"]])
         if "credentials" in repository:
             cmd.extend(["--credentials", utils.repo_credentials(repository)])
+    for credential in utils.netrc_credentials(get_home_netrc_contents(repository_ctx)):
```
shs96c, line 801: We probably only want to set this if we haven't already set the credentials (see the line immediately above this one!), and if users ask for us to add the credentials. There's an open issue that `coursier` leaves credentials in URLs, so I'm loathe to pull things from `.netrc` only to expose them accidentally in a PR. Could you please make this behaviour conditional and opt-in?

pcj: Yes, we can add a flag for it.

shs96c (approved):
LGTM. Thank you for doing this!

PR #721, opened 2022-07-27
Quote java arguments written to arguments file.

davidmandle:
This pull request is intended to fix the `java` Command-Line Argument File creation so that arguments are quoted (i.e. they are wrapped in quotes, and any quotes inside an argument are escaped). See #720 for context. In particular, this allows arguments with white space to be treated as single arguments. In particular this fixes the passing of `--credentials` arguments to coursier, which are expected to have a format like `--credentials "oss.sonatype.org alex-sonatype:aLeXsOnAtYpE"`

* See https://docs.oracle.com/en/java/javase/18/docs/specs/man/java.html#java-command-line-argument-files for the `java` Command-Line Argument File specs.
* See https://get-coursier.io/docs/other-credentials.html#inline for the coursier inline credentials format.

Issue #581, opened 2021-07-30
Broken @maven//:outdated with multiple hosts in NO_PROXY


ejona86:
I tried running `@maven//:outdated` and it failed to run:

```
.../7300081f768a13eaa40dfca3d4b14cce/execroot/io_grpc_grpc_java/bazel-out/k8-fastbuild/bin/external/maven/outdated: line 11: 127.0.0.0/8: No such file or directory
.../7300081f768a13eaa40dfca3d4b14cce/execroot/io_grpc_grpc_java/bazel-out/k8-fastbuild/bin/external/maven/outdated: line 11: ::1: command not found
Usage: java [-options] class [args...]
           (to execute a class)
   or  java [-options] -jar jarfile [args...]
           (to execute a jar file)
where options include:
    -d32	  use a 32-bit data model if available
    -d64	  use a 64-bit data model if available
...
```

Looking at the `external/maven/outdated` file I see:
```
java -Dhttp.nonProxyHosts=localhost|127.0.0.0/8|::1 -jar $outdated_jar_path external/maven/outdated.artifacts external/maven/outdated.repositories
```

Those `|`s are being interpreted as pipes and are not being passed to java. Single-quoting the string or escaping the bars with backslash is missing. It seems the bug is within this transform:
https://github.com/bazelbuild/rules_jvm_external/blob/c0a8a3ea3fed24a66eade21015249ed5ed99bf65/private/proxy.bzl#L74

My environment has both of these variables set:
```
NO_PROXY="localhost,127.0.0.0/8,::1"
no_proxy="localhost,127.0.0.0/8,::1"
```

jin:
cc @cheister

PR #784, opened 2022-11-08
fix NO_PROXY handling under Java 9+

### LINKED ISSUE #780: NO_PROXY doesn't work on Java 9+
I think #589 introduced a regression in the handling of the `NO_PROXY` environment variable, because the generated argsfile for Coursier will contain a line like `"-Dhttp.nonProxyHosts='artifactory.mydomain.com'"`, where the JVM interprets single quotes as part of the hostname to be excluded from proxy args.

Since `repository_ctx.execute` takes an array of args and doesn't involve the shell, would it be safe to drop the single quotes from private/proxy.bzl?

```diff
--- a/private/proxy.bzl
+++ b/private/proxy.bzl
@@ -73,6 +73,6 @@ def get_java_proxy_args(http_proxy, https_proxy, no_proxy):
     if no_proxy:
         if no_proxy.startswith("."):
             no_proxy = "*" + no_proxy
-        proxy_args.append("-Dhttp.nonProxyHosts='%s'" % no_proxy.replace(",", "|").replace("|.", "|*."))
+        proxy_args.append("-Dhttp.nonProxyHosts=%s" % no_proxy.replace(",", "|").replace("|.", "|*."))

     return proxy_args
```

beasleyr-vmw:
Remove single-quoting of the `-Dhttp.nonProxyHosts` argument as this arg may be spilled over to an argsfile, where quoting is considered by the JVM as part of the host expression.  (I expect the single-quoting to be unnecessary in the pre-Java9+ path, as my understanding of `rctx.execute` is that Bazel _doesn't_ use the shell, so special chars like pipes, etc. are harmless.)

Resolves #780.

### DIFF
--- a/coursier.bzl
+++ b/coursier.bzl
@@ -117,6 +117,13 @@ def _is_file(repository_ctx, path):
 def _is_directory(repository_ctx, path):
     return repository_ctx.which("test") and repository_ctx.execute(["test", "-d", path]).return_code == 0
 
+def _shell_quote(s):
+    # Lifted from
+    #   https://github.com/bazelbuild/bazel-skylib/blob/6a17363a3c27dde70ab5002ad9f2e29aff1e1f4b/lib/shell.bzl#L49
+    # because this file cannot load symbols from bazel_skylib since commit
+    # 47505f644299aa2483d0df06c2bb2c7aa10d26d4.
+    return "'" + s.replace("'", "'\\''") + "'"
+
 def _execute(repository_ctx, cmd, timeout = 600, environment = {}, progress_message = None):
     if progress_message:
         repository_ctx.report_progress(progress_message)
@@ -426,7 +433,7 @@ def _add_outdated_files(repository_ctx, artifacts, repositories):
         repository_ctx.attr._outdated,
         {
             "{repository_name}": repository_ctx.name,
-            "{proxy_opts}": " ".join(_get_java_proxy_args(repository_ctx)),
+            "{proxy_opts}": " ".join([_shell_quote(arg) for arg in _get_java_proxy_args(repository_ctx)]),
         },
         executable = True,
     )
--- a/private/proxy.bzl
+++ b/private/proxy.bzl
@@ -73,6 +73,6 @@ def get_java_proxy_args(http_proxy, https_proxy, no_proxy):
     if no_proxy:
         if no_proxy.startswith("."):
             no_proxy = "*" + no_proxy
-        proxy_args.append("-Dhttp.nonProxyHosts='%s'" % no_proxy.replace(",", "|").replace("|.", "|*."))
+        proxy_args.append("-Dhttp.nonProxyHosts=%s" % no_proxy.replace(",", "|").replace("|.", "|*."))
 
     return proxy_args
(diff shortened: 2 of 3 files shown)


beasleyr-vmw:
Just saw https://github.com/bazelbuild/rules_jvm_external/issues/581, so I'll need to revisit this.

cheister:
Thanks for the PR. Is it possible we only need to quote if there is a `|` in the string?

beasleyr-vmw:
> Thanks for the PR. Is it possible we only need to quote if there is a `|` in the string?

Thanks for the review. :) So, `NO_PROXY` _could_ contain asterisks or other special characters (brackets for IPv6 addresses?), and IMO it's simpler to unconditionally quote all args. IIUC the args are dumped to a shell script that users won't see, so isn't as if we need to think about error messages. Please lmk what you think.

cheister:
> > Thanks for the PR. Is it possible we only need to quote if there is a `|` in the string?
> 
> Thanks for the review. :) So, `NO_PROXY` _could_ contain asterisks or other special characters (brackets for IPv6 addresses?), and IMO it's simpler to unconditionally quote all args. IIUC the args are dumped to a shell script that users won't see, so isn't as if we need to think about error messages. Please lmk what you think.

Seems like this will break the case with |'s in the string again?  https://github.com/bazelbuild/rules_jvm_external/issues/581

beasleyr-vmw:
> > > Thanks for the PR. Is it possible we only need to quote if there is a `|` in the string?
> > 
> > 
> > Thanks for the review. :) So, `NO_PROXY` _could_ contain asterisks or other special characters (brackets for IPv6 addresses?), and IMO it's simpler to unconditionally quote all args. IIUC the args are dumped to a shell script that users won't see, so isn't as if we need to think about error messages. Please lmk what you think.
> 
> Seems like this will break the case with |'s in the string again? #581

How? In https://github.com/bazelbuild/rules_jvm_external/pull/784/files#diff-680ed1b34ec806ef9fcf63148507fe91922eecad5fde47db94ee8efe01f7295fR436 we're shell-quoting all of the args returned by `_get_java_proxy_args(rctx)`, and so `-Dhttp.nonProxyHosts=foo|bar` will become `'-Dhttp.nonProxyHosts=foo|bar'`. And so, after template expansion, `private/outdated.sh`'s
```bash
java {proxy_opts} -jar "$outdated_jar_path" "$artifacts" "$repositories"
```
will become
```bash
java '-Dhttp.nonProxyHosts=foo|bar' -jar "$outdated_jar_path" "$artifacts" "$repositories"
```
such that `java`'s `argv[1]` will be `-Dhttp.nonProxyHosts=foo|bar`.

All other evaluations of `cmd` are done by `repository_ctx.execute`, and that doesn't involve the shell, so pipes are nbd.

It's always possible that I'm missing something, but I _think_ I'm in the clear. If not, please lmk. Thanks!

cheister:
Sorry, the part that I was missing was https://github.com/bazelbuild/rules_jvm_external/pull/721 which started quoting everything in the argsfile and therefore caused the single quotes in the argsfile to come through in the properties value.  

When the single quotes were added in https://github.com/bazelbuild/rules_jvm_external/pull/588 the argsfile did not have quoted strings. 

After that change, this one makes more sense.

cheister (approved):

PR #790, opened 2022-11-16
fix: provide our own remove_prefix implementation

richardstephens:
`.removeprefix` was only added to starlark on bazel 5, so this fails on bazel 4.


jin (approved):
We do have support for Bazel 4+ so this looks reasonable to me.

PR #791, opened 2022-11-17
Add instructions for IPv6 support in rules_jvm_external.

jin:
These are documented in Bazel's docs, but we should also document that here as well.


shs96c (approved):
LGTM

PR #792, opened 2022-11-17
Bump coursier to v2.1.0-RC2

jin:
Also mirrored to https://mirror.bazel.build/coursier_cli/coursier_cli_v2.1.0-RC2.jar


shs96c (approved):
LGTM, though it's interesting that the lock files have changed.

jin:
I downloaded zookeeper-3.4.13 and verified that the new lock files reflects the deps in its pom file. Given that there's nothing stopping Maven owners from changing POM files for released artifacts, it'd explain changed lock files like these.

cheister:
LGTM, thanks for the update

shs96c:
The lock file is generated by serialising the `artifacts` from `maven_install` (in the common case, that's just a regular GAV tuple) and passing them to `coursier`. In this case, it looks like `io.opencensus:opencensus-api:0.24.0` did something interesting: it went from including `io.grpc:grpc-context:1.33.1` as a dependency to adding it to the `exclusions`.

PR #796, opened 2022-11-24
rules_jvm_external_deps: add possibility to override params used for maven install

gergelyfabian:
…maven_install

### DIFF
--- a/repositories.bzl
+++ b/repositories.bzl
@@ -9,7 +9,7 @@ _DEFAULT_REPOSITORIES = [
     "https://maven.google.com",
 ]
 
-def rules_jvm_external_deps(repositories = _DEFAULT_REPOSITORIES):
+def rules_jvm_external_deps(repositories = _DEFAULT_REPOSITORIES, fail_if_repin_required = True, strict_visibility = True):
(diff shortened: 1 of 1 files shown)


shs96c:
I'm wondering what the motivation for this PR is. The deps linked to are internal to `rules_jvm_external`, and we don't expect people to be relying or modifying them. I'm not sure why we'd want to allow people to change these values.

The two cases I can think of:

1. Someone is patching `rules_jvm_external`, at which point their patch can also modify the lock file.
2. Someone is attempting to rely on dependencies brought in transitively from these deps, which isn't something we want to encourage.

However, it's entirely possible I've missed something obvious :)

gergelyfabian:
When I did not override this parameter my project simply had an error (I believe it may have been before a rules_jvm_external update). I'm checking whether I can remove this from my project.

gergelyfabian:
It seems it's not anymore necessary. My project works without it. Thank you for the due diligence!

shs96c:
Thank you for making the effort to put this together! I'm glad to hear that your project is working now. Happy hacking :)

PR #801, opened 2022-12-03
Prefer artifacts with dependencies when de-duplicating coursier dependencies

cheister:
When coursier returns duplicate dependencies some of the duplicates will have dependencies and some won't.  When we de-duplicate the coursier dependencies we should prefer the ones that have dependencies as these are more likely to be the ones we want.


shs96c (approved):
LGTM

Issue #800, opened 2022-12-02
rules_jvm_external broken with Bazel@HEAD in Bazel downstream pipeline


meteorcloudy:
Context: https://github.com/bazelbuild/bazel/issues/16746

Some tests in rules_jvm_external seem to parse the output log of Bazel and https://github.com/bazelbuild/bazel/commit/6e945e8ab422ca0a195d09899cce61ba6c0cfa59 broke the tests, please help resolve this issue so that we can make sure rules_jvm_external is green in [Bazel downstream pipeline](https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/2760) and keep detecting breaking changes/Bazel bugs that break rules_jvm_external

cheister:
I think https://github.com/bazelbuild/rules_jvm_external/pull/803 should fix this issue, but I didn't know of an easy way to test it with bazel@head.

meteorcloudy:
You can test with Bazel@HEAD by setting `bazel: last_green` in https://github.com/bazelbuild/rules_jvm_external/blob/master/.bazelci/presubmit.yml in your PR.

For example, in rules_apple's [presubmit.yml file](https://github.com/bazelbuild/rules_apple/blob/master/.bazelci/presubmit.yml#L20)

Check Bazel CI documentation: https://github.com/bazelbuild/continuous-integration/tree/master/buildkite#using-a-specific-version-of-bazel

Of course, you can revert the change to keep testing with the latest Bazel release.

PR #803, opened 2022-12-05
For m2local integration tests send all output to test log and clear cache before each test to prevent caching of print output

cheister:
Attempting to address https://github.com/bazelbuild/rules_jvm_external/issues/800 for Bazel HEAD which has added caching for print statements.

It also sends all output from the tests to $TEST_LOG so it does go to stdout when the tests are run.

### DIFF
--- a/tests/bazel_run_tests.sh
+++ b/tests/bazel_run_tests.sh
@@ -16,61 +16,60 @@ function test_duplicate_version_warning() {
 }
 
 function test_m2local_testing_found_local_artifact_through_pin() {
-  m2local_dir="~/.m2/repository"
+  m2local_dir="${HOME}/.m2/repository"
+  jar_dir="${m2local_dir}/com/example/kt/1.0.0"
+  rm -rf ${jar_dir}
   mkdir -p ${m2local_dir}
+  bazel clean --expunge >> "$TEST_LOG" 2>&1 # for https://github.com/bazelbuild/rules_jvm_external/issues/800
   # Publish a maven artifact locally - com.example.kt:1.0.0
-  bazel run --define maven_repo="file://${HOME}/.m2/repository" //tests/integration/kt_jvm_export:test.publish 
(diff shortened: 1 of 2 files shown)


tests/bazel_run_tests.sh
```diff
@@ -16,61 +16,56 @@ function test_duplicate_version_warning() {
 }
 
 function test_m2local_testing_found_local_artifact_through_pin() {
-  m2local_dir="~/.m2/repository"
+  m2local_dir="${HOME}/.m2/repository"
```
shs96c, line 19: Perhaps this should be the result of `mktemp` instead of a user's home directory. I can see this having hilarious unforeseen consequences.

tests/bazel_run_tests.sh
```diff
@@ -16,61 +16,56 @@ function test_duplicate_version_warning() {
...
   mkdir -p ${m2local_dir}
+  bazel clean --expunge >> "$TEST_LOG" 2>&1 # for https://github.com/bazelbuild/rules_jvm_external/issues/800
   # Publish a maven artifact locally - com.example.kt:1.0.0
-  bazel run --define maven_repo="file://${HOME}/.m2/repository" //tests/integration/kt_jvm_export:test.publish 
+  bazel run --define maven_repo="file://${m2local_dir}" //tests/integration/kt_jvm_export:test.publish >> "$TEST_LOG" 2>&1
   bazel run @m2local_testing//:pin >> "$TEST_LOG" 2>&1
   rm -f m2local_testing_install.json
-  rm -rf ~/.m2/repository
+  rm -rf ${m2local_dir}
```
shs96c, line 26: Because we delete it at the end of the test

tests/bazel_run_tests.sh
cheister: Ok, it was working locally because I still had the kt example lib in my `~/.m2/repository` directory.  I changed it so it still goes to local but it only removes the directory for the kt example jar instead.

PR #804, opened 2022-12-06
Run 'buildifer -r .' to format all Bazel files

cheister:
Just keeping buildifier happy


shs96c (approved):
Nice :) Thank you for doing that!

PR #794, opened 2022-11-21
maven version expanded

knarukulla:
Issue: https://github.com/bazelbuild/rules_jvm_external/issues/618

Solution:

Expand maven version using bazel variables to publishing maven artifact to repo.
bazel run --stamp \
  --define "maven_repo=<> \
  --define "maven_user=<>" \
  --define "maven_password=<>" \
  --define "maven_version=<>" \
  //src/main/java/com/xxx/example:<targt>

PR #805, opened 2022-12-10
Extra maven artifacts support for java_export rule

krisnaru:
Issue:
Java_export needs to upload multiple artifacts as part of single rule. Artifacts could be war/tar files and it needs to support classifier.

Example:
BUILD:
java_export(
    name = "example-export",
    extra_artifacts = {
        "release": "//src/main/java/com/github/bazelbuild/rulesjvmexternal/example/export:tar",
    },
    maven_coordinates = "com.example:bazel-example:0.0.1",
    runtime_deps = [
        "//src/main/java/com/github/bazelbuild/rulesjvmexternal/example/export",
    ],
)

bazel run -define "maven_repo=file://$HOME/.m2/repository"  :example-export.publish


krisnaru:
@cheister  I have made one more commit towards templating maven coordinates. Could you please check?

examples/java-export/src/main/java/com/github/bazelbuild/rulesjvmexternal/example/export/BUILD
```diff
@@ -11,3 +12,12 @@ java_library(
         artifact("com.google.guava:guava"),
     ],
 )
+
+pkg_tar(
+    name = "tar",
+    srcs = glob(["*"]),
+    extension = "tar.gz2",
```
thirtyseven, line 19: I think you meant `tar.gz` here?

private/tools/java/rules/jvm/external/maven/MavenPublisher.java
```diff
@@ -104,6 +106,10 @@ public static void main(String[] args) throws IOException, InterruptedException,
         futures.add(upload(repo, credentials, coords, "-javadoc.jar", docJar, gpgSign));
       }
 
+      if(extraClassifiers != null && extraArtifacts != null) {
+        uploadextraArtifacts(repo, credentials, coords, extraClassifiers, extraArtifacts);
```
thirtyseven, line 110: The return value of this should probably be added to `futures`?

private/tools/java/rules/jvm/external/maven/MavenPublisher.java
```diff
@@ -165,6 +171,37 @@ private static CompletableFuture<Void> upload(
     return CompletableFuture.allOf(uploads.toArray(new CompletableFuture<?>[0]));
   }
 
+  private static CompletableFuture<Void> uploadextraArtifacts(
```
thirtyseven, line 174: `uploadExtraArtifacts`?

private/tools/java/rules/jvm/external/maven/MavenPublisher.java
```diff
@@ -165,6 +171,37 @@ private static CompletableFuture<Void> upload(
...
+      String repo,
+      Credentials credentials,
+      Coordinates coords,
+      String extraClassifiers,
+      String extraArtifacts
+    ){
+    List<CompletableFuture<?>> uploads = new ArrayList<>();
+    String[] extraClassifiersArr = extraClassifiers.split(",");
+    String[] extraArtifactsArr = extraArtifacts.split(",");
+    System.out.println("Working Directory = " + System.getProperty("user.dir"));
```
thirtyseven, line 184: Maybe should remove these debug print statements

thirtyseven (changes requested):

private/rules/maven_bom_fragment.bzl
```diff
@@ -12,9 +12,11 @@ MavenBomFragmentInfo = provider(
 )
 
 def _maven_bom_fragment_impl(ctx):
+    # Expand maven coordinates for any variables to be replaced.
+    coordinates = ctx.attr.maven_coordinates.format(**ctx.var)
```
thirtyseven, line 16: This line is actually fixing a bug introduced by https://github.com/bazelbuild/rules_jvm_external/pull/794 where the `<version>` in the generated pom isn't template expanded, might be better to move this to a separate PR since it's unrelated to this new feature

krisnaru: sure will move to different PR

krisnaru:
@cheister could you please review and approve if you are fine with changes

krisnaru:
@cheister could you please review and approve if you are fine with changes

private/rules/maven_publish.bzl
```diff
@@ -5,13 +5,15 @@ MavenPublishInfo = provider(
...
+        "extra_classifiers": "Extra classifiers for Artifacts",
+        "extra_artifacts": "Extra artifacts such as tars, wars",
     },
 )
 
 _TEMPLATE = """#!/usr/bin/env bash
 
 echo "Uploading {coordinates} to {maven_repo}"
-{uploader} {maven_repo} {gpg_sign} {user} {password} {coordinates} {pom} {artifact_jar} {source_jar} {javadoc}
+{uploader} {maven_repo} {gpg_sign} {user} {password} {coordinates} {pom} {artifact_jar} {source_jar} {javadoc} {extra_classifiers} {extra_artifacts}
```
shs96c, line 16: Can we pair up classifier and artifact? It'll make scanning the command line when debugging a little simpler. Something like `classifier=path` would be easy to iterate over, and it's really clear what would go with what. It would also mean that we could use a `for` loop to read the additional artifacts.

private/rules/maven_publish.bzl
```diff
@@ -104,6 +109,8 @@ When signing with GPG, the current default key is used.
         "source_jar": attr.label(
             allow_single_file = True,
         ),
+        "extra_classifiers": attr.string_list(),
```
shs96c, line 112: If you follow the suggestion above, this either becomes a `string_list`, or (if you'd like to wait as long as possible before preparing the command line) a `string_dict`

private/tools/java/rules/jvm/external/maven/MavenPublisher.java
```diff
@@ -89,6 +89,8 @@ public static void main(String[] args)
     Path binJar = getPathIfSet(args[6]);
     Path srcJar = getPathIfSet(args[7]);
     Path docJar = getPathIfSet(args[8]);
+    String extraClassifiers = args.length > 9 ? args[9] : null;
```
shs96c, line 92: Maybe default to the empty string or add a dep on guava and use `Strings.isNullOrEmpty` later on.

shs96c (commented):
My apologies it has taken so long to review this. A couple of comments, but I think this is looking good.

private/tools/java/rules/jvm/external/maven/MavenPublisher.java
krisnaru: will do

private/rules/maven_publish.bzl
krisnaru: I would need string_label_dict which is not supported yet on attr. https://stackoverflow.com/questions/43396925/emulate-string-to-label-dict

private/rules/maven_publish.bzl
krisnaru: i can make this change but this is not inline with attr change

private/rules/maven_publish.bzl
thirtyseven: Maybe can flip the keys/values and use [label_keyed_string_dict](https://bazel.build/rules/lib/attr#label_keyed_string_dict)?

private/tools/java/com/github/bazelbuild/rules_jvm_external/maven/MavenPublisher.java
```diff
@@ -106,6 +111,17 @@ public static void main(String[] args)
         futures.add(upload(repo, credentials, coords, "-javadoc.jar", docJar, gpgSign));
       }
 
+      if(args.length > 9) {
+        String ext = com.google.common.io.Files.getFileExtension(args[9]);
```
shs96c, line 115: Won't this value need to be calculated per tuple, rather than just on the final one? eg `war=foo.jar,tar=foo.tar` would result in `ext` being set to `tar` for both the `war` and `tar` classifiers.

private/tools/java/com/github/bazelbuild/rules_jvm_external/maven/MavenPublisher.java
```diff
@@ -106,6 +111,17 @@ public static void main(String[] args)
...
+        List<String> extraArtifactTuples = Splitter.onPattern(",").splitToList(args[9]);
+        for(String artifactTuple: extraArtifactTuples) {
+          String[] splits = artifactTuple.split("=");
+          String classifier = splits[0];
+          Path artifact = getPathIfSet(splits[1]);
```
shs96c, line 120: The `Path` should exist. If it doesn't, we've got real problems.... `Paths.get` is fine here.

shs96c (changes requested):
This is almost there. I've spotted one issue, but once that's fixed, let's merge this.

shs96c:
@krisnaru, can I help you in any way to get this one over the line? It'd be lovely to have this in the tree....

thirtyseven:
I have continued the work in this PR in #926 .

cjohnstoniv:
Is there any intention/support for extra classifier artifacts being in the BOM too? Would be useful for things like test-jars that sometime get published and depended on.

PR #806, opened 2022-12-13
Add a tool to format source in the repo, and run it.

shs96c:
Best reviewed by looking at the first diff, because the second is the result of running that, and has no manual changes.

For simplicity, I've chosen the default Google java format for Java code, and `buildifier` for the build files. I'm happy to pick an alternative java formatter, but the advantage with google java format is that there's a plugin for IJ, so (selfishly!) I won't forget to run it.

A sensible next step would be to hook this into the CI builds so we don't forget to format things, but I think that can wait for a later PR.


cheister:
Looks good to me.  There's always some things a Java auto-formatter does that looks terrible but easier to have a standard that everything conforms to.

It looks like the gazelle rules don't work with Bazel < 5.1.0
```
Error in fail:
--
  | Current Bazel version is 4.0.0, expected at least 5.1.0
  | ERROR: error loading package '': Encountered error while reading extension file 'deps.bzl': no such package '@bazel_gazelle//': error loading package 'external': Could not load //external package

```

WORKSPACE
```diff
@@ -79,6 +79,60 @@ yarn_install(
     yarn_lock = "//:yarn.lock",
 )
 
+# Required for buildifier (`//scripts:buildifier`)
```
Bencodes, line 82: Using https://github.com/keith/buildifier-prebuilt would remove the need for the additional go dependencies.

shs96c: It's pretty easy to avoid needing to build from source. I may just do that.

shs96c: Done!

cheister (approved):

PR #807, opened 2022-12-13
Introduce a new maven dependency resolver

shs96c:
This new resolver aims to have two advantages over the current scheme:

1. A more compact lock file format that is easier to review when changes are made in PRs
2. Support for Maven BOMs

While the first of these could be addressed easily enough with `coursier`, the latter has been an open issue for some time.

This PR introduces a new resolver which has two resolution mechanisms. One is based on Maven's resolver, and the other on Gradle's.

Work remaining:

- [x] Implement Maven resolver
- [x] Implement Gradle resolver
- [x] Allow current lock file to be consumed
- [x] Allow global exclusions
- [x] Allow exclusions at the dependency level
- [ ] Support BOMs
- [x] Ensure repinning works
- [x] Ensure the `outdated` target works
- [x] Support for the "unsafe shared cache"
- [x] [Compat repositories](https://github.com/bazelbuild/rules_jvm_external#repository-aliases)
- [x] Allow users to use `m2Local` as a shorthand for `~/.m2/repository`
- [x] Jetify support
- [x] Allow users to define default visibility of targets in generated build files
- [x] Dynamic usage, without a lock file
- [ ] Warn users about dependency convergence
- [ ] Ensure existing tests pass
- [ ] Gradle resolver should fetch `compile` scoped dependencies
- [ ] Add more tests!


thirtyseven:
@shs96c this looks awesome, I've scanned the changes and it looks like this will also address https://github.com/bazelbuild/rules_jvm_external/issues/445 since the new resolver is built as a java_binary and thus will use the toolchain resolved JRE instead of shelling out to `java`, can you confirm?

PR #813, opened 2022-12-28
bzlmod: Add `dev_dependency` functionality to tags

fmeum:
This attribute mimics the attribute of the same name on `bazel_dep` and makes it possible to exclude artifacts from dependency resolution when the module that specified them isn't the root module.

### DIFF
--- a/private/extensions/maven.bzl
+++ b/private/extensions/maven.bzl
@@ -21,6 +21,7 @@ _artifact = tag_class(
         "testonly": attr.bool(),
         "exclusions": attr.string_list(doc = "Maven artifact tuples, in `artifactId:groupId` format", allow_empty = True),
         "repositories": attr.string_list(default = DEFAULT_REPOSITORIES),
+        "dev_dependency": attr.bool(),
     },
 )
 
@@ -72,6 +73,7 @@ _install = tag_class(
         "fail_if_repin_required": attr.bool(doc = "Whether to fail the build if the maven_artifact inputs have changed but the lock file has not been repinned.", default = False),
         "lock_file": attr.label(),
         "repositories": attr.string_list(default = DEFAULT_REPOSITORIES),
+        "dev_dependencies": attr.bool(),
 
         # When using an unpinned repo
         "excluded_artifacts": attr.string_list(doc = "Artifacts to exclude, in `artifactId:groupId` format. Only used on unpinned installs", default = []),  # list of artifacts to exclude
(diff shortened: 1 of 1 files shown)


fmeum:
@shs96c This might be interesting for you.

fmeum:
CI failed with:

```


#
--
  | # A fatal error has been detected by the Java Runtime Environment:
  | #
  | #  SIGBUS (0x7) at pc=0x00007f0c91698807, pid=2, tid=3
  | #
  | # JRE version:  (11.0.12+7) (build )
  | # Java VM: OpenJDK 64-Bit Server VM (11.0.12+7-LTS, mixed mode, sharing, tiered, compressed oops, g1 gc, linux-amd64)
  | # Problematic frame:
  | # V  [libjvm.so+0xcd8807]  PerfMemory::alloc(unsigned long)+0x57
  | #
  | # Core dump will be written. Default location: Core dumps may be processed with "/usr/share/apport/apport -p%p -s%s -c%c -d%d -P%P -u%u -g%g -- %E" (or dumping to /var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/ec321eb2cc2d0f8f91b676b6d4c66c29/sandbox/linux-sandbox/120/execroot/rules_jvm_external/core.2)
  | #
  | # An error report file with more information is saved as:
  | # /var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/ec321eb2cc2d0f8f91b676b6d4c66c29/sandbox/linux-sandbox/120/execroot/rules_jvm_external/hs_err_pid2.log
```

These kind of failures are probably fixed by adding `--sandbox_tmpfs_path=/tmp`.

fmeum:
Converted to draft as I noticed that this needs `dev_dependency` on `use_repo` to be useful.

shs96c:
I don't think this is how dev dependencies work. In order to allow things to be dev dependencies, you need to do something like:

```starlark
bazel_dep(name = "rules_jvm_external", version = "6.0.0")

dev_maven = use_extension("@rules_jvm_external//:extensions.bzl", "maven")

dev_maven.artifact(...)
dev_maven.install(...)
```

shs96c:
The set of artifacts being considered for dependency resolution are used to construct the hashes within the lock file: altering that means that the check will fail, causing builds to break (if `fail_if_repin_required` is set) or issue warnings (if not) Excluding based on whether or not the module is a root module or not will cause some surprising behaviours.

fmeum:
I read the implementation of the Bazel-provided `dev_dependency` feature and thought that it wouldn't be possible to mix dev and non-dev usages of a given module extension, but that's not the case. This PR thus isn't needed.

I have a concrete use case in which I have to use `rules_jvm_external` to fetch both dev and non-dev dependencies, but I can probably use different repo names for that.

PR #815, opened 2022-12-29
Attempt to prevent CI from failing when running in parallel

shs96c:


shs96c:
See https://github.com/bazelbuild/bazel/issues/3236 for more context

jin (approved):

jin:
https://buildkite.com/bazel/rules-jvm-external/builds?branch=master doesn't appear to be flaky currently.

PR #818, opened 2023-01-01
Provide an abstraction over the lock file format

shs96c:
This starts laying the groundwork for us to replace the lock file format with something else without needing to rewrite everything from scratch.

### DIFF
--- a/private/rules/v1_lock_file.bzl
+++ b/private/rules/v1_lock_file.bzl
@@ -50,20 +50,19 @@ def _compute_lock_file_hash(lock_file_contents):
     return hash(repr(sorted(signature_inputs)))
 
 def create_dependency(dep):
-    if not dep.get("url") and not dep.get("file"):
-        return None
-
     url = dep.get("url")
     if url:
         urls = [_remove_auth_from_url(url)]
         urls.extend([_remove_auth_from_url(u) for u in dep.get("mirror_urls", []) if u != url])
-    else:
+    elif dep.get("file"):
         urls = ["file://%s" % dep["file"]]
+    else:
+        urls = []
 
     return {
         "coordinates": dep["coord"],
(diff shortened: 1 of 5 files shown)


shs96c:
There is a required follow-up PR to handle the unpinned repo case, which will allow us to delete some of the code in `coursier.bzl` that is currently duplicated in the `v1_lock_file.bzl` file.

private/rules/v1_lock_file.bzl
```diff
@@ -0,0 +1,166 @@
+def _remove_auth_from_url(url):
```
jin, line 1: nit: license header

private/rules/v1_lock_file.bzl
```diff
@@ -0,0 +1,166 @@
...
+    protocol, url_parts = _split_url(url)
+    host = url_parts[0]
+    if "@" not in host:
+        return url
+    last_index = host.rfind("@", 0, None)
+    userless_host = host[last_index + 1:]
+    new_url = "{}://{}".format(protocol, "/".join([userless_host] + url_parts[1:]))
+    return new_url
+
+def _is_valid_lock_file(lock_file_contents):
```
jin, line 14: Should we also apply a recursive spec check on the expected keys in the dependency_tree object?

jin (approved):
thanks! lgtm

private/rules/v1_lock_file.bzl
shs96c: Added

private/rules/v1_lock_file.bzl
shs96c: Agreed. This does need more checks in it :) Done!

jin (approved):

PR #821, opened 2023-01-05
Automate updating the prebuilt binaries

shs96c:
Also involves actually running the script. Because we're now using hermetic java toolchains, anyone else running this script should result in a no-op.


scripts/BUILD
```diff
@@ -93,3 +93,25 @@ EOL
         ":google-java-format",
     ],
 )
+
+genrule(
```
jin, line 97: Should this be a sh_binary (executed with bazel run) instead of a genrule?

scripts/copy-prebuilts.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+set -eufo pipefail
+
+root="$BUILD_WORKSPACE_DIRECTORY/private/tools/prebuilt"
```
jin, line 4: consider parameterizing the directory in the sh_binary args too?

jin (approved):

scripts/copy-prebuilts.sh
shs96c: Done

PR #822, opened 2023-01-06
V2 lock file format

shs96c:
Introduces (and uses!) a new lock file format. This format is designed to change less as dependencies are updated, which _should_ make reviewing update PRs in GitHub a nicer experience. As a nice side-effect, the new lock files are also (typically) more compact.


private/tools/prebuilt/BUILD
```diff
@@ -1,7 +1,9 @@
 # This package contains static executables used by the maven_install repository rule.
+# To update the prebuilt binaries, `bazel run //scripts:refresh-prebuilts`
 
 exports_files([
-    "hasher_deploy.jar",  # built from //private/tools/java:hasher_deploy.jar
-    "list_packages_deploy.jar",  # build from //private/tools/java/rules/jvm/external/jar:ListPackages_deploy.jar
+    "hasher_deploy.jar",  # built from //private/tools/java/rules/jvm/external:hasher-tool_deploy.jar
```
jin, line 5: Do you know if we can use buildkite to create these jars on each CI build? Then at least we'd have non-local dev builds.

private/versions.bzl
```diff
@@ -6,18 +6,3 @@ COURSIER_CLI_GITHUB_ASSET_URL = "https://github.com/coursier/coursier/releases/d
 # Run 'bazel run //:mirror_coursier' to upload a copy of the jar to the Bazel mirror.
 COURSIER_CLI_BAZEL_MIRROR_URL = "https://mirror.bazel.build/coursier_cli/" + COURSIER_CLI_HTTP_FILE_NAME + ".jar"
 COURSIER_CLI_SHA256 = "2188e58a31a34251b5898358245a5cf7db5d03440d354cc2b1bbd15a260b936b"
-
-JQ_VERSIONS = {
```
jin, line 10: Nice.

jin (approved):

meisterT:
Simon, can you given an example for the new lockfile format?

jin:
@meisterT the diffs contain the newly generated json lockfiles.

meisterT:
Thanks Jin, github didn't load it by default and so I missed it.

Is the summary of the new format to reduce nesting?

shs96c:
The fundamental goal is to have a lock file format that's easier to review by humans (eg. in PR reviews) A secondary goal is to make the build file more compact, but that's really just a nice benefit.

FWIW, I created a small diff to demonstrate what an update might look like. This is [bumping the guava version](https://github.com/bazelbuild/rules_jvm_external/pull/822/commits/884b96a5b280ad09c1e41523cc9ab2f9463dab8d).

meisterT:
> The fundamental goal is to have a lock file format that's easier to review by humans (eg. in PR reviews) A secondary goal is to make the build file more compact, but that's really just a nice benefit.

You achieved this goal by avoiding nesting or are there other things to consider?


> FWIW, I created a small diff to demonstrate what an update might look like. This is [bumping the guava version](https://github.com/bazelbuild/rules_jvm_external/pull/822/commits/884b96a5b280ad09c1e41523cc9ab2f9463dab8d).

That looks very easy to review indeed!

shs96c:
> You achieved this goal by avoiding nesting or are there other things to consider?

The existing lock file contains a lot of duplicate information, the main thing being that there are not only the direct dependencies listed in one property, but also all the transitive dependencies, in each and every dependency. By removing that duplication, the file gets a lot smaller (particularly in regular projects --- the examples in the `rules_jvm_external` repo tend to be tightly focused, so aren't that big).

private/tools/prebuilt/BUILD
shs96c: I'm not sure. We'd also (presumably) want buildkite to update the PR or fail the build if these change....

cheister:
LGTM.  

small nit:  Running `bazel run @unpinned_rules_jvm_external_deps//:pin` on your branch sorts the repositories differently in `rules_jvm_external_deps_install.json`

cheister (approved):

private/tools/prebuilt/BUILD
jin: Yeah that could work too. Mind adding a test?

coursier.bzl
```diff
@@ -476,7 +367,14 @@ def _pinned_coursier_fetch_impl(repository_ctx):
     )
     maven_install_json_content = json.decode(repository_ctx.read(repository_ctx.attr.maven_install_json))
 
-    importer = v1_lock_file
+    if v1_lock_file.is_valid_lock_file(maven_install_json_content):
```
jin, line 370: Let's keep a v1 JSON around for tests?

jin (approved):
lgtm some testing nits

shs96c:
I've been asking for comments on the lock file format from people at work, and they suggested that it might be easier for a human to review if the version and shasums were grouped closer to the artifact key in the lock file. I've created some straw-man example updates here.

The [first](https://github.com/shs96c/rules_jvm_external/commit/ffa4887c9093f7f0e94d307c475319b1eb23b310) uses the lock file format currently proposed in this PR.

The [second](https://github.com/shs96c/rules_jvm_external/commit/7c106157d0020736843f2c11defca8a97f109f0c) uses the format amended to bring shasums closer to the version.

I'd appreciate any input before I continue with the PR....

coursier.bzl
shs96c: Sure

alexeagle:
Congrats on shipping :)

rdesgroppes:
Nice, thank @shs96c! Would it be possible to expose the `conflict_resolution` section in a form or another?

shs96c:
Can you please file a new issue for that? It should be easy enough to add it. 

I'd also very happily accept a PR :)

rdesgroppes:
> Can you please file a new issue for that? It should be easy enough to add it.
> 
> I'd also very happily accept a PR :)

Thanks for doing it through #862!

PR #825, opened 2023-01-09
Normalize path returned by get_coursier_cache_or_default in _relativize_and_symlink_file_in_coursier_cache

### LINKED ISSUE #824: bazel run @unpinned_maven//:pin will fail if HOME as a trailing slash
# Problem

In coursier.bzl, `_relativize_and_symlink_file_in_coursier_cache` expects `get_coursier_cache_or_default` to return a normalized path. However, if environment variables like `HOME` or `COURSIER_CACHE` have a trailing slash, then `_relativize_and_symlink_file_in_coursier_cache` will fail.

# Steps to reproduce

In the root of `rules_jvm_external`, add a trailing slash to the `HOME` environment variables, and then run `bazel run @unpinned_maven//:pin`. This should cause a failure like "Error while trying to parse the path of file in the coursier cache:".

anthonyjpratti:
Fixes #824.

Ideally, this would use paths.normalize in skylib, but that would require changes to how rules_jvm_external setups up dependencies.

### DIFF
--- a/coursier.bzl
+++ b/coursier.bzl
@@ -190,12 +190,17 @@ def _relativize_and_symlink_file_in_coursier_cache(repository_ctx, absolute_path
     # Also replace '\' with '/` to normalize windows paths to *nix style paths
     # BUILD files accept only *nix paths, so we normalize them here.
     #
+    # Also, replace '//' with '/', otherwise parsing of the file path for the
+    # coursier cache will fail if variables like HOME or COURSIER_CACHE have a
+    # trailing slash.
+    #
     # We assume that coursier uses the default cache location
     # TODO(jin): allow custom cache locations
-    absolute_path_parts = absolute_path.split(get_coursier_cache_or_default(
+    coursier_cache_path = get_coursier_cache_or_default(
         repository_ctx,
         repository_ctx.attr.use_unsafe_shared_cache or repository_ctx.attr.name.startswith("unpinned_"),
-    ))
+    ).replace("//", "/")
+    absolute_path_parts = absolute_path.split(coursier_cache_path)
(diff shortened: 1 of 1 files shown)


shs96c:
FYI, `skylib` is pulled in as a dependency via the [repositories.bzl file](https://github.com/bazelbuild/rules_jvm_external/blob/master/repositories.bzl#L13-L21). It's safe to rely on it. Would you prefer to use that?

anthonyjpratti:
> FYI, `skylib` is pulled in as a dependency via the [repositories.bzl file](https://github.com/bazelbuild/rules_jvm_external/blob/master/repositories.bzl#L13-L21). It's safe to rely on it. Would you prefer to use that?

So that was my first attempt, but I ran into the following issue. In rules_jvm_external_deps func, there is a bazel_skylib http archive and a maven_install; since maven_install relies on coursier.bzl and bazel_skylib isn't defined until rules_jvm_external_setup is invoked, skylib fails to be loaded in coursier.bzl. Since rules_jvm_external_deps and rules_jvm_external_setup are part of the public api for rules_jvm_external rules, I didn't want to modify that.

I'm open to recommendations on how to import skylib without breaking backwards compatibility. If there aren't concerns about backwards compatibility, then I do have a path forward for relying on skylib, but that's y'alls call.

shs96c:
How frustrating! Well then, this will be excellent :) Thank you for the patch, and the explanation!

shs96c (approved):
LGTM

anthonyjpratti:
> How frustrating! Well then, this will be excellent :) Thank you for the patch, and the explanation!

I found it to be a great learning experience! I earnestly love when things break in ways I don't expect! :)

PR #829, opened 2023-01-15
Clean test run

shs96c:
With these changes, `bazel test //...` on macOS runs cleanly, without failures.


tests/com/jvm/external/UnsafeSharedCacheTest.java
```diff
@@ -1,54 +0,0 @@
...
-      return;
-    }
-    boolean foundGuavaJar = false;
-    ClassLoader cl = ClassLoader.getSystemClassLoader();
-    for (URL url : ((URLClassLoader) cl).getURLs()) {
-      URI uri = url.toURI();
-      String uriString = uri.toString();
-      if (uriString.contains("repo1.maven.org") && uriString.contains("guava")) {
-        foundGuavaJar = true;
-        if (OS.indexOf("mac") >= 0) {
```
cheister, line 34: Does this test not pass on Mac?  Or do we just not need this test anymore?

shs96c: It doesn't pass anywhere. Ever since we switched to adding explicit visibilities on targets rather than having it set at the package level (#649), this test hasn't passed anywhere since we're not exporting the files. I've no idea why our CI runs haven't caught this, but it's been that way for over a year.

jin: Let's delete the feature, then?

shs96c: Sure. I can do that.

private/rules/maven_bom_fragment.bzl
```diff
@@ -12,10 +12,22 @@ MavenBomFragmentInfo = provider(
 )
 
 def _maven_bom_fragment_impl(ctx):
+    java_info = ctx.attr.artifact[JavaInfo]
+
+    # Recent Bazel versions
```
jin, line 17: nit: mention the exact version cutover

jin (approved):

private/rules/maven_bom_fragment.bzl
shs96c: Done

PR #841, opened 2023-01-24
Add test to ensure generated docs and files are up to date

shs96c:
This also removes a need on the root build file to depend on stardoc.


cheister (approved):
I don't think the test failure is related to this change?

shs96c:
Apparently not needed any more: this has been open for years.

PR #849, opened 2023-02-03
Make the implementation of `maven_publish` "not private"

shs96c:
It's not officially part of our public API, but sophisticated users who are attempting to extend `maven_publish` will find this helpful

### DIFF
--- a/private/rules/maven_publish.bzl
+++ b/private/rules/maven_publish.bzl
@@ -5,6 +5,7 @@ MavenPublishInfo = provider(
         "javadocs": "Javadoc jar file for documentation files",
         "artifact_jar": "Jar with the code and metadata for execution",
         "source_jar": "Jar with the source code for review",
+        "target": "The original `Label` that is being published",
     },
 )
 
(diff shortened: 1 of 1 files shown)


jin:
Is there a particular reason that the entire impl needs to be exported? Can we keep it locked down and improve the attributes? Just for API simplicity reasons.

shs96c:
In this particular case, I want to add an extra provider to the output of the rule (for a ruleset that isn't public). In theory, I could add a new target, but that would break existing workflows, where people are already using the `.publish` implicit target. By exposing the implementation it becomes available for re-use.

jin:
When I hear "extra provider", I think of aspects. Would that work?

thirtyseven:
I'll just chime in that in my company, we have a use case where having `maven_publish` be public would be useful. We are migrating an existing workflow from Maven where we use the Maven assembly plugin to publish a tarball as the primary artifact for a given coordinate. It would be really useful for us to have `maven_publish` available as a primitive so we can write a custom `java_export` macro that supports this.

shs96c:
Leaving this

PR #859, opened 2023-02-23
transitive dependencies should be visible to their compat alias

caseyduquettesc:
When `generate_compat_repositories` is True, the `jvm_import` target  is not visible to the compat alias, causing an error when trying to use the compat label.

```
# <>/external/maven/BUILD:2881:11
jvm_import(
  name = "com_google_j2objc_j2objc_annotations",
  visibility = ["//visibility:private"],
  tags = ["maven_coordinates=com.google.j2objc:j2objc-annotations:1.3", "maven_url=<>/j2objc-annotations-1.3.jar"],
  jars = ["@maven//:<>/j2objc-annotations-1.3.jar"],
  srcjar = "@maven//:<>/j2objc-annotations-1.3-sources.jar",
  deps = [],
)

# <>/external/com_google_j2objc_j2objc_annotations/jar/BUILD:2:6
alias(
  name = "jar",
  actual = "@maven//:com_google_j2objc_j2objc_annotations",
)
```

I propose adding the alias path to the visibility of the `jvm_import` when compat repositories are being generated.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -339,6 +339,23 @@ maven_install(
 )
 
 maven_install(
+    name = "strict_visibility_with_compat_testing",
+    artifacts = [
+        # Must not be in any other maven_install where generate_compat_repositories = True
+        "com.google.http-client:google-http-client-gson:1.42.3",
+    ],
+    repositories = [
+        "https://repo1.maven.org/maven2",
+    ],
+    strict_visibility = True,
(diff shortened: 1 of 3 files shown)


shs96c (approved):
Thank you for adding the test. The build is green. Let's see what happens :) 

Merging now. Thank you for the PR!

PR #862, opened 2023-02-24
Add back conflict_resolutions to the lock file

shs96c:


cheister:
What was breaking without this?

cheister (approved):

shs96c:
I've heard from early users of this that they use the conflict resolution section to figure out if they need to update any dependencies in their `artifacts`. They try and keep it down to zero conflicts.

rdesgroppes:
Thanks, @shs96c!

> I've heard from early users of this that they use the conflict resolution section to figure out if they need to update any dependencies in their `artifacts`. They try and keep it down to zero conflicts.

Yes, this is extremely helpful in the absence of BOM support.

PR #863, opened 2023-02-26
Set http.proxyProtocol and https.proxyProtocol for coursier

cheister:
In coursier version v2.1.0-M7 and later there was a [change to the proxy setup](https://github.com/coursier/coursier/pull/2440) that requires `http.proxyProtocol` or `https.proxyProtocol` to be set for the proxy to work.  I created https://github.com/coursier/coursier/pull/2701 to fix this, but until that is merged we need to set the values in our proxy settings.

This PR also adds better verbose logging for args_files.


shs96c (approved):
LGTM. Ship it!

Issue #865, opened 2023-03-01
java.util.zip.ZipException: invalid entry CRC while build still compiles fine


nkoroste:
After upgrading to the latest version of these rules I started getting the following error:
```
2023-02-28 04:13:37 12:13:37 [Bazel] Error in fail: Error while indexing jar packages: Exception in thread "main" java.io.UncheckedIOException: java.util.zip.ZipException: invalid entry CRC (expected 0x8994712d but got 0xa0733c67)
```
While compiling using the same jar works fine. 

Taking a random malformed jar from maven https://repo.maven.apache.org/maven2/com/google/apis/google-api-services-compute/v1-rev235-1.25.0/ 
```
jar -tf google-api-services-compute-v1-rev235-1.25.0-javadoc.jar
java.util.zip.ZipException: zip END header not found
	at java.base/java.util.zip.ZipFile$Source.zerror(ZipFile.java:1581)
	at java.base/java.util.zip.ZipFile$Source.findEND(ZipFile.java:1476)
	at java.base/java.util.zip.ZipFile$Source.initCEN(ZipFile.java:1483)
	at java.base/java.util.zip.ZipFile$Source.<init>(ZipFile.java:1288)
	at java.base/java.util.zip.ZipFile$Source.get(ZipFile.java:1251)
	at java.base/java.util.zip.ZipFile$CleanableResource.<init>(ZipFile.java:732)
	at java.base/java.util.zip.ZipFile$CleanableResource.get(ZipFile.java:849)
	at java.base/java.util.zip.ZipFile.<init>(ZipFile.java:247)
	at java.base/java.util.zip.ZipFile.<init>(ZipFile.java:177)
	at java.base/java.util.zip.ZipFile.<init>(ZipFile.java:148)
	at jdk.jartool/sun.tools.jar.Main.list(Main.java:1500)
	at jdk.jartool/sun.tools.jar.Main.run(Main.java:380)
	at jdk.jartool/sun.tools.jar.Main.main(Main.java:1681)
```
```
unzip -l google-api-services-compute-v1-rev235-1.25.0-javadoc.jar
Archive:  google-api-services-compute-v1-rev235-1.25.0-javadoc.jar
  End-of-central-directory signature not found.  Either this file is not
  a zipfile, or it constitutes one disk of a multi-part archive.  In the
  latter case the central directory and zipfile comment will be found on
  the last disk(s) of this archive.
[...]

PR #867, opened 2023-03-03
list_packages should not fail when the jar is invalid

cheister:
Should fix https://github.com/bazelbuild/rules_jvm_external/issues/865


shs96c (approved):
Nice catch. Thank you.

PR #869, opened 2023-03-08
Add a doc on how to release RJE

shs96c:


RELEASING.md
```diff
@@ -0,0 +1,120 @@
...
+edit the values it gives you. The release notes should include user
+visible changes and all commits by people who aren't core
+committers. Bug fixes and little tweaks the core committers have made
+can be omitted if you don't think they're particularly noteworthy.
+
+Name the draft release after the version you've picked.
+
+## Create an archive of the release, and add as an asset
+
+GitHub periodically break how archives are created. To avoid this,
```
jin, line 60: I believe the URLs are different for manually created assets, right? Should we also update our default WORKSPACE snippets?

jin (approved):

RELEASING.md
shs96c: That's what I did for the 5.0 release. If people copy-and-paste from an earlier release, then they'll be mostly there already :)

Issue #872, opened 2023-03-13
Guava Maven Artifact with wrong classifier (-android instead of -jre) gets selected via dependency instead of user selected one


vorburger:
I have a full project which illustrates this problem, it's currently in a private repo but I'm hoping to open it within days, if it helps:

Despite using `"com.google.guava:guava:31.1-jre"` in my `MODULE.bazel`, it seemed to pull an older Android instead of JRE version:

```
$ bazelisk query "@maven//:*" | grep guava
Loading: 0 packages loaded
Loading: 0 packages loaded
@maven//:com_google_guava_failureaccess
@maven//:com_google_guava_failureaccess_1_0_1
@maven//:com_google_guava_failureaccess_1_0_1_extension
@maven//:com_google_guava_failureaccess_jar_sources_1_0_1_extension
@maven//:com_google_guava_guava
@maven//:com_google_guava_guava_30_1_1_android
@maven//:com_google_guava_guava_30_1_1_android_extension
@maven//:com_google_guava_guava_jar_sources_30_1_1_android_extension
@maven//:com_google_guava_listenablefuture
@maven//:com_google_guava_listenablefuture_9999_0_empty_to_avoid_conflict_with_guava
@maven//:com_google_guava_listenablefuture_9999_0_empty_to_avoid_conflict_with_guava_extension
@maven//:v1/https/repo1.maven.org/maven2/com/google/guava/failureaccess/1.0.1/failureaccess-1.0.1-sources.jar
@maven//:v1/https/repo1.maven.org/maven2/com/google/guava/failureaccess/1.0.1/failureaccess-1.0.1.jar
@maven//:v1/https/repo1.maven.org/maven2/com/google/guava/guava/30.1.1-android/guava-30.1.1-android-sources.jar
@maven//:v1/https/repo1.maven.org/maven2/com/google/guava/guava/30.1.1-android/guava-30.1.1-android.jar
@maven//:v1/https/repo1.maven.org/maven2/com/google/guava/listenablefuture/9999.0-empty-to-avoid-conflict-with-guava/listenablefuture-9999.0-empty-to-avoid-conflict-with-guava.jar
```

This seems to somehow be caused by me having also had a `"io.kubernetes:client-java:17.0.1"` in my `MODULE.bazel`. When I upgraded that to the latest `"io.kubernetes:client-java:18.0.0"` then it worked:

```
$ bazelisk query "@maven//:*" | grep guava
@maven//:com_google_guava_failureaccess
[...]

PR #873, opened 2023-03-13
docs: Remove hard-coded version in bazel_dep example

vorburger:
I think it's better for readers to NOT have an (ancient) version in an example, and let them go look it up.

At least I for one was too stupid to notice that inline `FIXME` and it caused me https://github.com/bazelbuild/rules_jvm_external/issues/872 pain.

Explicitly documenting how and where one has to go look up the correct version seems preferable.

### DIFF
--- a/docs/bzlmod.md
+++ b/docs/bzlmod.md
@@ -17,11 +17,11 @@ The simplest way is by adding this line to your `.bazelrc`:
 common --enable_bzlmod
 ```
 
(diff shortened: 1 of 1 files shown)


shs96c (approved):
Seems reasonable. We should really update our release notes to make something that people can copy-and-paste, but this will probably help avoid a common problem. 

Thank you for the PR!

PR #874, opened 2023-03-14
Handle the case where dependencies are empty

sitaktif:
Sometimes, some issue lead to the `pinned_maven_install()` function being completely empty, making the defs.bzl file syntactically incorrect, causing all sorts of issues making it hard to recover (needing a `bazel clean --expunge` or a manual edition of the file to fix the syntax).

This prevents that kind of situation.


shs96c (approved):
LGTM

PR #877, opened 2023-03-23
Add targets to make it easier to vendor the `@maven` repository

meteorcloudy:
This change is required to support Bazel's offline bootstrap build. More context in https://github.com/bazelbuild/bazel/pull/17112

Instead of checking in jar files in Bazel's source tree, Bazel wants to use rules_jvm_external to fetch jars dependencies. However, to support Bazel's bootstrap build, we need to patch rules_jvm_external for vendoring the `@maven` repository.

- Generate a BUILD.vendor file to be used in the vendored `@maven` repository. Added a jvm_import and a filegroup rule for each downloaded jar artifact. The filegroup rule is required by the bootstrap Java toolchain used in Bazel's bootstrap build. The bootstrap Java toolchain cannot depend on a jvm_import target. Because the jvm_import rule depends on a java_binary tool "AddJarManifestEntry", which requires a Java toolchain. Depending on the jar via a filegroup rule helps avoid this cyclic dependency.
- Added a filegroup rule to collect all sources needed for vendoring `@maven`, including BUILD.vendor, WORKSPACE and jar files.


meteorcloudy:
This PR upstream the rules_jvm_external patch we checked in Bazel repo: https://github.com/bazelbuild/bazel/blob/master/third_party/rules_jvm_external.patch

shs96c:
So, boiling this down to the smallest possible requirement, what we want is a `filegroup` containing all the jar files. Is that right? I'm asking because there are three different concepts that all seem closely related, and I'm loathe to add hacks to solve one when we could come up with something clean that meets all our needs. Those things are:

1. "[Distribution zips](https://docs.gradle.org/current/userguide/distribution_plugin.html)", which are collections of artifacts and their dependencies, which is how we used to distribute java libraries before Maven became A Thing.
2. "Frozen dependencies", as used by contrib_rules_jvm. These are used to "freeze" a valid workspace generated by `rules_jvm_external` into a zip file that can then be used directly as a repository. This means that rulesets using frozen deps can do the "load dependencies, setup dependencies" two-step process rather than needing a third step to call `pinned_maven_install`
3. This somewhat specialised use-case for Bazel.

I suspect that if we created a distribution zip it would be easy enough to use the [zip_repository](https://github.com/bazel-contrib/rules_jvm/blob/main/java/private/zip_repository.bzl) rule from `2` to allow a custom build file which would support this use case. Alternatively, the zip could be uploaded "somewhere"  and downloaded as an `http_archive` by Bazel during the build.

meteorcloudy:
To elaborate more on the Bazel use case, we mainly need this patch to support Bazel's bootstrap build, which has two main differences than the normal `bazel build //src:bazel`:

- We ensure users can bootstrap Bazel without internet access with the dist zip file we publish, e.g. [bazel-7.0.0-pre.20230306.4-dist.zip](https://github.com/bazelbuild/bazel/releases/download/7.0.0-pre.20230306.4/bazel-7.0.0-pre.20230306.4-dist.zip). This means we have to archive all external dependencies needed for building Bazel into the dist zip file, [this target](https://github.com/bazelbuild/rules_jvm_external/pull/877/files#diff-1e45d7cd4042693211991596278cb73efd11b128ca5084b7cacb61ed38e65acaR431-R432) added in the PR can collect all artifacts we need, then we build them into an archive [here](https://github.com/bazelbuild/bazel/blob/aa4f3580d277e50757ea696e3a312c0921c304f2/BUILD#L180-L190). I think this indeed sounds like the "Frozen dependencies" case, but the current solution here is quite simple, and can probably be used by other projects as well.
- The second one is what I described in the PR. We need a bootstrap Java toolchain which shouldn't depend on any Java rule itself. Therefore we need to [reference certain jar files with filegroup rule](https://github.com/bazelbuild/bazel/blob/aa4f3580d277e50757ea696e3a312c0921c304f2/third_party/BUILD#L249-L252) instead of jvm_import. So what we need is not "one `filegroup` containing all the jar files", but "one `filegroup` for each jar file". There are also some use cases, when users do know exact what jar files they need and don't want to fetch the entire transitive jar dependencies, for example, the [android_common jars](https://github.com/bazelbuild/bazel/blob/aa4f3580d277e50757ea696e3a312c0921c304f2/src/tools/android/java/com/google/devtools/build/android/BUILD#L130-L147).

Overall, I'm fine with keeping the patch as a custom solution for Bazel itself, but I was wondering if this change could also be helpful for other users.

jin:
So far I think this has _major_ utility for Bazel's unique bootstrapping process, but we haven't heard (or probably missed) requirements from other projects that require this. I'm leaning towards not adding this complexity _right now_, and come up with a generalizable blessed path if the FR comes up again for non-Bazel usage. 

The patched code in question also should not experience much churn, so I expect Bazel's downstream patches to continue to work for while.

Does that make sense?

meteorcloudy:
> The patched code in question also should not experience much churn, so I expect Bazel's downstream patches to continue to work for while.

It's good to know that!

> Does that sound reasonable?

Yes, this is totally reasonable, we can revisit again when we hear more similar needs from users.

PR #882, opened 2023-03-24
Ensure the generated pinned_maven_install target is always syntactically correct

shs96c:


jin (approved):

coursier.bzl
```diff
@@ -1121,6 +1121,7 @@ def _coursier_fetch_impl(repository_ctx):
         "load(\"@bazel_tools//tools/build_defs/repo:http.bzl\", \"http_file\")",
         "load(\"@bazel_tools//tools/build_defs/repo:utils.bzl\", \"maybe\")",
         "def pinned_maven_install():",
+        "    pass",  # Ensure we're syntactically correct even if no deps are added
```
jin, line 1124: Instead of `pass`, maybe a `print` with a warning that there are no deps would be more useful?

shs96c: I did think about that, but the `pass` statement is appended regardless of whether or not there are artifacts: it's a no-op, but adding it makes this function correct even if we don't append anything else, and we don't need to have an extra `if` statement in the code.

As for warning people that they have no dependencies, I think that's sufficiently unlikely to occur that the only time it does will be deliberate.

Issue #875, opened 2023-03-20
New lock-file format results in repositories with basic-auth having no artifacts


davidkurkov:
When a list of repositories is passed-in to `maven_install` where one of the repo-urls has basic-auth, the resulting v2 lockfile will not having any artifacts matched to the repo-url if it's using basic-auth.

**Steps to reproduce**
- Pass in a valid `repo-url-with-basic-auth` to maven-install as part of the repositories argument
    - i.e., ``https://un:pw@custom-artifactory.net/
- Re-pin maven_install.json file: `bazel run '@unpinned_maven//:pin`

**Expected:** the repo-url-with-basic-auth under the `repositories` key will contain resolved artifacts.
**Actual:** the repo-url-with-basic-auth under the `repositories` key contains no resolved artifacts.

I believe the incorrect matching is happening in [private/tools/java/com/github/bazelbuild/rules_jvm_external/coursier/NebulaFormat.java](https://github.com/bazelbuild/rules_jvm_external/blob/master/private/tools/java/com/github/bazelbuild/rules_jvm_external/coursier/NebulaFormat.java#L74). The basic-auth is stripped [here](https://github.com/bazelbuild/rules_jvm_external/blob/master/private/tools/java/com/github/bazelbuild/rules_jvm_external/coursier/NebulaFormat.java#L31) and not stripped further [down here](https://github.com/bazelbuild/rules_jvm_external/blob/master/private/tools/java/com/github/bazelbuild/rules_jvm_external/coursier/NebulaFormat.java#L74).

shs96c:
A PR on this repo would be very helpful!

PR #883, opened 2023-03-24
Strip auth from url-with-basic-auth when mapping coords

davidkurkov:
Strip auth from url-with-basic-auth when matching coords to repo-urls

Fixes https://github.com/bazelbuild/rules_jvm_external/issues/875


shs96c (approved):
Thank you for sending this!

Issue #855, opened 2023-02-16
Expose maven_export and clean up kotlin/scala merged jars


sacsar:
Following up on our Slack discussion, there are a few things to do around exposing `maven_export` so it can be used with Scala:
- expose `maven_export` via `defs.bzl`
- update the jar merge logic to exclude "compiler artifacts" (I don't know what to call this) that get pulled in for Kotlin and Scala because they don't have Maven coordinates:
  - For Kotlin:`@com_github_jetbrains_kotlin//:kotlin-stdlib`
  - For Scala: `@io_bazel_rules_scala_scala_reflect//:io_bazel_rules_scala_scala_reflect` and `@io_bazel_rules_scala_scala_library//:io_bazel_rules_scala_scala_library` (there may be others, but this was sufficient for my use case)

For anyone encountering the issue of standard library classes making it into the jar: add the appropriate targets to the `deploy_env` parameter.

I'm happy to make a PR or two for these. (My patch to expose `maven_export` in my project is a whopping three line diff, but a PR would need some docs too. The other is bigger, I think.)

shs96c:
@sacsar, I think that what we're pretty close to being done here. As you point out, we can use the `deploy_env` parameter to avoid packing standard libraries. Should we consider exposing a `scala_export` and `kt_java_export` that are simply macros that add the standard libraries to a regular `java_export`?

sacsar:
@shs96c IIRC, when I first wrote the ticket, I was concerned about the macro approach being fragile. However, it's a decent starting point and there can be a docstring to point people in the right direction in case my lists were incomplete.

One wrinkle is that there isn't currently a dependency on `rules_scala` -- are you okay with one being added?

I can take a crack at one or both this weekend.

shs96c:
Apologies for not replying. Adding a dep on `rules_scala` is fine. The way we've handled this for the kotlin rules is to have a top-level `kt_defs.bzl` that includes the rules that depend on `rules_kotlin`: it's expected that users will have loaded that themselves already. A similar approach for `rules_scala` would work well

sacsar:
Sounds good. I'll take a crack at it probably next week. Last I checked `rules_scala` isn't on bzlmod yet, so I may need help. (My attempts at converting my own project to bzlmod have not gone very well due to dependencies that aren't using it yet, even though I know it's meant to be possible.)

PR #884, opened 2023-03-27
Add kotlin_stdlib to kt_jvm_export deploy_env.

sacsar:
Update `kt_jvm_export` to add the kotlint standard library to the deploy env so that it is not pulled in by maven_export.

I got the debug message telling me to update the lock file, so I did.

This covers part of #855 -- I haven't handled Scala here because there is not an existing dependency on rules_scala (and it does not currently support bzlmod).


shs96c (changes requested):
Please revery the change to the v1 lock file: we left it in the old format so that we could continue to check whether any changes we make will break existing users who haven't updated their lock files.

Other than that, this looks good.

shs96c (approved):
I worry about what we'll do if the stdlib isn't available, but we'll cross the bridge when someone files a bug report, since I _think_ this is the way that `rules_kotlin` sets things up for everyone.

PR #886, opened 2023-04-05
Skip non-source maven local dependencies in `generate_imports`

### LINKED ISSUE #885: Repinning fails when using `m2local`
Steps to reproduce:
- Declare a `maven_install` rule with a `m2Local` in `repositories` and an artifact from that repository
- Follow [the steps for creating a `maven_install.json` file](https://github.com/bazelbuild/rules_jvm_external#pinning-artifacts-and-integration-with-bazels-downloader)
- Run `bazel run @unpinned_maven//:pin`

The build will fail with an error like:
```
ERROR: /private/var/tmp/_bazel_lukaciko/6412dece0b45142c73534609944ce663/external/unpinned_m2local_testing_repin/BUILD:9:11: @unpinned_m2local_testing_repin//:com_example_kt: invalid label '/Users/lukaciko/.m2/repository/com/example/kt/1.0.0/kt-1.0.0.jar' in element 0 of attribute 'jars' in 'jvm_import' rule: invalid target name '/Users/lukaciko/.m2/repository/com/example/kt/1.0.0/kt-1.0.0.jar': target names may not start with '/'
```

This worked in the original PR that added support for `m2local` #532 but not the rebased PR #695. I believe that's because that this check was moved and is only done for source jars:

```python
# Skip the maven local dependencies if requested
if skip_maven_local_dependencies and is_maven_local_path(artifact_path):
    continue  
```

lukaciko:
Fixes #885.

Haven't fully grokked the code but I think this could be the fix. It's closer to the original PR adding `m2local` support, #532. `bazel query @m2local_testing_repin//:com_example_kt --output=build` still gives me the expected generated import:

```
jvm_import(
  name = "com_example_kt",
  tags = ["maven_coordinates=com.example:kt:1.0.0", "maven_url=None"],
  jars = ["@m2local_testing_repin//:v1/com/example/kt/1.0.0/kt-1.0.0.jar"],
  deps = [],
)
```


tests/custom_maven_install/m2local_testing_with_pinned_file_install.json
```diff
@@ -0,0 +1,52 @@
+{
```
shs96c, line 1: How come this lock file is in the old format and not the new one? Could you please regenerate it using the current `HEAD` of `master`?

shs96c (commented):
Thank you for the PR! I think it makes sense, but it looks like the lock file was generated using an older version of `rules_jvm_external`. Could you please repin using the latest version?

tests/custom_maven_install/m2local_testing_with_pinned_file_install.json
lukaciko: My mistake, was working off an older commit. Regenerated the file now.

shs96c (approved):
Thank you for the changes! LGTM!

PR #900, opened 2023-05-12
fix(pinning): use runfiles lookup

### LINKED ISSUE #899: [windows] pin: cannot find sorted_deps.json
Getting this error on Windows:

```
$ bazel run @unpinned_maven//:pin
INFO: Analyzed target @rules_jvm_external~5.2~maven~unpinned_maven//:pin (0 packages loaded, 20 targets configured).
INFO: Found 1 target...
Target @rules_jvm_external~5.2~maven~unpinned_maven//:pin up-to-date:
  bazel-bin/external/rules_jvm_external~5.2~maven~unpinned_maven/pin
  bazel-bin/external/rules_jvm_external~5.2~maven~unpinned_maven/pin.exe
INFO: Elapsed time: 0.323s, Critical Path: 0.03s
INFO: 3 processes: 3 internal.
INFO: Build completed successfully, 3 total actions
INFO: Running command line: bazel-bin/external/rules_jvm_external~5.2~maven~unpinned_maven/pin.exe external/rules_jvm_external~5.2~maven~unpinned_maven/unsorted_deps.json
cp: cannot stat 'external/rules_jvm_external~5.2~maven~unpinned_maven/unsorted_deps.json': No such file or directory
```

It feels like maybe https://github.com/bazelbuild/rules_jvm_external/blob/60a54afe8a566653153cc41d6acb883a21caccc4/coursier.bzl#L63 needs a runfiles lookup through `rlocation`? 
I can reproduce this both with `--enable_bzlmod` and the bits declared in `MODULE.bazel` and also with WORKSPACE and no bzlmod.

I see you do have a windows+pinning test here: https://github.com/bazelbuild/rules_jvm_external/blob/master/.bazelci/presubmit.yml#L86 so I wonder why that's green and mine's not.

alexeagle:
This allows pinning to work with --noenable_runfiles, for example on a typical Windows setup.

Fixes #899


alexeagle:
Hm, this won't work with Bazel 4.0.0 - do you want a workaround so we can construct an rlocationpath the old way?

shs96c:
We're still trying to support Bazel 4, so a fix would be appreciated. If it's not possible, we can discuss whether we want to bump the minimum version to 5.x

alexeagle:
Actually Bazel 5.4.0 is what introduced `$(rlocationpath)`. I changed to `$(execpath)` and trim off the `external/` prefix in the shell script. My understanding is we know the file in this case is always a source file that we wrote into an external repo, so it happens that this will work.

shs96c (approved):
LGTM. Thank you. I find it interesting to note that I can tell you're using a linux machine for development because of the change to the lock file

PR #904, opened 2023-05-22
Allow semicolons in maven_repo

thirtyseven:
Artifactory uses [semicolons in URLs to set properties](https://jfrog.com/help/r/jfrog-artifactory-documentation/using-properties-in-deployment-and-resolution). Currently, when trying to use a maven_repo with semicolons in `maven_publish`, they are interpreted as statement terminators by bash. This PR adds quotes to all the arguments to {uploader} to solve this.


shs96c (approved):
LGTM. Thank you for spotting this and fixing it!

Issue #914, opened 2023-06-08
LockFileConverter does not handle maven SNAPSHOT correctly


jianxx:
LockFileConverter tries to match the expected path for jar files.
It expects the path as "[group]/[artifact]/[version]/[artifact]-[version](-[classifier])(.[extension])".

However, when a jar is from a SNAPSHOT version, its version is something like 1.1.0-SNAPSHOT and the jar file name is [artifact]-1.1.0-[classifier].jar.
LockFileConverter does not handle this SNAPSHOT version case.

PR #915, opened 2023-06-08
Remove SNAPSHOT suffix in version number (#914)

jianxx:
This is a fix for #914 .

SNAPSHOT suffix should not be included in version number.


cheister:
Is there a test that can be added for this case?

jianxx:
sure, I’ll add a function verification test for this scenario.

cheister ***@***.***>于2023年6月20日 周二上午7:02写道：

jianxx:
I just find that coursier uses different file name for SNAPSHOT and this different behavior causes the problem.
Need more investigation into coursier to clarify whether it is a bug or a feature.

shs96c:
@jianxx I'm not sure from your last comment whether you'd like us to review this patch, or leave it. Could you let us know, please?

jianxx:
> @jianxx I'm not sure from your last comment whether you'd like us to review this patch, or leave it. Could you let us know, please?

After I did more tests with different maven repo implementations, I can describe this issue more precisely. My previous statement with courier is not correct, and here is the detailed  root cause of this issue and how my patch works.
1. The issue could be reproduced with nexus3 (I have not tested nexus2 yet )and could not be reproduced with latest nexus build and jfrog.
2. The root cause is how the maven repo handles snapshot arfifacts. Let's take org.springframework.boot:spring-boot-cli:3.2.0-SNAPSHOT as an example. When we use coursier to fetch this artifact from the latest nexus or jfrog, the result will be "v1/https/repo.spring.io/snapshot/org/springframework/boot/spring-boot-cli/3.2.0-SNAPSHOT/spring-boot-cli-3.2.0-SNAPSHOT.jar". But when we fetch from nexus 3.x, the result would be something like "v1/https/repo.spring.io/snapshot/org/springframework/boot/spring-boot-cli/3.2.0-SNAPSHOT/spring-boot-cli-3.2.0-20230622.132711-54.jar"
3. In the maven repo, spring-boot-cli-3.2.0-SNAPSHOT.jar does not really exist. It is an alias to latest build like "spring-boot-cli-3.2.0-20230622.132711-54.jar".

What I did in the patch is using version string without 'SNAPSHOT' to match the artifact file, so it can work for both cases. 
I'm adding a test case with mocking data to simulate nexus 3.x behavior. It is a little expensive to make a full testing scenario.

Issue #917, opened 2023-06-16
v2 lock-file format contains artifacts with missing dependencies


davidkurkov:
We found that at least some artifacts defined in the v2 lock-file format are missing dependencies explicitly defined in an artifact's POM file which causes our builds/tests to fail.

For example, the [mockito-core 3.3.3 POM file](https://repo1.maven.org/maven2/org/mockito/mockito-core/3.3.3/mockito-core-3.3.3.pom) contains explicit dependencies on:
```xml
<dependencies>
    <dependency>
      <groupId>net.bytebuddy</groupId>
      <artifactId>byte-buddy</artifactId>
      <version>1.10.5</version>
      <scope>compile</scope>
    </dependency>
    <dependency>
      <groupId>net.bytebuddy</groupId>
      <artifactId>byte-buddy-agent</artifactId>
      <version>1.10.5</version>
      <scope>compile</scope>
    </dependency>
    <dependency>
      <groupId>org.objenesis</groupId>
      <artifactId>objenesis</artifactId>
      <version>2.6</version>
      <scope>compile</scope>
    </dependency>
  </dependencies>
```

In the v1-lockfile format, the `org.mockito:mockito-core:3.3.3` coordinate detailed the dependency on `bytebuddy`:

```json
{
    "coord": "org.mockito:mockito-core:3.3.3",
    "dependencies": [
        "net.bytebuddy:byte-buddy-agent:1.14.4",
        "net.bytebuddy:byte-buddy:1.14.4",
        "org.objenesis:objenesis:2.6"
    ],
    "directDependencies": [
        "net.bytebuddy:byte-buddy-agent:1.14.4",
        "net.bytebuddy:byte-buddy:1.14.4",
        "org.objenesis:objenesis:2.6"
    ],
}
```

In the v2-lockfile format, it is missing the `bytebuddy` dependency:

```json
{
  "artifacts": {
    "org.mockito:mockito-core": {
      "shasums": {
        "jar": "4be648c50456fba4686ba825000d628c1d805a3b92272ba9ad5b697dfa43036b",
        "sources": "b460650fbe05f5272e58df51628d3adff3dc870a3b19baf805bdea4d237aef1c"
      },
      "version": "3.3.3"
    }
  },
  "dependencies": {
    "org.mockito:mockito-core": [
      "org.objenesis:objenesis"
    ]
  },
  "skipped": [],
  "packages": {
    "org.mockito:mockito-core": []
  },
  "repositories": {
[...]

PR #908, opened 2023-05-26
Add preliminary support for `exe` packaging type

rdesgroppes:
Resolves #907.

This is to allow fetching prebuilt binaries (such as [protoc](https://repo1.maven.org/maven2/com/google/protobuf/protoc/)) in the same way as other artifacts served by maven repositories.

Background: particularly useful for defining (lang) proto toolchains, without having to compile nor manually maintain repository mappings (checksums, etc.) for all supported platforms.

A few related questions/contributions:
- #132
- https://github.com/bazelbuild/rules_proto/issues/35
- https://github.com/bazelbuild/rules_proto/pull/36
- https://github.com/bazelbuild/rules_proto/pull/83
- https://github.com/bazelbuild/rules_proto/pull/107
- https://github.com/bazelbuild/rules_proto/pull/113
- https://github.com/bazelbuild/rules_proto/pull/157
- https://github.com/rules-proto-grpc/rules_proto_grpc/issues/88
- https://github.com/stackb/rules_proto#workspace-boilerplate (`@build_stack_rules_proto//toolchain:prebuilt`)

PR #919, opened 2023-06-21
Fix transitive dep resolution when exclusion is present in other POM

artem-zinnatullin:
#908 uncovered a bug in `coursier` which leads to transitive dep not being resolved if an `exclusion` rule for it is present in POM of a non-related dependency.

This PR adds a regression test and a fix by @rdesgroppes.

TL;TR:

1. `mockito-core` depends on `byte-buddy`
2. `androidx.arch.core:core-testing` [POM](https://dl.google.com/android/maven2/androidx/arch/core/core-testing/2.1.0/core-testing-2.1.0.pom) has `exclusion` for `<groupId>net.bytebuddy</groupId>`
3. If both are present in `maven_install` then `mockito-core` doesn't get its `byte-buddy` dependency
4. If `androidx.arch.core:core-testing` is removed from `maven_install` then `mockito-core` gets `byte-buddy` dependency
5. Adding `byte-buddy` manually to `maven_install` doesn't help
6. `mockito-core` was getting `byte-buddy` dependency with `androidx.arch.core:core-testing` in `maven_install` prior to change #908 
7. This is likely a bug in coursier due to how #908 passes classifier and type for dependency resolution


tests/custom_maven_install/regression_testing_install.json
```diff
@@ -1390,16 +1426,22 @@
...
+        "jar": "5e168368fbc250af3c79aa5fef0c3467a2d64e5a7bd74005f25d8399aeb0708d"
       },
-      "version": "2.1"
+      "version": "2.6"
     },
     "org.openjfx:javafx-base": {
       "shasums": {
         "jar": "c5084a74417a89c69a0c122fae96a4b70bf619fc3d6218ea102a4047ec85ad04",
-        "linux": "2ebf6fa2cbbe1c8b4f7780e06e97beb038f644d5ecf9f15a41c5e88ee0ef9cf1"
+        "mac": "2d8052a08fd2e5d98e1d5a16d724ea5dd02102879de20a193225f57199803983"
```
artem-zinnatullin, line 1444: This is unfortunate, but I don't have a Linux machine/image with all required Bazel deps to regenerate this for Linux and have matching hashes

WORKSPACE
```diff
@@ -257,6 +257,10 @@ maven_install(
         # Any two platforms to ensure that it doesn't work _only_ under the host operating system
         "com.google.protobuf:protoc:exe:linux-x86_64:3.21.12",
         "com.google.protobuf:protoc:exe:osx-aarch_64:3.21.12",
+        # https://github.com/bazelbuild/rules_jvm_external/issues/917
+        # androidx core-testing POM has "exclusion" for "byte-buddy" but it should be downloaded as mockito-core dependency.
+        "org.mockito:mockito-core:jar:3.3.3",
```
rdesgroppes, line 262: Since `jar` is the default, it's usually omitted.

WORKSPACE
```diff
@@ -257,6 +257,10 @@ maven_install(
...
+        "androidx.arch.core:core-testing:aar:2.1.0",
```
rdesgroppes, line 263: I gave a try with these 2 artifacts exactly as written above and indeed faced the issue that `net.bytebuddy:byte-buddy` and `net.bytebuddy:byte-buddy-agent` weren't fetched.
But, despite the theoretical equivalence, stripping out the `jar` packaging type seems to lead to the intended behavior.
```suggestion
        "org.mockito:mockito-core:3.3.3",
        "androidx.arch.core:core-testing:aar:2.1.0",
```
Looking...

rdesgroppes: @artem-zinnatullin I just filed https://github.com/artem-zinnatullin/rules_jvm_external/pull/1 you may merge to your branch or not, at your convenience.

rdesgroppes:
Closes #917.

WORKSPACE
artem-zinnatullin: Merged!

WORKSPACE
artem-zinnatullin: Yep yep, it's just a convention in our large Android monorepo to see what deps are JVM and what are Android AARs I think it's fine for purpose of this regression test

artem-zinnatullin:
Merged your fix and updated PR title and description for the fix and test to be ready for merge!

WORKSPACE
rdesgroppes: Agree.

specs.bzl
```diff
@@ -258,7 +258,7 @@ json = struct(
 #
 def _artifact_to_coord(artifact):
     classifier = (",classifier=" + artifact["classifier"]) if artifact.get("classifier") != None else ""
-    type = (",type=" + artifact["packaging"]) if artifact.get("packaging") != None else ""
+    type = (",type=" + artifact["packaging"]) if artifact.get("packaging") not in (None, "jar") else ""
```
shs96c, line 261: Nice catch

shs96c (approved):
LGTM. Thank you for the PR!

PR #920, opened 2023-06-21
Add option to extract proguard configurations in jvm_import

mauriciogg:
Some jars include proguard specs in `META-INF/proguard/` `META-INF/com.android.tools`
We need to extract these files in order to pass them correctly to R8 for android builds.
Normally R8 should detect these files automatically inside the jar, but if proguard specs are specified in `META-INF/com.android.tools` and `META-INF/proguard` the files under `META-INF/proguard` are ignored. See https://r8.googlesource.com/r8/+/refs/heads/main/src/main/java/com/android/tools/r8/R8Command.java#1394 Given that bazel uses a single deploy jar when running R8 it is likely that both directories exist and several needed configs are ignored.

see https://github.com/bazelbuild/bazel/pull/14966#issuecomment-1563483175

### DIFF
--- a/private/rules/jvm_import.bzl
+++ b/private/rules/jvm_import.bzl
@@ -8,6 +8,7 @@
 # [1]: https://github.com/bazelbuild/bazel/issues/4549
 
 load("//settings:stamp_manifest.bzl", "StampManifestProvider")
+load("//settings:extract_proguard_config.bzl", "ExtractProguardConfigProvider")
 
 def _jvm_import_impl(ctx):
     if len(ctx.files.jars) != 1:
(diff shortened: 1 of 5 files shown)


jin:
jvm_import seems like the wrong place to add processing like this. Why isn't this part of the Android rules as an action to prepare the 3rd party jars before R8 sees them, instead?

mauriciogg:
> jvm_import seems like the wrong place to add processing like this. Why isn't this part of the Android rules as an action to prepare the 3rd party jars before R8 sees them, instead?

where would this live- as a top level action of android_binary? at that point its not possible to know whats a third_party jar vs a local jar. The other approach would be an aspect, but then rules_koltin would need to be aware of it to

shs96c:
Surely `rules_android` is the right place for this? An aspect when you create the binaries there would make a lot of sense.

PR #927, opened 2023-07-02
Allow passing sensitive credentials for java_export with environment variables

### LINKED ISSUE #679: java_export should not generate publisher script with credentials in it
credentials should be kept in

- files that are read-only to the current user or 
- in memory.

The current design of `_maven_publish_impl` will generate a file with
credentials in it and the file will be uploaded to the remote cache,
thus made available for anyone who has access to it.

All these credentials should be passed at the run time of the script,
not the build time, ie. `bazel run pkg.publish -- --user=x --password=y`
and more preferably from a maven config file, eg. `.m2/settings.xml`.

Relevant topics:

- bazelbuild/bazel#498
- #80

thirtyseven:
This PR fixes #679 by allowing users to specify `maven_user` and `maven_pass` via environment variable when using `maven_publish`.

### DIFF
--- a/private/tools/java/com/github/bazelbuild/rules_jvm_external/maven/MavenPublisher.java
+++ b/private/tools/java/com/github/bazelbuild/rules_jvm_external/maven/MavenPublisher.java
@@ -65,33 +65,22 @@ public class MavenPublisher {
     "file:/", "https://", "gs://", "s3://", "artifactregistry://"
   };
 
-  private static String getArg(String[] args, String envName, int argPos) {
-    String envVal = System.getenv(envName);
-    if (envVal != null) {
-      return envVal;
-    }
-    if (argPos < args.length) {
-      return args[argPos];
-    }
-    return null;
-  }
-
   public static void main(String[] args)
       throws IOException, InterruptedException, ExecutionException, TimeoutException {
-    String repo = getArg(args, "MAVEN_REPO", 0);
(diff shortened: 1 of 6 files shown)


private/tools/java/com/github/bazelbuild/rules_jvm_external/maven/MavenPublisher.java
```diff
@@ -65,21 +65,35 @@ public class MavenPublisher {
...
 
-    boolean gpgSign = Boolean.parseBoolean(args[1]);
-    Credentials credentials = new BasicAuthCredentials(args[2], args[3]);
+    boolean gpgSign = Boolean.parseBoolean(getArg(args, "GPG_SIGN", 1));
+    Credentials credentials = new BasicAuthCredentials(
+            getArg(args, "MAVEN_USER", 2),
+            getArg(args, "MAVEN_PASSWORD", 3)
+    );
 
     List<String> parts = Arrays.asList(args[4].split(":"));
```
shs96c, line 94: If the user isn't passing in the user or password on the command line, then the coordinates are likely to be at args position 2, right? Or is it okay because the script it quoting them, so we end up with empty values?

thirtyseven: Yes, the quoted empty string in `_TEMPLATE` in `maven_publish. bzl` will ensure the arg positions are consistent even if empty.

Issue #936, opened 2023-07-25
Passing Arguments to MavenPublisher


loganasherjones:
I'm trying to publish to a private artifactory which requires a custom CA. Whenever I use `maven_publish` (via `java_export`). I get failures because the embedded java I'm using comes with its own CA that I cannot modify even with a `~/.bazelrc` that looks like the following:

```
startup --host_jvm_args=-Djavax.net.ssl.trustStore=/path/to/cacerts \
        --host_jvm_args=-Djavax.net.ssl.trustStorePassword=changeit 
```

I'm not sure why that doesn't work, but whenever I attempt to run the `.publish` target, it uses the bundled CA.

I was hoping to be able to do something like:

```
bazel run :exported_lib.publish -- --jvm_flag=-Djavax.net.ssl.trustStore=/path/to/cacerts
```

However, this doesn't work because args aren't passed through.

It's possible I'm just not doing something right, but I'd really like to be able to specify the CA I use when I'm publishing.

I'll open up a PR with a potential fix, and am open to feedback.

Thanks.

loganasherjones:
Closed with #937

tnek:
This doesn't seem to have worked for me - 

```
$ bazel run --define "maven_repo=https://my-custom-maven-repo" //target.publish -- --jvm_flag=-Djavax.net.ssl.trustStore=/Library/Java/JavaVirtualMachines/jdk-21.0.2.jdk/Contents/Home/lib/security/cacerts
```

Is there something else I"m missing to make this work?

loganasherjones:
Just to close the loop on this, I was able to get this working with 6.1 and above via:

```
bazel run \
  --define "maven_repo=https://my-custom-maven-repo \
  //path/to/target.publish \
  -- \
  --wrapper_script_flag="--jvm_flag=-Djavax.net.ssl.trustStore=/path/to/truststore"
```

PR #937, opened 2023-07-25
Add the ability to pass args to MavenPublisher

loganasherjones:
This is an attempt to address #936

### DIFF
--- a/private/rules/maven_publish.bzl
+++ b/private/rules/maven_publish.bzl
@@ -15,7 +15,7 @@ export GPG_SIGN="${{GPG_SIGN:-{gpg_sign}}}"
 export MAVEN_USER="${{MAVEN_USER:-{user}}}"
 export MAVEN_PASSWORD="${{MAVEN_PASSWORD:-{password}}}"
 echo Uploading "{coordinates}" to "${{MAVEN_REPO}}"
-{uploader} "{coordinates}" '{pom}' '{artifact}' '{classifier_artifacts}'
+{uploader} $@ "{coordinates}" '{pom}' '{artifact}' '{classifier_artifacts}'
 """
 
 def _escape_arg(str):
--- a/private/rules/maven_publish.bzl
+++ b/private/rules/maven_publish.bzl
@@ -15,7 +15,7 @@ export GPG_SIGN="${{GPG_SIGN:-{gpg_sign}}}"
 export MAVEN_USER="${{MAVEN_USER:-{user}}}"
 export MAVEN_PASSWORD="${{MAVEN_PASSWORD:-{password}}}"
 echo Uploading "{coordinates}" to "${{MAVEN_REPO}}"
-{uploader} $@ "{coordinates}" '{pom}' '{artifact}' '{classifier_artifacts}'
+{uploader} "{coordinates}" '{pom}' '{artifact}' '{classifier_artifacts}' $@
 """
 
 def _escape_arg(str):


loganasherjones:
Signed CLA

loganasherjones:
Do I need to do something additional to get this reviewed? I don't anticipate getting it accepted immediately, but I was hoping to start a conversation around this PR. 

Is the CLA check causing people not to review it? I believe I've signed it, just wanted to make sure I'm moving the ball forward.

shs96c:
Sorry for the delay getting back to you. I was on vacation for a while. Reviewing now!

private/rules/maven_publish.bzl
```diff
@@ -11,7 +11,7 @@ MavenPublishInfo = provider(
 _TEMPLATE = """#!/usr/bin/env bash
 
 echo "Uploading {coordinates} to {maven_repo}"
-{uploader} '{maven_repo}' '{gpg_sign}' '{user}' '{password}' '{coordinates}' '{pom}' '{artifact_jar}' '{source_jar}' '{javadoc}'
+{uploader} $@ '{maven_repo}' '{gpg_sign}' '{user}' '{password}' '{coordinates}' '{pom}' '{artifact_jar}' '{source_jar}' '{javadoc}'
```
shs96c, line 14: The `MavenPublisher` is very sensitive to the order in which arguments appear, and the parsing of the command line is very basic. Could you please append additional arguments to the end of the command line? Alternatively, could you please update the `MavenPublisher` to skip arguments beginning with a dash? I think that second suggestion also allows you to do what you need.

shs96c (changes requested):

private/rules/maven_publish.bzl
thirtyseven: @shs96c: @loganasherjones is trying to use this to pass arguments to the Bazel wrapper script, which currently uses [this logic](https://github.com/bazelbuild/bazel/blob/48f081f6f18277d4bbbb5a1740f9c59cecdf88fc/src/main/java/com/google/devtools/build/lib/bazel/rules/java/java_stub_template.txt#L113) to distinguish between its own flags and flags passed to the underlying java process:
```bash
# Parse arguments sequentially until the first unrecognized arg is encountered.
# Scan the remaining args for --wrapper_script_flag=X options and process them.
ARGS=()
for ARG in "$@"; do
  if [[ "$ARG" == --wrapper_script_flag=* ]]; then
    process_wrapper_argument "${ARG#--wrapper_script_flag=}" \
      || die "invalid wrapper argument '%s'" "$ARG"
  elif [[ "${#ARGS}" -gt 0 ]] || ! process_wrapper_argument "$ARG"; then
    ARGS+=( "$ARG" )
  fi
done
```

Thus, any valid wrapper script flags will not be passed to the actual publisher binary. It could be worth implementing your second suggestion as an extra safeguard, but if going with the first option, wrapper script flags will need to be prefixed with `--wrapper_script_flag=`.

loganasherjones: Welcome back @shs96c hope you had a wonderful vacation!

@thirtyseven is correct, I mostly just hacked something together as a POC to prove that I could get to the wrapper script and get me unblocked.  I'm happy to add the args at the end, and use the `--wrapper_script_flag`.  I also wouldn't mind updating `MavenPublisher`. Whatever you all think is best.

If you do want me to update `MavenPublisher`, is it really just a matter of ignoring args with `-` in their name? Would you rather me expand `MavenPublisher` to have more robust arg parsing all around?

loganasherjones: Another option I suppose is to take the part of the function from the wrapper script and copy it into the template.

loganasherjones: One more option that I was thinking of was updating `MavenPublisher` to work off of environment variables that get set in the script instead. All of these seem do-able, just let me know which one you want, and I'll get a PR with that functionality up.

loganasherjones: Ok @shs96c thanks for the review. I've moved the args to the end to prevent any unexpected behavior. Please let me know if you need any additional changes!

shs96c (approved):
LGTM. Thank you for your patience and understanding!

PR #939, opened 2023-07-28
Fix pinning error with Bzlmod

fmeum:
Fixes the `bazel run` command suggested when the lockfile is outdated and Bzlmod is used.


shs96c (approved):
This has been something that a few people have mentioned as being a little weird. Thank you for the fix!

Issue #942, opened 2023-08-23
maven_fetch javadoc artifacts are at default visibility


erickj:
Due to https://github.com/bazelbuild/bazel/issues/13681 the BUILD file generated from maven fetch does not set default visiblity to public.  The `jvm_import` and `alias` rules manually set `visibility = //visibility:public` to compensate for this.

The javadoc filegroup does not set this however, making it impossible to access.

cheister:
Fixed with https://github.com/bazelbuild/rules_jvm_external/pull/943

PR #943, opened 2023-08-23
Sets javadoc filegroup targets generated from maven_install to include visibility = //visiblity:public (#942)

erickj:
Verified manually:

```
filegroup(
        name = "aopalliance_aopalliance_javadoc",
        srcs = ["None"],
        tags = ["javadoc"],
)
```


shs96c (approved):
I can see why this would be useful :) My apologies for the slow review, but LGTM.

PR #944, opened 2023-08-24
Fix CODEOWNERS only requesting reviews from @cheister

jin:
Only @cheister was getting owners review because GitHub doesn't support multiline aggregation and took the last row :-)

@Wyverald let me know if you'd still like to be in RJE's CODEOWNERS.


shs96c (approved):
I am amazed it all needs to be on one line. LGTM

PR #952, opened 2023-09-01
Add compile-only deps to javadocs classpath

fmeum:
javadoc may have to inspect compile-only dependencies.

Also removes a line that only added elements to a depset that are already contained in this depset.


fmeum:
@shs96c Not directly related to this PR, but it looks like the `additional_dependencies` attribute of the `javadoc` rule is a no-op.

shs96c (approved):
LGTM. Thanks for the catch.

PR #955, opened 2023-09-06
Fix `testonly` attribute handling

meteorcloudy:
Fix the following error when using `test_only` in Bzlmod
```
ERROR: Traceback (most recent call last):
	File "/private/var/tmp/_bazel_pcloudy/829441223e9fec5a5a2e3d1dd743fdf0/external/rules_jvm_external~5.2/private/extensions/maven.bzl", line 304, column 52, in _maven_impl
		artifacts_json = [_json.write_artifact_spec(a) for a in artifacts]
	File "/private/var/tmp/_bazel_pcloudy/829441223e9fec5a5a2e3d1dd743fdf0/external/rules_jvm_external~5.2/specs.bzl", line 230, column 38, in _artifact_spec_to_json
		"\", \"version\": \"" + artifact_spec["version"] + "\""
Error: unsupported binary operation: string + bool
```

### DIFF
--- a/private/extensions/maven.bzl
+++ b/private/extensions/maven.bzl
@@ -206,7 +206,7 @@ def _maven_impl(mctx):
                 to_add.update({"neverlink": artifact.neverlink})
 
             if artifact.testonly:
-                to_add.update({"version": artifact.testonly})
+                to_add.update({"testonly": artifact.testonly})
 
             if artifact.exclusions:
                 artifact_exclusions = []


meteorcloudy:
Is there any way we can improve the quality of Bzlmod support and have more test coverage perhaps?

meteorcloudy:
Also I noticed `"fetch_sources": attr.bool(default = True),` in Bzlmod: https://github.com/bazelbuild/rules_jvm_external/blob/5.2/private/extensions/maven.bzl#L41

And eventually it's an OR of all maven installs. Is there any reason for this attribute to be True by default in Bzlmod?

shs96c:
We set the default to be `True` because the common case is for people to use an IDE for working on their code, and having the sources available makes that experience significantly nicer. We can change it to `False` if necessary, and it's probably best if we take the value from the root module only.

PR #959, opened 2023-09-15
Support so/dylib/dll packaging types

dzbarsky:
I think it would be a bit nicer to set up the data dependencies in the generated rules, but I wasn't able to get it working when trying a combination of cc_import, java_library, etc. The downside of doing that is we wouldn't be able to use `select` to depend on just the platform-specific jar that we want. 

If anyone has ideas how to do better, let me know!

### DIFF
--- a/private/dependency_tree_parser.bzl
+++ b/private/dependency_tree_parser.bzl
@@ -96,8 +96,11 @@ def _generate_target(
     elif packaging == "aar":
         import_rule = "aar_import"
         jar_versionless_target_labels.append(target_label)
+    elif packaging in ["dylib", "so", "dll"]:
+        import_rule = "java_library"
     else:
         fail("Unsupported packaging type: " + packaging)
+
     target_import_string = [import_rule + "("]
 
     # 2. Generate the target label.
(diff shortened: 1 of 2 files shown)


private/dependency_tree_parser.bzl
```diff
@@ -169,6 +170,24 @@ def _generate_imports(repository_ctx, dependencies, explicit_artifacts, neverlin
...
+                is_dylib = True
+                jar_versionless_target_labels.append(target_label)
+                dylib = simple_coord.split(":")[-1] + "." + packaging
+                all_imports.append(
+                    """
+genrule(
+    name = "{dylib}_extension",
+    srcs = ["@{repository}//file"],
+    outs = ["{dylib}"],
+    cmd = "cp $< $@",
```
dzbarsky, line 184: these copy rules exist to bring the artifacts into the @maven// top-level, which lets the user have a single `java.library.path` flag

shs96c:
@dzbarsky, my apologies for how long it's taking me to review this PR: I've traveled and have some work commitments, but I will get to this as soon as possible.

dzbarsky:
> @dzbarsky, my apologies for how long it's taking me to review this PR: I've traveled and have some work commitments, but I will get to this as soon as possible.

No worries, this is not long at all compared to other projects I've contributed to in the past :) Thanks for the update!

dzbarsky:
@shs96c Any idea when you'll have a chance to look at this one?

shs96c (approved):
LGTM. Thank you for your patience! It's very much appreciated!

PR #963, opened 2023-09-22
jvm_import: Include sha256 in tags

illicitonion:
This may help folks to do things like generate sbom files by looking at the attributes of jvm_import targets.


shs96c (approved):
LGTM

PR #964, opened 2023-09-22
Add explicit tag for repository used

illicitonion:
This may help folks to do things like generate sbom files by looking at the attributes of jvm_import targets.


shs96c (approved):
It would be nice if there was a test for this....

PR #969, opened 2023-10-03
Treat empty files in pinning as if they weren't available

gergelyfabian:
We have a strange behavior that in our artifactory server source jars happen to sometimes appear as a 0-byte jar.
We'd like to treat them as if they weren't there.

There is a legitimate question here though, is it legal to have a 0-byte jar file? If this is a legal case, then I'd be happy to change this PR to hide such a feature behind a switch (to be able to remove such jars from the results of coursier), otherwise will leave this on a fork.


gergelyfabian:
> There is a legitimate question here though, is it legal to have a 0-byte jar file? If this is a legal case, then I'd be happy to change this PR to hide such a feature behind a switch (to be able to remove such jars from the results of coursier), otherwise will leave this on a fork.

I've hidden the feature behind a feature flag.

private/extensions/maven.bzl
```diff
@@ -170,6 +171,7 @@ def _maven_impl(mctx):
     # - strict_visibility_value: a string list. Build will fail is duplicated and different.
     # - use_starlark_android_rules: bool. A logical OR over all `use_starlark_android_rules` for all `install` tags with the same name.
     # - version_conflict_policy: string. Fails build if different and not a default.
+    # - ignore_empty_files: Treat jars that are empty as if they were not found.
```
cheister, line 174: Do you mind adding something to the README.md for this option?

cheister:
Thanks for the PR. 

I'm sort of surprised that maven repositories will serve 0 byte jars.  Have you only seen this with source jars or with non-source jars as well?

private/extensions/maven.bzl
gergelyfabian: Sure, I could do that.

gergelyfabian:
> Thanks for the PR.
> 
> I'm sort of surprised that maven repositories will serve 0 byte jars. Have you only seen this with source jars or with non-source jars as well?

Yes, it's unknown why these empty jars are appearing. We have an artifactory where this is appearing, when downloading artifacts from a 3rd party artifactory. It's only appearing for source jars.

private/extensions/maven.bzl
gergelyfabian: Done.

cheister (approved):

gergelyfabian:
Resolved conflicts and fixed one bug.

shs96c (approved):
LGTM. Thank you for your patience with the review!

gergelyfabian:
I'm not sure about the test failures... They fail e.g. due to:

```
Error: 'JavaInfo' value has no field or method 'transitive_runtime_deps'
...
Error: 'JavaInfo' value has no field or method 'transitive_deps'
...
error: Source option 7 is no longer supported. Use 8 or later.
error: Target option 7 is no longer supported. Use 8 or later.
```

I think this is not related to this PR. In fact master is also failing.

shs96c:
The build appears to be failing for unrelated issues (`JavaInfo` no longer having `transitive_runtime_deps`), so merging.

gergelyfabian:
> The build appears to be failing for unrelated issues (`JavaInfo` no longer having `transitive_runtime_deps`), so merging.

Rebased with master after the build failures were fixed there.

PR #967, opened 2023-09-29
Load JavaInfo and java_common from rules_java

comius:
This prepares rules_jvm_external for https://github.com/bazelbuild/bazel/issues/19455

PR #971, opened 2023-10-05
Remove "jetify" functionality

shs96c:
This has been deprecated for a while, and the list of dependencies that required jetifying is old. It is hoped that people will have updated their own dependencies by now, rather than requiring this tooling to handle that case.


cheister:
Seems fine, just need to get the last tests passing
```
Error: file '//private:dependency_tree_parser.bzl' does not contain symbol 'JETIFY_INCLUDE_LIST_JETIFY_ALL'
```

cheister:
Is the plan to merge this PR before https://github.com/bazelbuild/rules_jvm_external/pull/967 ?

shs96c:
Yes, I think it makes sense to get this in before #967.

shs96c:
@cheister, this is ready for review again

PR #972, opened 2023-10-05
Bump minimum version of Java to 11

shs96c:
The new Java LTS release (version 21) has shipped, so targeting the LTS-2 release as a minimum should be viable.


cheister:
Do we need to update the `rbe-ubuntu-16-dot-04-openjdk-8` test to java 11 as well?

shs96c:
@jin could you please tweak the BuildKite config so that we no longer require Java 8 support?

shs96c:
No need to tweak the build. I figured it out.

shs96c:
It would be nice if we could rebuild the jars per-release and add them as artifacts that we download as `http_file`s. Checking them into the tree made sense when they didn't change often, but we now seem to be doing quite a few builds.

The alternative thing we could do is push each of the binaries to Maven Central as releases using `java_export` and then we can download them using a `maven_install` with a lock file.

PR #977, opened 2023-10-13
Update rules_kotlin to stop issue with 'JavaInfo.transitive_deps' with

katre:
latest bazel releases.


katre:
Closing this since it doesn't fix the problem (see https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/3376#018b253f-b092-48db-a561-ed8968c24489 for an example).

Does this _really_ need to support Bazel 4.0.0?

PR #986, opened 2023-10-25
Migrate to using `bzlmod` for our own builds

shs96c:


jin:
I'm seeing these after checking out this PR and running `bazel test //...` locally:

```
ERROR: no such package '@android_gmaven_r8//jar': The repository '@android_gmaven_r8' could not be resolved: Repository '@android_gmaven_r8' is not defined
ERROR: /usr/local/google/home/jingwen/.cache/bazel/_bazel_jingwen/8484bc4fff18ee4a905b69a9ddb0e143/external/androidsdk/BUILD.bazel:13:25: no such package '@android_gmaven_r8//jar': The repository '@android_gmaven_r8' could not be resolved: Repository '@android_gmaven_r8' is not defined and referenced by '@androidsdk//:d8_jar_import'

//tests/integration:testonly_artifacts_test                              FAILED in 0.1s
  /usr/local/google/home/jingwen/.cache/bazel/_bazel_jingwen/8484bc4fff18ee4a905b69a9ddb0e143/execroot/_main/bazel-out/k8-fastbuild/testlogs/tests/integration/testonly_artifacts_test/test.log
//tests/integration:version_interval_deps_test                           FAILED in 0.1s
  /usr/local/google/home/jingwen/.cache/bazel/_bazel_jingwen/8484bc4fff18ee4a905b69a9ddb0e143/execroot/_main/bazel-out/k8-fastbuild/testlogs/tests/integration/version_interval_deps_test/test.log
//tests/integration/java_export:check-excluded-packages-are-excluded     FAILED in 0.3s
  /usr/local/google/home/jingwen/.cache/bazel/_bazel_jingwen/8484bc4fff18ee4a905b69a9ddb0e143/execroot/_main/bazel-out/k8-fastbuild/testlogs/tests/integration/java_export/check-excluded-packages-are-excluded/test.log
//tests/unit/manifest_stamp:diff_manifest_test                           FAILED in 0.1s
  /usr/local/google/home/jingwen/.cache/bazel/_bazel_jingwen/8484bc4fff18ee4a905b69a9ddb0e143/execroot/_main/bazel-out/k8-fastbuild/testlogs/tests/unit/manifest_stamp/diff_manifest_test/test.log
```

Presumably you're seeing the same?

shs96c:
That's exactly what I'm seeing

jin:
I just pushed https://github.com/bazelbuild/rules_jvm_external/pull/986, which seem to make the Android related errors go away, but the tests still need work.

private/extensions/maven.bzl
```diff
@@ -37,7 +37,7 @@ _install = tag_class(
 
         # What do we fetch?
         "fetch_javadoc": attr.bool(default = False),
-        "fetch_sources": attr.bool(default = True),
+        "fetch_sources": attr.bool(default = False),
```
jin, line 40: Was this enabled by default by mistake?

tests/bazel_run_tests.sh
```diff
@@ -5,6 +5,13 @@
 #
 # Add a new test to the TESTS array and send all output to TEST_LOG
 
+function force_bzlmod_lock_file_to_be_regnerated() {
```
jin, line 8: typo in "regenerated"

jin (approved):
Thank you very much!

private/extensions/maven.bzl
shs96c: The simple answer is that this sets the behaviour in the `bzlmod` world to be the same as what we have in the non-bzlmod world.

The slightly longer answer is that we use a logical OR to figure out whether or not to fetch sources. If we set the default to `True`, by default we include all sources, and it would only take one transitive dep in (eg). the default `maven` namespace not to be aware of this to make it something users can't turn off.

The really hand-wavy answer is that I think leaving it set to `True` broke some tests and this was the simplest fix.

The correct answer is to say that if a namespace is specified more than once, and one of those times is the root module, we should do what the root module asks, and only if that's not the case do a logical OR (I think).

tests/bazel_run_tests.sh
shs96c: Fixed!

PR #992, opened 2023-11-06
format files with bazel run //scripts:format

cheister:
Just cleaning up formatting and deprecated imports.


shs96c (approved):
Thank you!

PR #994, opened 2023-11-16
Allow the "raw" version of overridden targets to be used

shs96c:
This makes the pattern of overriding specific targets by adding additional information far simpler. For example, if one wants to add a dep on guava to the OpenTelemetry SDK jar, one could use:

```
maven_install(
   ...
   override_targets = {
       "io.opentelemetry:opentelemetry-sdk": "@//:otel-sdk"
   }
```

And then in the `BUILD.bazel` file:

```
java_library(
    name = "otel-sdk",
    exports = [
      "@maven//:raw_io_opentelemetry_opentelemetry_sdk",
    ],
    deps = [
      "@maven//:com_google_guava_guava",
    ],
)
```

Once this is done, should the version of the dependency be updated, there's no need to update anything else in the build.


jin (approved):
Mind updating https://github.com/bazelbuild/rules_jvm_external#overriding-generated-targets with this new target?

PR #997, opened 2023-11-21
feat: Add @name//:pin target to pinned repository for convenience

blorente:
Add `@maven//:pin` (an alias to `@unpinned_maven//:pin`) for convenience.

Note that this relies strongly on @maven existing, so `fail_if_repin_required` should be honored at all times (and therefore this PR maybe depends on https://github.com/bazelbuild/rules_jvm_external/pull/996)

### DIFF
--- a/coursier.bzl
+++ b/coursier.bzl
@@ -74,6 +74,14 @@ sh_binary(
 )
 """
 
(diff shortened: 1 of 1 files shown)


shs96c (approved):
LGTM. Thank you for the PR!

PR #998, opened 2023-11-22
Pin CI builds to the last release of each major version for Bazel 4, 5, and 6

shs96c:


jin (approved):

jin:
CI is trying to use --bes_instance_name for the 4.2.2 build, which presumably doesn't exist yet (I think it uses --project_id instead).

This is already a problem @ HEAD: https://buildkite.com/bazel/rules-jvm-external/builds/3776#018bf45f-f7e7-47fb-a4e3-45badde20147 - maybe let's omit the 4.2.2 build for now?

https://github.com/bazelbuild/continuous-integration/issues?q=is%3Aissue+bes_instance_name

cc @fwe

jin:
Filed https://github.com/bazelbuild/continuous-integration/issues/1810

shs96c:
Commenting out the 4.x builds. Given that Bazel 7 is around the corner, we might want to consider our position on supporting Bazel 4 moving forwards, but that's a different topic entirely :)

jin (approved):

PR #1000, opened 2023-11-22
feat(maven_install): Customizable repin instructions

blorente:
Depends on https://github.com/bazelbuild/rules_jvm_external/pull/996, only review the last commit.

Add customizable instructions to re-pin the repository if required. Many people have wrapper scripts for keeping dependencies up to date, and would like to point users to that instead of the default.


shs96c (approved):
LGTM

PR #1002, opened 2023-11-24
Publish java_export target without sources jar

### LINKED ISSUE #1001: Publish java_export target without sources jar
Is there a way to publish a `java_export` target without its sources jar?

Locally I patched the current version so that targets can be tagged with `no-sources`, similar to `no-javadocs`, but wonder if there's another solution. If you like I can provide that change as PR.

bertschneider:
Tagging a java_export target with `no-sources` prevents publishing of its sources jar.

Closes #1001


shs96c (approved):
LGTM. Thanks for explaining the use case!

PR #1006, opened 2023-12-06
Allow users to specify a specific version of a dependency should be used

shs96c:
Normally, one would use a maven version range, such as `[1.2.3]`, which should give exactly the same result. However, if the specified version is lower than that required by a transitive dep, then `coursier` will fail the resolution with a complaint about a version mismatch. This diff avoids that problem.


jin (commented):
Seems duplicative with https://github.com/bazelbuild/rules_jvm_external?tab=readme-ov-file#resolving-user-specified-and-transitive-dependency-version-conflicts?

shs96c:
The problem with `pinned` is that it will force _all_ values in the `maven_install` coordinates to be locked in place, whereas this PR allows people to rely on `coursier`'s normal resolution algorithm to be used except for a specific coordinate. Though this section of the docs would be a good place to document this.

jin:
Yes, documenting the distinction between this and `version_conflict_policy` would be ideal.

coursier.bzl
```diff
@@ -763,8 +763,17 @@ def make_coursier_dep_tree(
...
+    forced_versions = []
     for a in artifacts:
-        artifact_coordinates.append(utils.artifact_coordinate(a))
+        coordinates = utils.artifact_coordinate(a)
+
+        # Special-case handling of specific versions that are requested. Without this, coursier will claim that
+        # if a lower version of a dependency is requested by a transitive dep than what is specified here, a
+        # version mismatch will occur, and this will fail the resolution.
+        if a.get("force_version", False):
+            forced_versions.append(coordinates)
```
jin, line 774: What happens with multiple `maven.artifact` with the same `group:artifact` coordinates enabling `force_version`?

shs96c: I've no idea. I don't expect people will use this much, so I'm okay with this being undefined behaviour until the first bug report comes in and tells us what people expect.

PR #1008, opened 2023-12-13
Updates for Bazel 7.

jin:
Bazel CI has changed its default version to the 7, the latest major version.

This change contains:

__Fixes for bzlmod related test breakages__: bzlmod introduces the `@@` canonical label prefix, and `~` module name separator. Our tests look for these.

__Fix for JavaInfo changes__. `transitive_runtime_deps` was renamed to `transitive_runtime_jars`.

__Simplifying windows/unix diff tests__. Now the tests removes ^M line endings from the windows outputs, so the diffs work on a unix-based golden regardless of execution platform.

__Bazel CI config updates__. I've updated presubmit.yml to specifically test `//tests/integration/override_targets:override_targets` in the `Android tests` job, and remove the negative targets from the other configurations. This works because I've tagged that test as `manual`, so `//...` doesn't pick it up.

__rules_android__: this is now a dev dependency. 

__bzlmod-ified the failing examples__: these used io_bazel_rules_kotlin with `http_archive`, so taking this chance to migrate that to rules_kotlin and `bazel_dep`.


gergelytraveltime:
Thanks @jin!

gergelyfabian:
Thanks @jin!

jin:
All tests are green.

shs96c (approved):
LGTM. I like the fix for the golden files. I'd been wondering how to do that.

Issue #1016, opened 2024-01-04
generate_compat_repositories in maven module extension doesn't generate target names that provide "compat"-ibility with older repositories


ekohlwey:
I am attempting to use the grpc kotlin plugin via an extension. This is the code of my extension:

```python
load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")

def grpc_kotlin():
    grpc_kotlin_version = "1.4.1"
    http_archive(
        name = "com_github_grpc_grpc_kotlin",
        urls = ["https://github.com/grpc/grpc-kotlin/archive/refs/tags/v%s.tar.gz" % grpc_kotlin_version],
        strip_prefix = "grpc-kotlin-%s" % grpc_kotlin_version,
    )
    grpc_java_version = "1.60.1"
    http_archive(
        name = "io_grpc_grpc_java",
        urls = ["https://github.com/grpc/grpc-java/archive/refs/tags/v%s.tar.gz" % grpc_java_version],
        strip_prefix = "grpc-java-%s" % grpc_java_version,
    )
    protobuf_version = "25.1"
    http_archive(
        name = "com_google_protobuf",
        urls = ["https://github.com/protocolbuffers/protobuf/releases/download/v25.1/protobuf-%s.tar.gz" % protobuf_version],
        strip_prefix = "protobuf-%s" % protobuf_version,
    )
    rules_ruby_version = "0.6.0"
    http_archive(
        name = "rules_ruby",
        strip_prefix = "rules_ruby-%s" % rules_ruby_version,
        urls = [
            "https://github.com/protocolbuffers/rules_ruby/archive/v%s.zip" % rules_ruby_version,
        ],
    )


def _grpc_kotlin_repositories(ctx):
    grpc_kotlin()

grpc_kotlin_dependencies = module_extension(
    implementation = _grpc_kotlin_repositories,
)
```

This is the relevant section of my `MODULE.bazel` file:

```python
maven.install(
    artifacts = [
        "com.google.errorprone:error_prone_annotations:2.3.2",
        "com.squareup:kotlinpoet:1.15.3",
        "com.google.guava:guava:31.1-jre",
        "com.google.guava:failureaccess:1.0.2",
    ],
    generate_compat_repositories = True,
)

use_repo(
    maven,
    "maven",
    "com_google_guava_guava",
    "com_google_j2objc_j2objc_annotations",
    "com_google_code_findbugs_jsr305",
[...]

PR #1017, opened 2024-01-04
Remove the mangled repository name from the target created by generate_compat_repositories

ekohlwey:
This fixes the issue described in #1016 by removing the mangled non-root-repo name from the targets created by generate_compat_repositories.

### DIFF
--- a/private/compat_repository.bzl
+++ b/private/compat_repository.bzl
@@ -29,7 +29,7 @@ def _compat_repository_impl(repository_ctx):
     repository_ctx.file(
         "BUILD",
         _ROOT_BUILD.format(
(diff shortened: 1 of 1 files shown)


shs96c:
Could you please check whether this PR is still needed now we've landed #986? On current `HEAD` of `master`, I see this when querying the compat repositories when using the `bzlmod`-based build

```
bazel query "@com_google_guava_guava//:*"
@com_google_guava_guava//:BUILD
@com_google_guava_guava//:com_google_guava_guava
```

ekohlwey:
Looks like its no longer needed on the new master. Thanks!

PR #1027, opened 2024-01-05
Load `JavaInfo` and `java_common` from `rules_java`

shs96c:
This prepares rules_jvm_external for bazelbuild/bazel#19455

Closes #967 which has been left in limbo.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -12,17 +12,19 @@ bazel_dep(
     version = "0.0.8",
 )
 bazel_dep(
+    name = "rules_java",
+    version = "7.3.2",
+)
+bazel_dep(
     name = "rules_kotlin",
     version = "1.9.0",
 )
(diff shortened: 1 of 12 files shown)


MODULE.bazel
```diff
@@ -11,18 +11,20 @@ bazel_dep(
     name = "platforms",
     version = "0.0.8",
 )
+bazel_dep(
+    name = "rules_java",
+    version = "7.3.2",
```
fmeum, line 16: I would recommend staying on a lower version as this one requires Bazel 7. 7.2.0 still works with Bazel 6.4.0.

shs96c: We already use `use_repo_rule`, so we're already expecting Bazel 7 when people use `bzlmod`

jin (approved):

PR #1029, opened 2024-01-10
Place-holder for removed jetifier functionality

shs96c:
Some rulesets (eg. `contrib_rules_jvm`) have "frozen" their `rules_jvm_external` dependencies to avoid needing consumers of those rulesets from needing to run a resolution. Provide a place-holder `jetifier.bzl` file so that those rulesets don't need to jump to the latest `rules_jvm_external` immediately.

### DIFF
--- a/private/rules/jetifier.bzl
+++ b/private/rules/jetifier.bzl
@@ -1,0 +1,6 @@
+
+def jetify_aar_import(**kwargs):
+    fail("This functionality is no longer supported")
+
+def jetify_jvm_import(**kwargs):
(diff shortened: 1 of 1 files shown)


illicitonion (approved):

private/rules/jetifier.bzl
```diff
@@ -0,0 +1,6 @@
+
```
jin, line 1: Mind leaving a comment on why this file exists?

jin (approved):

PR #1031, opened 2024-01-11
fix: handle coursier_fetch repository name with bzlmod

wade-arista:
With --bzlmod_enable, the "unpinned" repository name no longer starts with `unpinned_`, but `rules_jvm_external~override~maven~unpinned_` instead.

Handle either version wherever we're checking for `unpinned_`. Without this, using bzlmod, it's not possible to use the bazel downloader (and associated `--experimental_downloader_config` flag) for offline operation.

### DIFF
--- a/coursier.bzl
+++ b/coursier.bzl
@@ -182,6 +182,30 @@ def _execute_with_argsfile(
 def _normalize_to_unix_path(path):
     return path.replace("\\", "/")
 
+def _should_use_unsafe_shared_cache(repository_ctx):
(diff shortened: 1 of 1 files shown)


wade-arista:
Hi folks, thanks for having a look at this PR. If it looks OK, that's great, otherwise I thought it might be the easiest way to start a discussion on `rules_jvm_external` with the lock file and bzlmod enabled.

wade-arista:
I've been trying to set up a min repro and seeing different behaviour, so I'll close this until I can figure out that.

PR #1034, opened 2024-01-15
Ensure checking for zero-length or conflict markers in lock file works in bzlmod

shs96c:


private/coursier_utilities.bzl
```diff
@@ -89,3 +89,10 @@ def escape(string):
 def is_maven_local_path(absolute_path):
     # Return whether or not the provided absolute path corresponds to maven local
     return absolute_path and len(absolute_path.split(".m2/repository")) == 2
+
+def contains_git_conflict_markers(file_name, lock_file_content):
+    for line in lock_file_content.splitlines():
+        if line.startswith("<<<<<<<"):
```
jin, line 95: Consider checking for `=======` and `>>>>>>>` as well since I've most definitely left them in by mistake before.

private/coursier_utilities.bzl
```diff
@@ -89,3 +89,10 @@ def escape(string):
...
+            print("Conflict markers detected in lock file {}. You should reset the file and repin your dependencies".format(file_name))
+            return True
```
jin, line 97: Should we just fail? This seems too permissive / lax.

jin (approved):

private/coursier_utilities.bzl
shs96c: Sure. No problem.

I think the workflow for most people is:

1. `git pull`
2. See that there's a conflict
3. Run `@unpinned_maven//:pin`

In that case, this check is sufficient, but it's no problem to add the other checks too.

private/coursier_utilities.bzl
shs96c: This allows us to repin through a corrupted lock file. I'll add comment.

Issue #1028, opened 2024-01-10
Can't install maven dependencies with a top-level "build" directory on case-insensitive filesystems


alex-shinn:
Currently it's impossible to import protovalidate (https://mvnrepository.com/artifact/build.buf/protovalidate/0.1.9) on OSX because that repository contains a top-level "build" directory as part of the package path, which conflicts with the generated "BUILD" file.  Trying results in the error:

```
ERROR: An error occurred during the fetch of repository 'build_buf_protovalidate_0_1_9':
   Traceback (most recent call last):
	File "/private/var/tmp/_bazel_tkanda/7710aed0b0cdc40a869d8ca35bc8abd7/external/bazel_tools/tools/build_defs/repo/http.bzl", line 183, column 13, in _http_file_impl
		ctx.file("file/BUILD", _HTTP_FILE_BUILD.format(downloaded_file_path))
Error in file: java.io.IOException: /private/var/tmp/_bazel_tkanda/7710aed0b0cdc40a869d8ca35bc8abd7/external/build_buf_protovalidate_0_1_9/file/BUILD (Directory not empty)
```

I believe this stems from the hard-coded "BUILD" filename in: https://github.com/bazelbuild/rules_jvm_external/blob/80ce39ad50ed49c00ca7cd26e66107fbbde98483/coursier.bzl#L622

It should be fine to just replace the filename with BUILD.bazel (see https://github.com/bazelbuild/bazel/issues/4517).  Alternatively, we could do this only on OSX, or more conservatively only if a given flag is specified.

alex-shinn:
Actually, it looks like this error is coming from `_http_file_impl` in core bazel, though we probably need to fix both places.

cheister:
Interesting, I'm not surprised that the `build` package is creating a conflict.  

I realize it's not ideal but I think you could workaround this for now by downloading the jar in your WORKSPACE with something like
```
http_file(
    name = "build_buf_protovalidate_0_1_9",
    sha256 = "d9c141bb6172c3a42283bde148f369772014d79bffc0f243a1da73b9da84e213",
    url = "https://repo.maven.apache.org/maven2/build/buf/protovalidate/0.1.9/protovalidate-0.1.9.jar",
)
```

and then set up a jvm_import rule in a BUILD file

```
load("@rules_jvm_external//private/rules:jvm_import.bzl", "jvm_import")

jvm_import(
    name = "build_buf_protovalidate",
[...]

Issue #970, opened 2023-10-03
V2 lock file does not support pinning artifacts with a `groupId` that starts with `build.`


andrew-lockwood:
When upgrading from v4.5 to v5.3 (and thus from lock file v1 to v2), `http.bzl` started throwing IOExceptions with the message `Directory not empty` whenever `build.buf.protoc-gen-validate:pgv-java-grpc` was included as a given targets dependency. After a bit of digging, the error seems to be due to setting  `downloaded_file_path` to [groupId as directory](https://maven.apache.org/repository/layout.html#maven2-repository-layout) when generating pinned dependencies:

```python
http_file(
    name = "build_buf_protoc_gen_validate_pgv_java_grpc_0_6_13",
    sha256 = "b74f3de8648ff9545cf22046eea784b8017bd4ad5409866c581400df4520f1dc",
    netrc = "../maven/netrc",
    urls = ["https://repo.maven.apache.org/maven2/build/buf/protoc-gen-validate/pgv-java-grpc/0.6.13/pgv-java-grpc-0.6.13.jar"],
    downloaded_file_path = "build/buf/protoc-gen-validate/pgv-java-grpc/0.6.13/pgv-java-grpc-0.6.13.jar",
)

# For reference, `http_file` tries to create the following internal directory structure
#
# build_buf_protoc_gen_validate_pgv_java_grpc_0_6_13/
# ├── file/
# │   ├── build/buf/protoc-gen-validate/pgv-java-grpc/0.6.13/pgv-java-grpc-0.6.13.jar
# │   └── BUILD
# └── WORKSPACE        
```

I currently have a patch that appends `v1` to `downloaded_file_path` in [coursier.bzl](https://github.com/bazelbuild/rules_jvm_external/blob/5.3/coursier.bzl) and that seems to resolve the issue. From what I can tell, the path inside `http_file` is purely an implementation detail and doesn't seem to leak any functionality elsewhere.  I'm willing to work on a PR and shepherd it through if that fix sounds good. 

```patch
diff --git a/coursier.bzl b/coursier.bzl
index 061b067..ebbfe16 100644
--- a/coursier.bzl
+++ b/coursier.bzl
@@ -463,7 +463,7 @@ def _pinned_coursier_fetch_impl(repository_ctx):
         http_files.append("        urls = %s," % repr(
             [remove_auth_from_url(url) for url in artifact["urls"]],
         ))
[...]

PR #1037, opened 2024-01-16
Fix can't install maven dependencies with top-level build directory

cheister:
https://github.com/bazelbuild/rules_jvm_external/issues/1028

Just change the downloaded file path so it does not create a `build` directory when using http_file.


shs96c:
Couldn't we just rename all our `BUILD` files to `BUILD.bazel`? That would avoid needing to do this work, and seems to be what is now common practice in the bazel OSS community (looking at the rulesets in `bazel-contrib`?

cheister:
> Couldn't we just rename all our `BUILD` files to `BUILD.bazel`? That would avoid needing to do this work, and seems to be what is now common practice in the bazel OSS community (looking at the rulesets in `bazel-contrib`?

This isn't a problem with a BUILD file we write, it's one generated by [http_file](https://github.com/bazelbuild/bazel/blob/master/tools/build_defs/repo/http.bzl#L163)  and i[t checks if there is already a BUILD file in the path](https://github.com/bazelbuild/bazel/blob/master/tools/build_defs/repo/http.bzl#L176), on a case-insensitive file system it compares the `build/` directory to `BUILD` and fails.  So the workaround is to not have `build/` in the downloaded_file_path passed to http_file

shs96c (approved):
LGTM. Thanks for the clarification. I'd missed that entirely

cheister:
Should also address https://github.com/bazelbuild/rules_jvm_external/issues/970

PR #1038, opened 2024-01-16
Stop using doctoc for our docs

shs96c:
GitHub already provides this functionality, so there's no need for us to do this too.


jin (approved):
`+2 −1,114`. Beautiful. Thanks!

PR #1039, opened 2024-01-16
Since we removed rules_js, we can remove the bazelrc setting for it

cheister:
After https://github.com/bazelbuild/rules_jvm_external/pull/1038, we no longer need this .bazelrc setting


shs96c (approved):
Thank you!

PR #1040, opened 2024-01-16
Allow the root module to always set the lock file

### LINKED ISSUE #1035: With `bzlmod` if the root module and a dependency both specify a lock file, use the root module's
Rulesets like `protobuf` need to add to the default namespace (`maven`) when using `bzlmod`. For their own tests, they'd like to use a lock file, but if they specify one, then currently `rules_jvm_external` will fail the build, since there will be conflicting lock files if the user also uses a lock file in the default namespace.

To work around this, if there's a lock file in the root module for a particular namespace, _always_ use that in preference to the one from another ruleset. If two rulesets both want a lock file that _isn't_ in the root namespace, then fail the build.

shs96c:
It's nice to have a lock file in your project, but if you're also a ruleset that others depend on (eg. `protobuf`) the root module's lock file should be the one that is used, not yours.

Fixes #1035


jin (approved):
Do/can we have a test for this?

jin (approved):
thanks!

Issue #1015, opened 2023-12-29
 ERROR: no such target '//src/test:greeter_test_app': target 'greeter_test_app' not declared in package 'src/test' 


sgowroji:
https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/3556#018cb373-7013-4915-9d3f-29a0c2c3f313

Platform : [rules_jvm_external - examples (Android instrumentation test example on ](https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/3556#018cb373-7010-46cf-8042-9d17becdce25):mac:, [rules_jvm_external - examples (Android instrumentation test example on ](https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/3556#018cb373-7013-4915-9d3f-29a0c2c3f313):windows:

steps:
```
git clone ttps://github.com/bazelbuild/rules_jvm_external.git
git reset 1dd7329412843a45d3694b0774ddcad540e79d1c --hard
export USE_BAZEL_VERSION=30452433de6f98fa2cdd994688edf71066ee2e84
bazel build  -- //src/main:greeter_app //src/test:greeter_test_app
```

Logs:

```
(03:08:17) ERROR: C:/b/bk-windows-knsj/bazel-org-repo-root/rules_jvm_external/examples/android_instrumentation_test/src/test/BUILD:9:1: name 'android_instrumentation_test' is not defined
(03:08:17) ERROR: Skipping '//src/test:greeter_test_app': no such target '//src/test:greeter_test_app': target 'greeter_test_app' not declared in package 'src/test' defined by C:/b/bk-windows-knsj/bazel-org-repo-root/rules_jvm_external/examples/android_instrumentation_test/src/test/BUILD
(03:08:17) ERROR: no such target '//src/test:greeter_test_app': target 'greeter_test_app' not declared in package 'src/test' defined by C:/b/bk-windows-knsj/bazel-org-repo-root/rules_jvm_external/examples/android_instrumentation_test/src/test/BUILD

```

CC Greenteam @SalmaSamy

mai93:
I think this is due to https://github.com/bazelbuild/bazel/commit/2eda429be6523390e75bf0fb55ece62defd08da0. @ajsinclair, is there another version of this rule that can be used?

ajsinclair:
[...]

PR #1041, opened 2024-01-17
Delete android_instrumentation_test example

jin:
Fixes https://github.com/bazelbuild/rules_jvm_external/issues/1015 - `native.android_instrumentation_test` has been deleted from Bazel HEAD and there isn't a planned replacement rule.


jin:
CI failures are WAI because they're not picking up the new bazelci config.

shs96c (approved):
Lovely. LGTM

PR #1044, opened 2024-01-18
Fix mirror_coursier script

jin:
The script was referencing a location in `$output_base/execroot/external` that no longer existed after the switch to bzlmod. The version number is also no longer in the pathname, so load it from versions.bzl directly.


shs96c (approved):
Looks like a reformat at this point, but it's a nice one.

PR #1049, opened 2024-01-25
`maven_export` support for bundling AARs

sugarmanz:
This PR adds (limited) support for bundling AARs for distribution with `maven_publish`. It optimistically uses an `.aar` from `AndroidLibraryAarInfo` as the basis for bundling and then merges the transitive closure with the nested `classes.jar` using the existing logic.

Usage of `AndroidLibraryAarInfo` requires an `--experimental_google_legacy_api` flag, which is pretty undesirable given it would require consumers to configure the same flag, even if not using for publishing Android artifacts. As such, I understand this PR probably isn't acceptable from a generic standpoint, but I wanted to open it up for feedback, ideas, and inspiration. This is what we'll be using to publish our Android targets until we can find a better alternative.

Example usages:

```
kt_jvm_library(
    name = "kt-lib",
    srcs = glob(["src/main/java/**/*.kt"]),
    resources = glob(["src/main/resources/**"]),
    exports = main_exports,
    deps = main_deps,
)

android_library(
    name = "android-lib",
    manifest = ":src/main/AndroidManifest.xml",
    resource_files = glob(["src/main/res/**"]),
    tags = ["maven_coordinates=group:artifact:aar:%s" % VERSION],
    exports = main_exports + [
        ":kt-lib",
    ],
)

maven_export(
    name = "export",
    lib_name = "android-lib",
    maven_coordinates = "group:artifact:aar:%s" % VERSION,
)
```

External example (we're using an Android databinding wrapper which does work to create a base `.aar` instead of `android_library`):
https://github.com/player-ui/player/blob/78345ac70c5bb34fcd58147ab49e38f22d3524fd/android/player/BUILD#L10-L27


private/tools/java/com/github/bazelbuild/rules_jvm_external/maven/MavenPublisher.java
```diff
@@ -93,14 +93,15 @@ public static void main(String[] args)
...
       List<CompletableFuture<Void>> futures = new ArrayList<>();
-      futures.add(upload(repo, credentials, coords, ".pom", pom, gpgSign));
 
       if (mainArtifact != null) {
         String ext =
             com.google.common.io.Files.getFileExtension(mainArtifact.getFileName().toString());
         futures.add(upload(repo, credentials, coords, "." + ext, mainArtifact, gpgSign));
       }
 
+      futures.add(upload(repo, credentials, coords, ".pom", pom, gpgSign));
```
sugarmanz, line 103: Relevant discussion for this change:
https://github.com/vaticle/bazel-distribution/pull/318

Might break this out into a separate contribution since this is unrelated, yet necessary for publishing support for Artifactory repositories.

jin:
@ahumesky could you advise on the use of `--experimental_google_legacy_api` and `AndroidLibraryAarInfo` for this? 

Or are there alternative upstream mechanisms that tackles https://github.com/bazelbuild/bazel/issues/348?

Issue #962, opened 2023-09-22
LockFileConverter replaces "+" with space in version strings


gergelyfabian:
If I set up a dependency on a version like "2.0.17+15-d6f2c09b+20230921-1427-SNAPSHOT" (e.g. sbt tends to generate such versions for snapshots), then the coursier generated json (`coursier-deps.json`) contains:

```
"coord": "groupId:artifactId:2.0.17+15-d6f2c09b+20230921-1427-SNAPSHOT",
```

But `unsorted_deps.json` has:

```
    "groupId:artifactId": {
      "shasums": {
        "jar": "f36e2210fa925792a6667f5c1a8515b3b7aa8d85fd0ffe20886cc6a4a40d9956",
        "sources": "805cc867deb5952f97ba2076f14926eeff02599af4edd05f980c5395b656bcc8"
      },
      "version": "2.0.17 15-d6f2c09b 20230921-1427-SNAPSHOT"
    },
```

This causes an issue when building my project:

```
Error in repository_rule: invalid repository name '@groupdId_artifactId_2_0_17 15_d6f2c09b 20230921_1427_SNAPSHOT': workspace names may contain only A-Z, a-z, 0-9, '-', '_' and '.'
```

cheister:
I don't think `+` is a character [maven supports in versions](https://maven.apache.org/guides/mini/guide-naming-conventions.html).  That being said, is there an example of a maven artifact with a `+` in the version we could test with?

cheister:
Ok, I found https://github.com/bazelbuild/rules_jvm_external/issues/170, which runs a test with an artifact that has `+` in version.  Any ideas why that one works but your's is failing?

gergelyfabian:
How do I differentiate a V1 lock file from a V2 lock file?

cheister:
> How do I differentiate a V1 lock file from a V2 lock file?

Any time you pin deps with rules_jvm_external 5.1 or later you'll get a V2 lock file.

You can compare the diff in the PR that introduced V2 lock file https://github.com/bazelbuild/rules_jvm_external/pull/822/files#diff-9c6156c53763ef90d5f88fd646fd01e0abfc1ab16ece3884a444462d1ed1840a,

gergelyfabian:
I'm using a V2 lockfile I think, and I still can reproduce this issue. Just checked with building latest snapshot of zio, publishing it to local `.m2` and using at as a dependency.

```
$ sbt
sbt:zio> ++2.12
[...]

PR #1053, opened 2024-02-01
tests for artifact with plus

gergelyfabian:
Related to #962, reproducing base case that works. Added maven_install stanzas for an artifact that contains plus in its version.

Also fixed some test issues related to deleting generated files. File names changed after switching to bzlmod.
Added a sanity check in expect_not_in_file() to ensure it's being called for a good file (and not reporting missing string due to missing file).

Moved around one test for proper grouping.


gergelyfabian:
Related to #170, #169.

shs96c (approved):
LGTM. Thank you for adding these tests!

PR #1054, opened 2024-02-06
Do not remove LICENSE files when combining jars

gkdn:
maven_project_jar has a jar combining action which is attempting
to exclude dependencies' contents from the jar. However this also
removes LICENSE file from the Jar file since it is quite common for
external dependencies to include as well.

This is similar to META_INF case which the jar merging tool is special
casing and the PR piggy backs on that special case.


gkdn:
@jin could you pls take a look?

jin (approved):

jin:
thanks!

PR #1058, opened 2024-02-22
Make jvm_import use `java` from resolved toolchain instead of host machine.

jin:
This ensures that the AddJarManifestEntry jar used in the `jvm_import` actions is executed with the jdk version set by the downstream project's language version (which was used to correctly build the jar itself), which resolves to the right toolchain, instead of using whatever `@local_jdk` resolves to.

This fixes the case where the jar is built with a newer JDK version/lang level, but executed with an older runtime.

Unblocks Bazel's JDK 21 upgrade.

Fixes #895

PR #1059, opened 2024-02-23
Use tool instead of compilation runtime in `jvm_import` actions

fmeum:
PR #1058 made the `jvm_import` actions use Java toolchains rather than host tools, but chose the runtime used to run `javac` instead of the runtime used to run build tools. The latter can be configured via `--tool_java_runtime_version` while the former is fixed in the Java compilation toolchain.


fmeum:
FYI @jin

jin (approved):
Thank you :)

PR #1061, opened 2024-02-23
Revert #1058 and #1059

jin:
These PRs aren't needed since the downstream Bazel build just needed to pass `--tool_java_runtime_version` on top of
`--tool_java_language_version`.

cc @fmeum


shs96c (approved):
D'oh! LGTM

PR #1063, opened 2024-02-29
Index service implementations as well as packages

illicitonion:
This can be useful for detecting things like annotation processors, and
knowing whether a target represents a java_plugin or just a
java_library.


WORKSPACE
```diff
@@ -425,6 +425,14 @@ maven_install(
     ],
 )
 
+maven_install(
+    name = "service_indexing_testing",
```
shs96c, line 429: I am mildly surprised this doesn't need to be included in the `use_repo` for `dev_maven` further down this file (around line 665ish). Because we now alias `pin` into the generated workspace, you won't need to include the `unpinned_` version there too.

coursier.bzl
```diff
@@ -1112,7 +1112,10 @@ def _coursier_fetch_impl(repository_ctx):
                 artifact.pop("url")
             continue
         artifact.update({"sha256": shas[path]})
-        artifact.update({"packages": jars_to_packages[path]})
+        artifact.update({"packages": jars_to_index_results[path]["packages"]})
+        service_implementations = jars_to_index_results[path]["serviceImplementations"]
```
shs96c, line 1116: This will be safer with a `get`

private/tools/java/com/github/bazelbuild/rules_jvm_external/jar/IndexJar.java
```diff
@@ -18,27 +23,47 @@
...
 
-public class ListPackages {
+public class IndexJar {
 
   private static final Predicate<String> IS_NUMERIC_VERSION =
       Pattern.compile("[1-9][0-9]*").asPredicate();
 
+  private static final String SERVICES_DIRECTORY_PREFIX = "META-INF/services/";
+
+  public static class PerJarIndexResults {
```
shs96c, line 35: It's fine to put this into a top-level class.

private/tools/java/com/github/bazelbuild/rules_jvm_external/jar/IndexJar.java
```diff
@@ -54,13 +79,28 @@ public static void main(String[] args) throws IOException {
...
+  public static PerJarIndexResults index(Path path) throws IOException {
     SortedSet<String> packages = new TreeSet<>();
-    try (InputStream fis = new BufferedInputStream(Files.newInputStream(path));
-        ZipInputStream zis = new ZipInputStream(fis)) {
-      try {
-        ZipEntry entry;
-        while ((entry = zis.getNextEntry()) != null) {
+    SortedMap<String, SortedSet<String>> serviceImplementations = new TreeMap<>();
+    try {
+      try (ZipFile zipFile = new ZipFile(path.toFile())) {
```
shs96c, line 86: It's not clear why you've switched to a `ZipFile` here. You can just read to EOF of `zis` to read the contents of an entry from the `ZipInputStream`, which would save the extra try-with-resources block you have later in this.

tests/com/github/bazelbuild/rules_jvm_external/jar/IndexJarTest.java
```diff
@@ -0,0 +1,122 @@
...
+
+  @Test
+  public void noPackages() throws Exception {
+    doTest("hamcrest_core_srcs_for_test/file/hamcrest-core-1.3-sources.jar", sortedSet(), new TreeMap<>());
+  }
+
+  @Test
+  public void serviceImplementations() throws Exception {
+    doTest("lombok_for_test/file/lombok-1.18.22.jar",
+        sortedSet("lombok", "lombok.delombok.ant", "lombok.experimental", "lombok.extern.apachecommons", "lombok.extern.flogger", "lombok.extern.jackson", "lombok.extern.java", "lombok.extern.jbosslog", "lombok.extern.log4j", "lombok.extern.slf4j", "lombok.javac.apt", "lombok.launch"),
```
shs96c, line 69: Nit, can we put these on separate lines? This is hard to read.

tests/com/github/bazelbuild/rules_jvm_external/jar/IndexJarTest.java
```diff
@@ -0,0 +1,122 @@
...
+  }
+
+  private static class Lockfile {
+    public TreeMap<String, TreeMap<String, TreeSet<String>>> services;
+  }
+
+  @Test
+  public void testLockfile() throws Exception {
+    Path lockfilePath = Paths.get(Runfiles.create().rlocation("rules_jvm_external/tests/custom_maven_install/service_indexing_testing.json"));
+    Reader reader = Files.newBufferedReader(lockfilePath);
```
shs96c, line 86: This should be in a try-with-resources.

shs96c (approved):
Needs a rebase, and there are comments to address, but looks good.

WORKSPACE
illicitonion: I believe this is included in the `use_repo` for `dev_maven` (currently on line 687) - is that not what you mean?

shs96c (approved):
LGTM. Thank you!

PR #1069, opened 2024-03-14
fix: determine the version of java correctly

yesudeep:
Fedora Linux has updated its latest JDK to version `22-ea` and that breaks the version parsing routine on here. 

```
zsh❯ java --version
openjdk 22-ea 2024-03-19
OpenJDK Runtime Environment (Red_Hat-22.0.0.0.36-1) (build 22-ea+36)
OpenJDK 64-Bit Server VM (Red_Hat-22.0.0.0.36-1) (build 22-ea+36, mixed mode, sharing)

zsh❯ which java
/usr/lib/jvm/java-22-openjdk/bin/java
```

The rewritten version parser should adapt better.

### DIFF
--- a/private/java_utilities.bzl
+++ b/private/java_utilities.bzl
@@ -2,19 +2,26 @@
 # Utilites for working with java versions
 #
 
-# Get numeric java version from `java -version` output e.g:
-#
-#     openjdk version "11.0.9" 2020-10-20
-#     OpenJDK Runtime Environment AdoptOpenJDK (build 11.0.9+11)
-#     OpenJDK 64-Bit Server VM AdoptOpenJDK (build 11.0.9+11, mixed mode)
-#
-def parse_java_version(java_version_output):
-    # Look for 'version "x.y.z"' in the output'
-    if len(java_version_output.split("version ")) > 1:
-        java_version = java_version_output.split("version ")[1].partition("\n")[0].split(" ")[0].replace("\"", "")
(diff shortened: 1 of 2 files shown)


private/java_utilities.bzl
```diff
@@ -24,7 +31,21 @@ def parse_java_version(java_version_output):
...
         return int(java_version.split(".")[0])
 
-    return None
+# Get numeric java version from `java -version` output e.g:
+#
+#     openjdk version "11.0.9" 2020-10-20
+#     OpenJDK Runtime Environment AdoptOpenJDK (build 11.0.9+11)
+#     OpenJDK 64-Bit Server VM AdoptOpenJDK (build 11.0.9+11, mixed mode)
+#
+def parse_java_version(java_version_output):
```
jin, line 40: Function can be simplified:

```
def parse_java_version(java_version_output):
    first_line = java_version_output.strip().split("\n")[0]
    if not first_line:
        return None
    i = index_of_version_character(first_line)
    if i == None:
        return None
    j = index_of_non_version_character_from(first_line, i + 1)
    return get_major_version(first_line[i:j])
```

jin (commented):
Thanks!

The SecurityManager failure is not related, already happening at HEAD. https://buildkite.com/bazel/rules-jvm-external-examples/builds/3809

jin (approved):

PR #1071, opened 2024-03-15
Use lock-file for maven_jar_migrator repo

scaiper:
This PR enables use of a lock-file for `maven_jar_migrator` repo. This in turn enables caching and reproducible builds.


scaiper:
Failed checks seem unrelated.

shs96c:
We deleted the maven jar migrator in 647826ab3ab1d5a41000199302328d31228b248c, so I think it would be better to remove this from the module file entirely....

scaiper:
Removed it.

Issue #1070, opened 2024-03-15
Maven install with lock manifest is making inefficient use of caches


manuelnaranjo:
I've been for a while struggling with those _extension targets we create, we have a setup where we're using build buddy open source over grpcs for caching and bes. Our repository has 3 maven_install trees of at least 300 coordinates each, on top of that we have a few maven dependencies that contain resources and are really large, literally 250MB for some testing. 

Whenever the network connection to the remote cache is flaky we end up spending a lot of time in updating the cache or downloading from the cache this _extension targets.

After looking at the source code I noticed two problems, which seem to have been unnoticed for long time as git blame shows it got refactored from coursier over 5 years ago.

The first problem is the use of genrule which doesn't provide a mnemonic like other actions and doesn't allow override it's execution features to disable remote caching or remote execution. By using a custom rule with a mnemonic the consumers of rules_jvm_external could disable remote caching as the operation is very IO sensitive and very lightweight to do locally.

The second problem is that it uses cp, we literally copy the http_file into the @maven repository, while instead we could just have a symlink when the OS allows.

This whole makes me thing adding a custom private rule for this, or depending on something like aspect Bazel lib copy_file would make things better and hopefully even speed up CI operations on large repositories. As it would reduce the IO load.

I'm going to hack something soon, but yet I wanted to open the discussion, most likely I'm missing something here.

PR #1074, opened 2024-03-18
Stop copying jar files

manuelnaranjo:
https://github.com/bazelbuild/rules_jvm_external/issues/1070

Stop making use of a genrule that copies the already downloaded jar handled by http_file file into the target repository and instead make use of an alias for proxying the artifact into the target repository.


shs96c:
@manuelnaranjo, the idea seems like a good one, but I note that none of the CI tests are working yet. You can run these locally using `bazel test //...` assuming you have the Android pieces set up locally. If you're using macOS, we have [some basic instructions on how to set up the Android SDK/NDK](https://github.com/bazelbuild/rules_jvm_external?tab=readme-ov-file#installing-the-android-sdk-on-macos) so that it works with the project.

shs96c:
@manuelnaranjo, do you need a hand getting this one over the line? It'd be nice to get it into the tree. If you do, just LMK.

shs96c:
@manuelnaranjo, I tend to lurk on the Bazel slack (both the "general" and the "java" channels) so feel free to ping me there, but will be on holiday for the next two weeks, so there's no hurry :)

PR #1077, opened 2024-03-19
Run `bazel run scripts:refresh-prebuilts`

shs96c:


jin:
Can we start building these from source? Alternatively I can work on migrating to trusted builders.

shs96c:
We only need the pre-built binaries because we use them in our repository rules. If we're happy flipping the switch and requiring a lock file in all cases (even an empty one) so that we can pin using a `bazel run` instead of `bazel build`, then we could remove these entirely. If we're not happy with that, it'd be nice to have a trusted builder check these into the repo.

Having said that, `outdated` is probably unnecessary as a prebuilt, given that I _think_ the only way to use it is via `bazel run`.

PR #1079, opened 2024-03-20
Fixes problem with bom file generation with Bazel 7

shs96c:


jin (commented):
Why wasn't this caught by the examples build? https://github.com/bazelbuild/rules_jvm_external/blob/478a05df4b49232a3fd45ebfe05a7c2023fc8fc0/examples/java-export/BUILD#L21-L29?

What's the failure mode? `Label#relative` appears to be also a deprecated API: https://bazel.build/rules/lib/builtins/Label#relative

shs96c:
I do need to add a test for this. Apparently we don't generate the required a Maven BOM in our existing repo. Should be easy to do.

I'll switch to `same_package_label` instead. I'd not noticed that API was deprecated. Thank you for the catch.

shs96c:
I switched the `java_export` example to use `bzlmod`, and that demonstrated the problem pretty clearly. In the course of fixing that, I found that if people weren't fetching sources for their repos, then we weren't properly generating the build file iff the lock file being used had `fetch_sources = True`. That problem is also fixed, and was also highlighted by the `java_export` example.

jin (approved):

PR #1082, opened 2024-03-21
[resolver] Add conflict information from the Maven resolver

shs96c:


private/tools/java/com/github/bazelbuild/rules_jvm_external/resolver/cmd/Main.java
```diff
@@ -256,21 +259,31 @@ private static void writeLockFile(
...
+      Set<DependencyInfo> infos,
+      Set<Conflict> conflicts)
       throws IOException {
     listener.onEvent(new PhaseEvent("Building lock file"));
     Path output = config.getOutput();
 
     listener.close();
 
+    Function<Coordinates, String> simpleForm =
+        c -> c.getGroupId() + ":" + c.getArtifactId() + ":" + c.getVersion();
```
jin, line 271: would it make sense to implement `Coordinates#toString`  instead?

jin (approved):

private/tools/java/com/github/bazelbuild/rules_jvm_external/resolver/Conflict.java
```diff
@@ -0,0 +1,57 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.github.bazelbuild.rules_jvm_external.resolver;
+
+import com.github.bazelbuild.rules_jvm_external.Coordinates;
+import java.util.Objects;
+
+public class Conflict {
```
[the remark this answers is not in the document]
shs96c: Sure.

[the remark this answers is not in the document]
shs96c: We still support Java 11, but it would be a wonderful thing if we could use a record

private/tools/java/com/github/bazelbuild/rules_jvm_external/resolver/cmd/Main.java
shs96c: This format is what we had before with the `coursier`-based resolver, and my assumption is that people have built tooling around that specific format. If we don't mind, then I can definitely use `toString` as it's less code for us :)

shs96c: Switched to `toString`. People can file an issue if they've built a workflow around an undocumented part of the lock file.

private/tools/java/com/github/bazelbuild/rules_jvm_external/resolver/maven/MavenResolver.java
```diff
@@ -212,7 +215,50 @@ public Graph<Coordinates> resolve(ResolutionRequest request) {
...
-    return buildGraph(coordsListener.getRemappings(), directDependencies);
+    Graph<Coordinates> resolution = buildGraph(coordsListener.getRemappings(), directDependencies);
+
+    Set<Coordinates> simpleRequestedDeps =
+        request.getDependencies().stream()
+            .map(com.github.bazelbuild.rules_jvm_external.resolver.Artifact::getCoordinates)
+            .map(
+                c ->
+                    new Coordinates(
+                        c.getGroupId() + ":" + c.getArtifactId() + ":" + c.getVersion()))
```
jin, line 226: `toString` here too and below.

jin (approved):

private/tools/java/com/github/bazelbuild/rules_jvm_external/resolver/maven/MavenResolver.java
shs96c: Good catch. Done!

PR #1092, opened 2024-04-04
Specify minimum bazel version in the module file

shs96c:


fmeum:
The MODULE.bazel specific formatting has been unapplied here. Just curious, how did you format the module file?

shs96c:
I format the codebase using `bazel run scripts:format` which also applies a java code formatter too. We use a pinned `buildifier`, so whoever runs the script should get the same result. However, I suspect that my system buildifier is being used when I modify the file in IJ, which may be at a different version.

jin (approved):
It'd be great to minimize the diff to the `bazel_compatibility` line, but that's a nit.

PR #1094, opened 2024-04-04
Temporarily turn down android_local_test examples builds on macOS

shs96c:
The CI tests are failing again.


jin (approved):

shs96c:
The warning comes from the underlying JDK, so it looks like maybe that changes between runs?

PR #1101, opened 2024-04-15
Strip comments, spaces, and tabs from services

illicitonion:


cheister (approved):
Thanks for the quick fix!

PR #1103, opened 2024-04-15
Create a regression testing install for the maven resolver

cheister:
So we can test dependency resolution with coursier and with maven


shs96c:
Any chance you can add a reduced test case to the ResolverTestbase? Writing it may take some time, but the iteration speed is so much better.

cheister:
> Any chance you can add a reduced test case to the ResolverTestbase? Writing it may take some time, but the iteration speed is so much better.

Yes, I think we can also do that for test cases, but it's been useful to have a regression test set of dependencies for the coursier resolver and I think we'll want the same full integration regression style tests for the maven resolver.

shs96c (approved):
LGTM, but the test failure looks relevant

Issue #1106, opened 2024-04-19
[Bug] Resources not being properly merged when a `java_export` depends on another `java_export`


vinnybod:
I have a minimal reproduction of this issue in a repo here: https://github.com/vinnybod/bazel_java_example/tree/license-file-issues
The README also explains it in detail, but I'll summarize some of it here.

The use case is I need to add a `COPYRIGHT` file to all of my jars that get published to a Maven repo.

I put the `COPYRIGHT` as a resource to `java_export` `app` and `java_export` `lib`.
I have `app` depend on `lib`.
`COPYRIGHT` appears in `lib-project.jar`, but is noticeably absent from `app-project.jar`.
If I remove `COPYRIGHT` from `lib`, then it will appear in `app-project.jar`

If I swap out `lib` from `java_export` to `java_library`, then `app-project.jar` gets the `COPYRIGHT` file.

I would expect the `COPYRIGHT` file to appear in `-project.jar` files when being added as a resource.

---
**Edit:** Changed the example to `COPYRIGHT`. It appears there is a special carve-out in `MergeJars` to allow this use case for `LICENSE` https://github.com/bazelbuild/rules_jvm_external/blob/master/private/tools/java/com/github/bazelbuild/rules_jvm_external/jar/MergeJars.java#L134

It's possible this is actually the desired behavior, but we need a way to dynamically allow other filenames in addition to `LICENSE`.

Related https://github.com/bazelbuild/rules_jvm_external/pull/1054

vinnybod:
It seems simple enough to fix this with two lines. I am wondering if it is worth making it so that the `java_export` macro can pass down a list to `MergeJars`, or if this is not a big enough use case.

```
diff --git a/private/tools/java/com/github/bazelbuild/rules_jvm_external/jar/MergeJars.java b/private/tools/java/com/github/bazelbuild/rules_jvm_external/jar/MergeJars.java
index c8b2e8d..218abf1 100644
--- a/private/tools/java/com/github/bazelbuild/rules_jvm_external/jar/MergeJars.java
+++ b/private/tools/java/com/github/bazelbuild/rules_jvm_external/jar/MergeJars.java
@@ -33,6 +33,7 @@ import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
[...]

PR #1107, opened 2024-04-22
Add an exception for COPYRIGHT and NOTICE files in the same way as LICENSE

vinnybod:
Addresses https://github.com/bazelbuild/rules_jvm_external/issues/1106

### DIFF
--- a/private/tools/java/com/github/bazelbuild/rules_jvm_external/jar/MergeJars.java
+++ b/private/tools/java/com/github/bazelbuild/rules_jvm_external/jar/MergeJars.java
@@ -111,6 +111,7 @@ public class MergeJars {
 
     Map<String, Set<String>> allServices = new TreeMap<>();
     Set<String> excludedPaths = readExcludedFileNames(excludes);
+    Set<String> duplicateExceptions = Set.of("COPYRIGHT", "LICENSE", "NOTICE");
 
     // Ultimately, we want the entries in the output zip to be sorted
     // so that we have a deterministic output.
@@ -131,7 +132,7 @@ public class MergeJars {
 
           if ("META-INF/".equals(entry.getName())
               || (!entry.getName().startsWith("META-INF/")
-                  && !entry.getName().equals("LICENSE")
+                  && !duplicateExceptions.contains(entry.getName())
                   && excludedPaths.contains(entry.getName()))) {
             continue;
           }


shs96c:
Thank you for the PR. It looks great, and I'll happily merge it, but could you please sign the Google CLA? If you [follow the link in the Google CLA bot's comment](https://github.com/bazelbuild/rules_jvm_external/pull/1107#issuecomment-2070440784) you'll be able to do that.

vinnybod:
Thanks @shs96c , I'm just waiting on the authorized signer's signature. Hopefully will go through today.

vinnybod:
CLA is signed, but its still not going through. May still be waiting for it to be processed on Google's side

shs96c:
@jin any idea why the CLA check isn't passing?

@vinnybod, the CLA check is complaining that `<v***e​@confluent.io>` hasn't signed the CLA. Did you use that email address? Maybe it's something to do with that?

vinnybod:
Yes, I got a corporate CLA signed and submitted and I am in the Google group.
According to https://opensource.google/documentation/reference/cla#cla-accepted , it might take a couple days for it to process.

shs96c:
Great. I'll re-run the check tomorrow and see if it passes. Thank you for being patient!

vinnybod:
@shs96c I'm fine merging this since it solves my immediate issue.

But I do want to continue this discussion a bit more, because I can see myself needed to add more files to this list in the near future.

Why should file `NOTICE` in `app-export` be excluded just because it also exists in `lib-export`? I wonder if this is actually a bug.
It seems like if I want to include `NOTICE` in `app-export`, that should happen regardless of whether `lib-export` contains it or not. Does that make sense?

vinnybod:
Actually, I think we should wait to merge this. I am seeing that in some cases this does not do what I need.
For example I have a java_export where I Add a `NOTICE`, but instead of my `NOTICE`, it is including the `NOTICE`, from my scala dependency.

shs96c:
The `LICENSE` files are kept because it's often a legal requirement to keep them. The default `ZipOutputStream` in java forbids multiple files with the same name, but in theory what we need to do is write each license to the zip and preserve it. For Buck, I wrote [a custom zip implementation that did this](https://github.com/facebook/buck/blob/main/src/com/facebook/buck/util/zip/AppendingZipOutputStreamImpl.java)

vinnybod:
Okay, so I now believe this to be fine to merge. The issue I was running into was due to a misconfiguration with `excluded_workspaces` that was causing some extra files in the jars from dependents that should not have been there.

TL;DR: This solves the immediate issue of files being excluded from the `-project` jar when one of its deps contains a file of the same name. Treating `COPYRIGHT` and `NOTICE` files the same as we do `LICENSE`.

shs96c (approved):
LGTM. Thank you!

PR #1108, opened 2024-04-23
Improve repinning performance by not recomputing cache path for every artifact

BalestraPatrick:
We recently investigated the performance of `rules_jvm_external` because running `REPIN=1 bazel run @unpinned_maven//:pin` was taking about 140s in our project. With this patch, it now takes about 20s.

The bottleneck seemed to be `get_coursier_cache_or_default` which was being invoked for every single artifact in a loop. Our repos have more ~1k artifacts, so this was a substantial amount of work.

### DIFF
--- a/coursier.bzl
+++ b/coursier.bzl
@@ -212,10 +212,6 @@ def _relativize_and_symlink_file_in_coursier_cache(repository_ctx, absolute_path
     #
     # Also replace '\' with '/` to normalize windows paths to *nix style paths
     # BUILD files accept only *nix paths, so we normalize them here.
-    #
-    # Also, replace '//' with '/', otherwise parsing of the file path for the
-    # coursier cache will fail if variables like HOME or COURSIER_CACHE have a
-    # trailing slash.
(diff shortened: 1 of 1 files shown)


coursier.bzl
```diff
@@ -216,13 +216,6 @@ def _relativize_and_symlink_file_in_coursier_cache(repository_ctx, absolute_path
     # Also, replace '//' with '/', otherwise parsing of the file path for the
```
jin, line 216: Also move this comment.

jin (approved):
Nice catch!

coursier.bzl
BalestraPatrick: Done!

jin (approved):

mattnworb:
@jin any clue if there is a new release of rules_jvm_external in the plans for any time soon? We'd love to not have to add this as a patch on top of RJE in our workspace (I work with @BalestraPatrick )

jin:
Yeah we can cut 6.1 pretty soon.  @shs96c

PR #1109, opened 2024-04-24
Handle maven artifact not found

cheister:
Maven does not fail when a transitive dependency is missing.  It prints out a warning similar to
```
[WARNING] The POM for org.mortbay.jetty:jetty:jar:6.1.26.cloudera.2 is missing, no dependency information available
```
this PR implements the same behavior for the maven resolver.

This changes the manageDependencies generation to match on all artifact fields groupId:artifactId:extension:version:classifier instead of just groupId:artifactId

### DIFF
--- a/tests/com/github/bazelbuild/rules_jvm_external/resolver/MavenResolverTest.java
+++ b/tests/com/github/bazelbuild/rules_jvm_external/resolver/MavenResolverTest.java
@@ -14,12 +14,15 @@
 
 package com.github.bazelbuild.rules_jvm_external.resolver;
 
+import static org.junit.Assert.fail;
+
 import com.github.bazelbuild.rules_jvm_external.Coordinates;
 import com.github.bazelbuild.rules_jvm_external.resolver.cmd.ResolverConfig;
 import com.github.bazelbuild.rules_jvm_external.resolver.events.EventListener;
 import com.github.bazelbuild.rules_jvm_external.resolver.maven.MavenResolver;
 import com.github.bazelbuild.rules_jvm_external.resolver.netrc.Netrc;
 import java.nio.file.Path;
+import org.eclipse.aether.transfer.ArtifactNotFoundException;
 import org.junit.Test;
 
 public class MavenResolverTest extends ResolverTestBase {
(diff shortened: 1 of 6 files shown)


MODULE.bazel
```diff
@@ -595,13 +592,36 @@ dev_maven.artifact(
...
 
+# These artifacts helped discover limitations in the Maven resolver. Each
+# artifact listed here *must have* an accompanying issue. We build_test these
+# targets to ensure that they remain supported by the rule.
+dev_maven.install(
+    name = "regression_testing_maven",
+    artifacts = [
+        # Depends on org.apache.yetus:audience-annotations:0.11.0 which has an invalid pom
+        "org.apache.parquet:parquet-common:1.11.1",
+        # To match Maven's behavior the resolver should not error when there is an artifact
```
shs96c, line 610: For top-level artifacts, this seems like something we'd want to fail on. For transitive deps, the warning seems sensible.

cheister: Yeah, that makes sense.  I'll try updating it so it throws the exception if the missing artifact is in dependency list.

tests/com/github/bazelbuild/rules_jvm_external/resolver/MavenResolverTest.java
```diff
@@ -39,4 +42,19 @@ public void shouldSuccessfullyResolveNettyStaticClasses() {
     // There should be no cycle detected by this dependency
     resolver.resolve(prepareRequestFor(repo.toUri(), main));
   }
+
+  @Test
+  public void shouldErrorOnMissingTopLevelDependency() {
```
shs96c, line 47: This is probably something we want in `ResolverTestBase`, but we can land this PR and update.

shs96c (approved):
The tests pass, and the pathological `maven_install` I have from a large project works just fine. LGTM.

tests/com/github/bazelbuild/rules_jvm_external/resolver/MavenResolverTest.java
cheister: Since this one is checking for a Maven specific exception I put it in the MavenResolverTest as other resolvers would have a different exception.  We could just look for a RuntimeException and move it into `ResolverTestBase` if you want.

PR #1112, opened 2024-04-26
Ensure @unpinned_maven//:pin continues to work with the maven resolver 

shs96c:


MODULE.bazel
```diff
@@ -1,8 +1,8 @@
 module(
     name = "rules_jvm_external",
```
jin, line 2: What's going on with the formatting? I feel like I've seen these types of alternating formatting changes a lot.

jin (approved):

MODULE.bazel
shs96c: I'll revert the change to `MODULE.bazel`, but I think the underlying cause is that there are different versions of `buildifier` in use. There's the version on the dev's machine, the pinned version the format script uses, and I've an IJ plugin that also likes to run buildifier. I think they're fighting amongst themselves.

Worse, I think the formatting that one version ("X") of buildifier applies is fine with the different version ("Y"), but the formatting that Y applies is something that X wants to undo.

PR #1115, opened 2024-04-27
Fix java version parsing when _JAVA_OPTIONS are returned from 'java -version'

cheister:
https://github.com/bazelbuild/rules_jvm_external/pull/1069 broke the case where `java -version` returns 
```
Picked up _JAVA_OPTIONS:
```
in the first line of the version string


shs96c (approved):
LGTM. Thank you for the test too!

PR #1116, opened 2024-04-28
Since we now require Java 11 we don't need to check for java 8 when building an argsfile

cheister:
Since we moved the [minimum supported Java version to 11 ](https://github.com/bazelbuild/rules_jvm_external/pull/972)we no longer need to check for argsfile support on Java version 8.


shs96c (approved):
Deleting code? Wonderful! LGTM

PR #1117, opened 2024-04-28
Move coursier.bzl under private/rules

cheister:
Shouldn't the coursier rules be next to all of the other private rule declarations?


shs96c (approved):
Makes sense, and it makes the code a little cleaner. Thank you! LGTM

PR #1119, opened 2024-04-29
Fix invalid reference in coursier.bzl

### LINKED ISSUE #1118: coursier.bzl contains invalid reference
Version 6.1 contains an invalid reference in coursier.bzl:

```
label '@@[unknown repo '' requested from @@rules_jvm_external~]//does/not:exist' is not visible
```

mortenmj:
coursier.bzl contains a reference which is not compatible with bzlmod. Fixes #1118.

### DIFF
--- a/private/rules/coursier.bzl
+++ b/private/rules/coursier.bzl
@@ -1364,7 +1364,7 @@ pinned_coursier_fetch = repository_rule(
             doc = "Instructions to re-pin the repository if required. Many people have wrapper scripts for keeping dependencies up to date, and would like to point users to that instead of the default.",
         ),
         "excluded_artifacts": attr.string_list(default = []),  # only used for hash generation
-        "_workspace_label": attr.label(default = Label("@//does/not:exist")),
+        "_workspace_label": attr.label(default = Label("@@//does/not:exist")),
     },
     implementation = _pinned_coursier_fetch_impl,
 )


private/rules/coursier.bzl
```diff
@@ -1364,7 +1364,7 @@ pinned_coursier_fetch = repository_rule(
             doc = "Instructions to re-pin the repository if required. Many people have wrapper scripts for keeping dependencies up to date, and would like to point users to that instead of the default.",
         ),
         "excluded_artifacts": attr.string_list(default = []),  # only used for hash generation
-        "_workspace_label": attr.label(default = Label("@//does/not:exist")),
+        "_workspace_label": attr.label(default = Label("@@//does/not:exist")),
```
shs96c, line 1367: We still need to support Bazel 5 and 6, so this will cause a regression there. I think we need to do something smart to figure out whether or not `bzlmod` is being used (not just checking the version number), and then Do The Right Thing

shs96c (commented):
Thank you for the PR! One comment, and it would be very helpful to have a test case to prevent us accidentally introducing a regression once this has landed.

private/rules/coursier.bzl
shs96c: The `maven_bom.bzl` file has a utility method to determine whether we're using `bzlmod` or not. We might want to lift that out into another file and depend on that.

mortenmj: @shs96c will you be dropping support for Bazel 6 once 8 is out?

mortenmj:
Looked like this was fixed by https://github.com/bazel-contrib/rules_jvm_external/pull/1265

PR #1125, opened 2024-05-03
skip non-class files in the exclusions processing for MergeJars

vinnybod:
This is sort of an addendum to https://github.com/bazelbuild/rules_jvm_external/issues/1106 and https://github.com/bazelbuild/rules_jvm_external/pull/1107
In the previous PR, I added an exception for `COPYRIGHT` and `NOTICE` since it was the same kind of use case as the existing `LICENSE` exception.

The issue that I am having is that I have files that should be present in all of the jars being published in a repo that are not `COPYRIGHT`, `NOTICE`, or `LICENSE`. For example a `project-version.properties` file that contains info about the build.

Changing the `readExcludedFileNames` function to only process `*.class` files allows for these files to propagate to the final jar. It also still passes all existing test cases.

The only test for this functionality is specifically looking at `.class` files.https://github.com/bazelbuild/rules_jvm_external/blob/master/tests/com/github/bazelbuild/rules_jvm_external/jar/MergeJarsTest.java#L221

I am open to other solutions, but this seems to have fixed a lot of issues for my use case without breaking any existing functionality that is being tested for.

### DIFF
--- a/private/tools/java/com/github/bazelbuild/rules_jvm_external/jar/MergeJars.java
+++ b/private/tools/java/com/github/bazelbuild/rules_jvm_external/jar/MergeJars.java
@@ -110,8 +110,7 @@ public class MergeJars {
     manifest.getMainAttributes().put(Attributes.Name.MANIFEST_VERSION, "1.0");
 
     Map<String, Set<String>> allServices = new TreeMap<>();
-    Set<String> excludedPaths = readExcludedFileNames(excludes);
-    Set<String> duplicateExceptions = Set.of("COPYRIGHT", "LICENSE", "NOTICE");
+    Set<String> excludedPaths = readExcludedClassNames(excludes);
 
     // Ultimately, we want the entries in the output zip to be sorted
     // so that we have a deterministic output.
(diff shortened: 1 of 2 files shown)


shs96c:
How would this cope with multiple versions of the same file? Things like log4j config files, or other elements that are often found in the root directory of a jar, and that could (would!) cause problems.

vinnybod:
This would still handle those duplicate files however the `--duplicates` flag is set, which defaults to `last-wins`  I added two additional tests to the pull request to show that behavior.

To me, this makes more sense than the existing behavior which is excluding files completely. I’m not 100% sure why the exclusions work the way they do. But, I don’t have the full picture of this tool, so please let me know me if I am missing anything.

Specifically, the case I am running into is with `java_export`. The `lib` that gets created has the files I expect, but then after the `-project` jar is generated, it is missing files.

shs96c (approved):
LGTM. Thank you for the PR!

shs96c:
Could you please rebase? Apparently there are merge conflicts :(

vinnybod:
@shs96c Thanks for your review.

Apologies, but I had to reopen this PR from a different org due to permission changes. 

https://github.com/bazelbuild/rules_jvm_external/pull/1239

PR #1133, opened 2024-05-13
fix: Use customized repositories

honnix:
It seems setting `repositories` is not sufficient to use customized repository because `@rules_jvm_external//:rules_jvm_external_deps_install.json` will always be picked up.


jin (approved):
LGTM

PR #1137, opened 2024-05-15
feat: Pass additional coursier cli options

honnix:
The primary use case this change tries to solve is passing down `--max-iterations` to coursier because the default value could be too small for certain artifacts.

### DIFF
--- a/private/extensions/maven.bzl
+++ b/private/extensions/maven.bzl
@@ -109,6 +109,7 @@ _install = tag_class(
         ),
         "ignore_empty_files": attr.bool(default = False, doc = "Treat jars that are empty as if they were not found."),
         "repin_instructions": attr.string(doc = "Instructions to re-pin the repository if required. Many people have wrapper scripts for keeping dependencies up to date, and would like to point users to that instead of the default. Only honoured for the root module."),
+        "additional_coursier_options": attr.string_list(doc = "Additional options that will be passed to coursier."),
     },
 )
 
@@ -220,6 +221,7 @@ def _maven_impl(mctx):
     # - use_starlark_android_rules: bool. A logical OR over all `use_starlark_android_rules` for all `install` tags with the same name.
     # - version_conflict_policy: string. Fails build if different and not a default.
     # - ignore_empty_files: Treat jars that are empty as if they were not found.
+    # - additional_coursier_options: Additional options that will be passed to coursier.
 
     # Mapping of `name`s to `bazel_module.name` This will allow us to warn users when more than
     # module attempts to update a maven repo (which is normally undesired behaviour)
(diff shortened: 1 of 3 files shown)


private/rules/coursier.bzl
```diff
@@ -794,6 +794,7 @@ def make_coursier_dep_tree(
         fetch_sources,
         fetch_javadoc,
         timeout,
+        additional_options,
```
shs96c, line 797: Since this is just for `coursier`, could you please name the parameter `additional_coursier_options`?

private/rules/coursier.bzl
```diff
@@ -1428,6 +1432,7 @@ coursier_fetch = repository_rule(
             ],
         ),
         "ignore_empty_files": attr.bool(default = False, doc = "Treat jars that are empty as if they were not found."),
+        "additional_options": attr.string_list(doc = "Additional options that will be passed to coursier."),
```
shs96c, line 1435: And rename here too?

private/rules/maven_install.bzl
```diff
@@ -27,7 +27,8 @@ def maven_install(
         aar_import_bzl_label = DEFAULT_AAR_IMPORT_LABEL,
         duplicate_version_warning = "warn",
         repin_instructions = None,
-        ignore_empty_files = False):
+        ignore_empty_files = False,
+        resolver_additional_options = []):
```
shs96c, line 31: `coursier_additional_options`, please.

shs96c (changes requested):
You will also need to mirror the same changes into the tag classes used for `bzlmod`. These can be found in `//private/extensions:maven.bzl`.

private/rules/maven_install.bzl
honnix: Sure. I was not sure whether it was fine to expose `coursier` in the parameter name here. Will fix.

private/rules/coursier.bzl
honnix: Since this file is dedicated to coursier, this may not be needed, right?

honnix: NVM I think using the same name is better for readability.

private/extensions/maven.bzl
```diff
@@ -338,6 +340,8 @@ def _maven_impl(mctx):
             if mod.is_root:
                 repo["repin_instructions"] = install.repin_instructions
 
+            repo["additional_coursier_options"] = repo.get("additional_coursier_options", []) + getattr(install, "additional_coursier_options", [])
```
honnix, line 343: I'm not very sure I'm doing the right thing, so please let me know if anything missed.

shs96c: This is perfect

shs96c (approved):
LGTM

peakschris:
Thank you so much for this fix. It's allowed me to pass the --pom_property option to coursier to allow it to resolve osgi jars. I've been stumbling around this for a year - much appreciated.

PR #1102, opened 2024-04-15
Generate java_plugin targets for annotation processors

illicitonion:

PR #1142, opened 2024-05-16
Propagate `testonly` attribute to generated plugins

rdesgroppes:
"Generate java_plugin targets for annotation processors" (#1102) brought a much-appreciated improvement by synthesizing a frequently repeated pattern. However, it doesn't work for annotation processor-related artifacts whose `testonly` attribute is set to `True`:
> ERROR: /path/to/external/maven/BUILD:12345:12: in java_plugin rule @@maven//:group_artifact__java_plugin__package_Class: non-test target '@@maven//:group_artifact__java_plugin__package_Class' depends on testonly target '@@maven//:group_artifact' and doesn't have testonly attribute set

The present change aims at covering the use case by taking a [Java Microbenchmark Harness](https://github.com/openjdk/jmh)-based target as a practical example.


illicitonion (approved):
Thanks! This LGTM, but unfortunately I'm not an owner of the repo - hopefully someone else will come and approve/merge soon!

jin (approved):

PR #1149, opened 2024-05-24
Update rules_jvm_external to use the Starlark version of aar_import after the native version was removed from Bazel

### LINKED ISSUE #1139: [Bazel CI] Build failed with errors: no native function or rule 'aar_import'
CI: https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/3858#018f7a6e-1b58-4ec1-a31d-6067bf5a3da1

Platform: Ubuntu

Logs:

```
ERROR: [0mTraceback (most recent call last):
Error: no native function or rule 'aar_import'
ERROR: [0mSkipping '//tests/integration/override_targets:override_targets': Error evaluating '//tests/integration/override_targets:override_targets': error loading package 'tests/integration/override_targets': Package 'tests/integration/override_targets' contains errors
ERROR: [0mError evaluating '//tests/integration/override_targets:override_targets': error loading package 'tests/integration/override_targets': Package 'tests/integration/override_targets' contains errors
```

Culprit: bazelbuild/bazel@947c72d

Steps:

```
 git clone https://github.com/bazelbuild/rules_jvm_external
git reset 0dfbe26a16f78d4669c8c11b2f0d79da8c086062  --hard
export USE_BAZEL_VERSION=88a230f4cf28deec1455cb2caed4dc9f81e108c9
bazel test //tests/integration/override_targets:override_targets 
```

CC Greenteam @fweikert

ahumesky:
The native version of `aar_import` has been removed from bazel at head, so use the Starlark version instead.

This PR:

1. Updates rules_jvm_external to use the Starlark version of aar_import automatically if the version of Bazel used does not contain the native version of aar_import.

1. Updates the version of Bazel in `.bazelversion` to 7.2.1 because the latest Starlark Android rules are only compatible with 7.2.1+

1. Updates the version of rules_android used to the latest commit, because version 0.1.1 is basically just wrappers around the native versions of the rules. However, version 0.1.1 of rules_android is still used in the `WORKSPACE` file so that test that run with Bazel 5 and Bazel 6 use the native rules, and Bazel 7+ uses the Starlark rules via bzlmod / `MODULE.bazel`, because the latest version of the Starlark rules are only compatible with Bazel 7.2.1+.

1. Updates the default label to import aar_import from, from `@build_bazel_rules_android//android:rules.bzl` to `@rules_android//rules:rules.bzl`, both to match the bzlmod style name of the repo ("rules_android") and to use the latest label inside the repo (`//android:rules.bzl` exists, but only for backwards compatibility).

1. Fixes some dependency issues with `aar_import_that_consumes_the_downloaded_file_directly` and `sceneform_rendering` `aar_import`s, because the Starlark version of `aar_import` performs additional checks ("ImportDepsChecker") to ensure that the dependencies of every class in the imported aar are in the dependencies. Notably, the class `com.google.ar.sceneform.assets.Loader` is used somewhere in the aar or its dependencies, but I couldn't find this class anywhere in any aars on maven.google.com, so I ended up having to stub it out with an empty implementation to satisfy the check.

1. Disables `--experimental_sibling_repository_layout` added in https://github.com/bazelbuild/rules_jvm_external/commit/b6631f90a46c23be32000bf14b8e817f44017cc4 because the latest rules_android uses protos in go, and there is a problem compiling protos in go with `--experimental_sibling_repository_layout`: https://github.com/bazelbuild/rules_go/issues/3947

1. Adds `--experimental_google_legacy_api` to `.bazelrc` until https://github.com/bazelbuild/rules_android/issues/219 is fixed

Fixes #1139.


ahumesky:
Oh, I just noticed the tests run on bazel as far back as bazel 5.4.0:
https://github.com/bazelbuild/rules_jvm_external/blob/71e08c82709b24d12e2d1040d86caa05bf6a8399/.bazelci/presubmit.yml#L40

The starlark Android rules will only work with 7.2.0 (releasing soon). Might need to segment the tests so that the tests with the starlark version of aar_import are only run with 7.2.0

ahumesky:
The windows tests will also be a problem because windows support for starlark rules_android is low priority

shs96c:
We still need to support Windows users, so those tests will need to pass properly before we can land this PR.

ahumesky:
https://github.com/bazelbuild/rules_android/pull/235 should fix up the windows problems

jin:
Error looks legit:

```

(19:43:56) ERROR: C:/b/rxn2cr7v/external/rules_android~/tools/android/BUILD:273:6: in alias rule @@rules_android~//tools/android:dexsharder: target '@@bazel_tools//src/tools/android/java/com/google/devtools/build/android/dexer:DexFileSplitter' is not visible from target '@@rules_android~//tools/android:dexsharder'. Check the visibility declaration of the former target if you think the dependency is legitimate
--
  | (19:43:56) ERROR: C:/b/rxn2cr7v/external/rules_android~/tools/android/BUILD:273:6: Analysis of target '@@rules_android~//tools/android:dexsharder' failed
  | (19:43:57) ERROR: Analysis of target '//tests/unit/aar_import:starlark_aar_import_test' failed; build aborted: Analysis failed
```

from https://buildkite.com/bazel/rules-jvm-external/builds/4289#018fc5e0-bc05-4701-80b8-55fdf1f6ddb0 (and others)

ahumesky:
That's fixed in 7.2.0, specifically https://github.com/bazelbuild/bazel/pull/22416, but there are other problems revealed after https://github.com/bazelbuild/rules_android/pull/235

I'll make this a draft until CI is passing

ahumesky:
The tests now work with the latest version of bazel:
https://buildkite.com/bazel/rules-jvm-external/builds/4383#01909944-3f4b-4b31-9bc5-128f223a586e

and the others are failing due to an older version of bazel being used and visibility changes that have been put into `@bazel_tools`. It might be tricky to work around these but I'll see what workarounds can be put into rules_android.

shs96c:
We do need the CI tests to pass before we can land this. We're fine asking for the latest versions of Bazel 5 and 6, but our version 7 support is more flexible.

ahumesky:
> We do need the CI tests to pass before we can land this.

Right I wouldn't want to submit something that breaks CI :)

> We're fine asking for the latest versions of Bazel 5 and 6, but our version 7 support is more flexible.

After going through most of the initial issues, it looks like the starlark android rules won't work with bazel 5 because at minimum they use `visibility()` for bzl visibility, which was introduced in bazel 6:

https://github.com/search?q=repo%3Abazelbuild%2Frules_android+%22visibility%28%22&type=code

I might be able to get the rules to work with bazel 6, but I'm not sure how long we can guarantee that (really we've only been developing for bazel 8 and backporting some things as needed to bazel 7).

Maybe I can arrange this so that the native android rules are used with the tests that use bazel 5 and bazel 6, and the starlark rules with latest bazel (8+). That might just amount to removing the `use_starlark_android_rules = True` and letting it auto select the starlark rules on latest bazel where native aar_import is deleted. There would still be problems with loading and registering the starlark android rules in the WORKSPACE file (e.g. `visibility()` above) -- but blzmod is used by default in bazel 7+, so I could perhaps just remove the starlark android rules from the WORKSPACE and leave them in MODULE.bazel.

ahumesky:
This stubbed out action is failing on windows:
https://github.com/bazelbuild/rules_android/blob/main/src/validations/aar_import_checks/BUILD

So that will have to be another update in rules_android

ahumesky:
Hm, well https://github.com/bazelbuild/rules_android/commit/746589b75f77655d4db9fa6bd5941311fbae96bc should have fixed the windows tests, but looks like that action is still failing... I'll have to try something else but out of time today

ahumesky:
With https://github.com/bazelbuild/rules_android/commit/f4b6edb93e13ba0bd17ecf7f99c80f4d3d6ca767 all the tests are now passing

ahumesky:
I updated the description and marked this ready for review

MODULE.bazel
```diff
@@ -28,6 +28,12 @@ bazel_dep(
     name = "rules_android",
     version = "0.1.1",
 )
+# Remove this once rules_android has an updated BCR release.
```
jin, line 31: ...newer than `0.5.0.bcr.1`?

private/rules/coursier.bzl
```diff
@@ -242,7 +242,9 @@ def _relativize_and_symlink_file_in_maven_local(repository_ctx, absolute_path):
     return artifact_relative_path
 
 def _get_aar_import_statement_or_empty_str(repository_ctx):
-    if repository_ctx.attr.use_starlark_android_rules:
+    # Use the Starlark version of aar_import is requested, or if this version of Bazel
+    # does not have native aar_import.
+    if repository_ctx.attr.use_starlark_android_rules or getattr(native, "aar_import", None) == None:
```
jin, line 247: Does `hasattr` not work?

private/rules/maven_install.bzl
```diff
@@ -106,6 +107,11 @@ def maven_install(
     if additional_netrc_lines and maven_install_json == None:
         fail("`additional_netrc_lines` is only supported with `maven_install_json` specified", "additional_netrc_lines")
 
+    if getattr(native, "aar_import", None) == None:
+      # If this version of bazel does not have the native version of
```
jin, line 111: nit: fix indent

tests/integration/override_targets/Loader.java
```diff
@@ -0,0 +1,14 @@
+package com.google.ar.sceneform.assets;
+
+/**
+ * Some class in com.google.ar.sceneform:rendering:aar:1.10.0 loads
+ * this class, but there is no class Loader in
+ * com.google.ar.sceneform:assets:1.10.0 or any other aar in sceneform.
+ * This is only to satisfy the ImportDepsChecker check since none of
```
jin, line 7: `..satisfy aar_import's built-in ImportDepsChecker check..`

jin (approved):

MODULE.bazel
ahumesky: https://github.com/bazelbuild/rules_android/commit/f4b6edb93e13ba0bd17ecf7f99c80f4d3d6ca767 is required for the windows tests to pass, which was after the 0.5.0 release

if it would be preferred we can do a 0.5.1 release with the latest code

private/rules/maven_install.bzl
ahumesky: fixed

tests/integration/override_targets/Loader.java
ahumesky: Clarified

MODULE.bazel
ahumesky: we made a 0.5.1 release: https://github.com/bazelbuild/rules_android/releases/tag/v0.5.1
just have to add it to BCR: https://github.com/bazelbuild/bazel-central-registry/pull/2554

ahumesky: 0.5.1 is available on the BCR now so I'll rerun the tests with that version

jin (approved):

Issue #1055, opened 2024-02-16
`use_repo_rule` usage breaks Bazel 5/6 support


luispadron:
`use_repo_rule` is currently only in Bazel 7+ and `rules_jvm_external` 6.0 uses this. It causes the following failure:

```
ERROR: https:/bcr.bazel.build/modules/rules_jvm_external/6.0/MODULE.bazel:65:13: name 'use_repo_rule' is not defined
ERROR: Error computing the main repository mapping: in module dependency chain <root> -> rules_jvm_external@6.0: error executing MODULE.bazel file for rules_jvm_external@6.0
```

shs96c:
This is a known issue, and why the `bzlmod` section of the release notes starts with `If you are using Bazel 7 or above`.

Since Bazel 7 is the latest LTS, `bzlmod` behaviour changed between the 6.5.0 and 7.0.0, and the previous WORKSPACE-based approach still works, we won't be making changes.

PR #1151, opened 2024-05-24
Restore bazel 6.x support

keith:
As far as I can tell all the uses of http_file in the MODULE.bazel are dev dependencies, therefore they're easy to move to the WORKSPACE.bzlmod to avoid using the bazel 7.x only API. I submitted this despite the conversation in the linked issue because rules_jvm_external is sa transitive dependency of tons of repos which are still pinning bazel 6.x. This stems from the fact that protobuf and grpc have this as a dependency.

Fixes https://github.com/bazelbuild/rules_jvm_external/issues/1055

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -1,8 +1,8 @@
 module(
     name = "rules_jvm_external",
-    bazel_compatibility = [">=7.0.0"],
-    compatibility_level = 2,
     version = "6.1",
+    bazel_compatibility = [">=6.0.0"],
+    compatibility_level = 2,
 )
 
 bazel_dep(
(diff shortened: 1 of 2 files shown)


shs96c:
The `coursier` dependency is a production dependency, and needs to be included.

My experience with `bzlmod` is that it behaves sufficiently differently between Bazel 6 and 7 that I'd rather not invest effort in attempting to support `bzlmod` with Bazel 6. For people doing the migration to `bzlmod`, the smoother path (IME) has been to use `--noenable_bzlmod` and update to Bazel 7 first, and then update to using `bzlmod` afterwards.

shs96c:
I notice that I've left this PR open for far too long without making a clear decision. For the sake of simplicity, my stance is that Bazel 6 builds will need to remain as workspace-based builds, and Bazel 7+ will support `bzlmod`. I'm going to close this issue, but please come and find me on Slack if you'd like to discuss.

Issue #1146, opened 2024-05-23
Maven Resolver Ignores BOM Ordering


tharakadesilva:
The current Maven resolver appears to ignore the order of BOMs when resolving dependencies. In Maven, the order of BOMs is significant. When multiple BOMs specify conflicting versions of a dependency, Maven will select the version from the BOM that appears *first* in the project's configuration.

However, `rules_jvm_external` seems to prioritize lower versions regardless of BOM order. This behavior can lead to unexpected and incorrect dependency resolution, potentially causing build failures or runtime issues.

**Expected Behavior:**

The `rules_jvm_external` Maven resolver should respect the order of BOMs, mirroring Maven's default conflict resolution behavior. When encountering conflicting versions, the resolver should prioritize the version from the BOM that appears first.

**Steps to Reproduce:**

1. Include two BOMs in your dependencies, where both BOMs specify conflicting versions of the same dependency (`com.fasterxml.jackson:jackson-bom:2.16.2` and `io.dropwizard:dropwizard-dependencies:2.1.12` should work).
2. Ensure that the BOM containing the desired version is listed *first* (in our example the Jackson BOM).
3. Build the project and observe that the resolver selects the lower version of the conflicting dependency, ignoring the BOM order.

**Additional Information:**

Bazel Version: 6.5.0
rules_jvm_eternal Version: 6.1
Operating System: MacOS

tharakadesilva:
@cheister closing with your changes in: https://github.com/bazelbuild/rules_jvm_external/pull/1176.

PR #1155, opened 2024-05-30
Add support for BOM priotization

tharakadesilva:
With the current code, if multiple BOMs have the same coords, the coordinates would be resolved through all the BOMs when we try to resolve the artifact versions. ~When two or more versions were present, Maven resolver would resort to the lower version (IIUC, default Maven behavior).~ The issue was that the set that was created was not ordered, so it would not give priority to the first defined BOM.

~With these changes, we will ignore the artifact if it was previously found, thus giving priority to the BOM that was defined first.~

We now have a `LinkedHashSet` that will maintain the ordering so that the BOMs will be given priority based on the order they were defined.

---

Closes https://github.com/bazelbuild/rules_jvm_external/issues/1146.

### DIFF
--- a/private/tools/java/com/github/bazelbuild/rules_jvm_external/resolver/maven/MavenResolver.java
+++ b/private/tools/java/com/github/bazelbuild/rules_jvm_external/resolver/maven/MavenResolver.java
@@ -404,7 +404,7 @@ public class MavenResolver implements Resolver {
       RepositorySystemSession session,
       List<RemoteRepository> repositories,
       List<Dependency> boms) {
-    Set<Dependency> managedDependencies = new HashSet<>();
+    Set<Dependency> managedDependencies = new LinkedHashSet<>();
 
     for (Dependency bom : boms) {
       ArtifactDescriptorRequest request =


tharakadesilva:
I am yet to add any tests, will do that once I've received the go-ahead on the strategy.

tharakadesilva:
Tested this as a patch in our repository and the resolution is working as expected! 🎉

private/tools/java/com/github/bazelbuild/rules_jvm_external/resolver/maven/MavenResolver.java
```diff
@@ -404,7 +404,7 @@ private List<Dependency> resolveArtifactsFromBoms(
       RepositorySystemSession session,
       List<RemoteRepository> repositories,
       List<Dependency> boms) {
-    Set<Dependency> managedDependencies = new HashSet<>();
```
cheister, line 407: I think the only change you need for this is
```
Set<Dependency> managedDependencies = new LinkedHashSet<>();
```
So the Set keeps the order of the dependencies added to it.

tharakadesilva: Yep, that was it, updated the PR.

tharakadesilva:
@cheister for the tests, I am thinking of something like this:
* Will be placed in `MavenResolverTest.java` file.
* Do a `MavenRepo.create()`.
* Add two BOMs to it with the same artifact at two different versions.
* Add a coordinate without version.
* See if the version from the first defined BOM resolves.

Do you have any thoughts on this?

cheister:
> 
> Do you have any thoughts on this?

That sounds good to me

PR #1158, opened 2024-06-04
Update java_export.bzl docs

michaelschuett-tomtom:
Fix docs to correctly use no-javadocs instead of no-javadoc which does not work.


shs96c:
This looks great. Thank you for the PR and updating the docs. Could you also please run `./scripts/generate_docs.sh` to make sure that the api reference is also updated?

michaelschuett-tomtom:
hmm i imagine I should see a few more change but this was the output from running that script.

```
(base) schuettm@MAC-LP7FQ66X7J rules_jvm_external % ./scripts/generate_docs.sh                                                                                       
INFO: Analyzed target //scripts:generate_api_reference (68 packages loaded, 2942 targets configured).
INFO: Found 1 target...
Target //scripts:generate_api_reference up-to-date:
  bazel-bin/scripts/api.md
INFO: Elapsed time: 21.987s, Critical Path: 5.29s
INFO: 23 processes: 13 internal, 7 darwin-sandbox, 3 worker.
INFO: Build completed successfully, 23 total actions
```

shs96c:
Thank you for running that. I think that only thing that's left is get the CLA signed, and then I'm more than happy to merge this!

shs96c:
Would you like us to look at this PR again? Do you need a hand getting the CLA signed?

PR #1171, opened 2024-06-15
Fix bzlmod lockfile always not up-to-date

BoleynSu:
The below error will be gone with the fix.
```
MODULE.bazel.lock is no longer up-to-date because: The implementation of the extension 'ModuleExtensionId{bzlFileLabel=@@rules_jvm_external~//:extensions.bzl, extensionName=maven, isolationKey=Optional.empty}' or one of its transitive .bzl files has changed,The repo mappings of certain repos used by the extension 'ModuleExtensionId{bzlFileLabel=@@rules_jvm_external~//:extensions.bzl, extensionName=maven, isolationKey=Optional.empty}' have changed. Please run `bazel mod deps --lockfile_mode=update` to update your lockfile.
```


shs96c (approved):
If this fixes things, that's good, but also a little weird. Good catch! Thank you!

Issue #1161, opened 2024-06-05
FR: Warnings about deviations from the BOM


tharakadesilva:
We might be doing a maven.artifact for a coordinate that is coming from a BOM with the same version. Would be nice to get a warning that tells me to start relying on the bom instead of explicitly defining the version.

cheister:
There is already the [duplicate_version_warning](https://github.com/bazelbuild/rules_jvm_external?tab=readme-ov-file#duplicate-artifact-warning) option to maven install to warn about duplicate artifact declarations.

I think overriding versions from BOM files is different though and probably not a warning.  I put together a PR to log when a BOM version is overridden by an artifact if you use `RJE_VERBOSE=1` is that sufficient for finding these cases?

tharakadesilva:
> There is already the [duplicate_version_warning](https://github.com/bazelbuild/rules_jvm_external?tab=readme-ov-file#duplicate-artifact-warning) option to maven install to warn about duplicate artifact declarations.

Let me tell you a little about the use-case why this came about. We are using the latest Spring Boot BOM (3.3.0) and it has MySQL driver defined at `8.3.0`. The latest MySQL driver (`8.4.0`) brings OTel support and automatically adds spans for tracing and we wanted to use that functionality, so we had an explicit version declaration for it. In the future, I don't want to manage this manually, whenever SB BOM bumps the MySQL version I want to remove the explicit version declaration that I did and start using the BOM for the MySQL driver version.

> I think overriding versions from BOM files is different though and probably not a warning. I put together a PR to log when a BOM version is overridden by an artifact if you use `RJE_VERBOSE=1` is that sufficient for finding these cases?

Can you share a sample of what a log would look like, please?

PR #1172, opened 2024-06-16
Add RJE_VERBOSE output when a BOM version is overridden by an explicit dependency

cheister:
Address https://github.com/bazelbuild/rules_jvm_external/issues/1161 with verbose logging since I don't think overriding a version from dependencyManagement is uncommon in Maven.

### DIFF
--- a/private/tools/java/com/github/bazelbuild/rules_jvm_external/resolver/maven/MavenResolver.java
+++ b/private/tools/java/com/github/bazelbuild/rules_jvm_external/resolver/maven/MavenResolver.java
@@ -368,19 +368,37 @@ public class MavenResolver implements Resolver {
     // Without this, the versions requested in BOMs will be preferred to the
     // one that the user requested
     // First add all the dependencies that have versions
+    Map<String, String> artifactVersions = new HashMap<>();
     Set<String> artifactKeys = new HashSet<>();
     dependencies.stream()
         .map(Dependency::getArtifact)
         .filter(artifact -> artifact.getVersion() != null)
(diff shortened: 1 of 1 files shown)


private/tools/java/com/github/bazelbuild/rules_jvm_external/resolver/maven/MavenResolver.java
```diff
@@ -368,19 +368,37 @@ private List<Dependency> createManagedDependencies(
...
     ImmutableList.Builder<Dependency> managedDependencies = ImmutableList.builder();
     managedDependencies.addAll(dependencies);
 
     // Then add all the BOM dependencies that are not in the declared dependency list
+    boolean rjeVerbose = System.getenv("RJE_VERBOSE") != null;
     bomDependencies.stream()
-        .filter(dep -> !artifactKeys.contains(getArtifactKey(dep.getArtifact())))
+        .filter(
+            dep -> {
+              String artifactKey = getArtifactKey(dep.getArtifact());
```
shs96c, line 391: Nit: This could be done in a `peek`, but I guess here is fine too

shs96c (approved):
LGTM

private/tools/java/com/github/bazelbuild/rules_jvm_external/resolver/maven/MavenResolver.java
cheister: I think you're right, the peek is better.  Added in https://github.com/bazelbuild/rules_jvm_external/pull/1178

PR #1173, opened 2024-06-18
Avoid a split package in the maven bom integration tests

shs96c:


tests/integration/maven_bom/OneDep.java
```diff
@@ -1,4 +1,4 @@
-package com.jvm.external.maven_bom;
+package com.github.bazelbuild.rules_jvm_external.example.maven_bom;
```
cheister, line 1: nit: should `example` be `tests` instead?

cheister (approved):

PR #1176, opened 2024-06-20
Keep BOM's dependencies ordered in the same way they are declared

cheister:
Follow-up to https://github.com/bazelbuild/rules_jvm_external/pull/1155 to add a test.


shs96c (approved):
LGTM. Thank you for adding this!

PR #1178, opened 2024-06-20
Use peek for verbose logging instead of combining it in the filter clause

cheister:
Follow up to comment in https://github.com/bazelbuild/rules_jvm_external/pull/1172.  I think the peek is cleaner so the logging is separated from the filter logic


shs96c (approved):
LGTM

PR #1179, opened 2024-06-20
Consolidate maven versions of classifiers for an artifact

cheister:
We only allow one version for a dependency and all of its classifier in our json file.  If a dependency is declared with different classifiers and different versions then we still want to resolve the same version of the dependency for all of the classifiers.

For example, if you had the dependencies

"io.netty:netty-tcnative-boringssl-static:osx-aarch_64:2.0.61.Final",
"io.netty:netty-tcnative-boringssl-static:2.0.47.Final",

we want only one version to be resolved in the json file

```
"artifacts": {
  "io.netty:netty-tcnative-boringssl-static": {
    "shasums": {
      "jar": "b6f974972c44cd6f9cecabc255290286faac40b6393c66c3c3c0db7f421cc28e",
      "linux-aarch_64": "3614395218ae379cec22ccaa089c4f27b9329a660e0d53c93e7cb12b7a2cee46",
      "linux-x86_64": "4ff9d14f1ec6ccee35b78f53a6f3d9c7c54535aa2a76138311c2f619c5e150e1",
      "osx-aarch_64": "58e0302c9fde3db984c3ff7ee7ec7159dc0320bdb91533cc290e12e40911cd1a",
      "osx-x86_64": "fcfea887f4f0802d363c699b444d504b7109a7cb198ae6845eeff63745e5b0ba",
      "windows-x86_64": "17cd2fa3c63b7ed23edea01c945e55cb7baed1faa0f553732c3f5f56da90b3e0"
    },
    "version": "2.0.61.Final"
  },
```
In this case, we'll use the first version of the dependency we get in the dependency list

### DIFF
--- a/private/tools/java/com/github/bazelbuild/rules_jvm_external/resolver/maven/MavenResolver.java
+++ b/private/tools/java/com/github/bazelbuild/rules_jvm_external/resolver/maven/MavenResolver.java
@@ -402,12 +402,61 @@ public class MavenResolver implements Resolver {
         .filter(dep -> !artifactKeys.contains(getArtifactKey(dep.getArtifact())))
         .forEach(managedDependencies::add);
 
-    return managedDependencies.build();
+    // Last, normalize versions across different classifiers of an artifact
+    // The rules_jvm_external json version file only supports one version for an artifact even when
+    // the artifact has multiple classifiers e.g.
+    //
+    //  "artifacts": {
+    //    "io.netty:netty-tcnative-boringssl-static": {
+    //      "shasums": {
+    //        "jar": "b6f974972c44cd6f9cecabc255290286faac40b6393c66c3c3c0db7f421cc28e",
+    //        "linux-aarch_64": "3614395218ae379cec22ccaa089c4f27b9329a660e0d53c93e7cb12b7a2cee46",
+    //        "linux-x86_64": "4ff9d14f1ec6ccee35b78f53a6f3d9c7c54535aa2a76138311c2f619c5e150e1",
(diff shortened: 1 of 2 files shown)


private/tools/java/com/github/bazelbuild/rules_jvm_external/resolver/maven/MavenResolver.java
```diff
@@ -402,12 +402,61 @@ private List<Dependency> createManagedDependencies(
         .filter(dep -> !artifactKeys.contains(getArtifactKey(dep.getArtifact())))
         .forEach(managedDependencies::add);
 
-    return managedDependencies.build();
+    // Last, normalize versions across different classifiers of an artifact
```
shs96c, line 405: It feels better to pick the most recent version, and to output a `LogEvent` when we find more than one version of something

cheister: I'm not sure, we don't pick the most recent version of a dependency in any other case, would it be confusing if we only picked the most recent version for artifacts with classifiers?   Should we just pick the most recent for all artifacts in the dependency list?

cheister: We would also have to add logic to only update to the latest version if a version for one of the classified dependencies wasn't an explicit dependency listed by the user.  That feels more complicated than the feature is worth.

shs96c: Let's land what we have and iterate.

shs96c (approved):
Ship it!

Issue #1190, opened 2024-07-03
Support in memory PGP signing


ThomasCJY:
**Background**
The export rule only supports `gpg_sign` option which requires the `gpg` binary to be installed on the machine and set up correctly. This doesn't fit the need if we want to build, sign and release artifacts through our CI builder. In that use case, we should use OpenPGP in memory signing and keep the signing key and pwd inside Jenkins job as a secret env variable.

ThomasCJY:
proposed change in: https://github.com/bazelbuild/rules_jvm_external/pull/1191

PR #1191, opened 2024-07-03
[Upstream] Add pgp in memory signing support

ThomasCJY:
**Background**
Address issue https://github.com/bazelbuild/rules_jvm_external/issues/1190
Maven central requires OpenPGP signatures for all the artifacts. Gradle did it through the [signing plugin](https://docs.gradle.org/current/userguide/signing_plugin.html#using_in_memory_ascii_armored_openpgp_subkeys). In Bazel jvm rule the only existing support is using the gpg command line to sign those artifacts which doesn't fit our use case to sign artifacts on CI builder.

**Change**
* Add in memory PGP signing in bazel jvm rules. Mainly followed Gradle implementation https://cs.android.com/android-studio/gradle/+/master:platforms/software/security/src/main/java/org/gradle/security/internal/

**Test**
* Added verification step in the implementation
* test on internal jenkins jobs


ThomasCJY:
@jin @shs96c PTAL

private/tools/java/com/github/bazelbuild/rules_jvm_external/maven/MavenPublisher.java
```diff
@@ -78,6 +103,10 @@ public static void main(String[] args)
     boolean gpgSign = Boolean.parseBoolean(System.getenv("GPG_SIGN"));
     Credentials credentials =
         new BasicAuthCredentials(System.getenv("MAVEN_USER"), System.getenv("MAVEN_PASSWORD"));
+    boolean useInMemoryPgpKeys = Boolean.parseBoolean(System.getenv("USE_IN_MEMORY_PGP_KEYS"));
```
shs96c, line 106: You can use `Boolean.getBoolean`

private/tools/java/com/github/bazelbuild/rules_jvm_external/maven/MavenPublisher.java
```diff
@@ -174,10 +203,27 @@ private static CompletableFuture<Void> upload(
     uploads.add(upload(String.format("%s%s.md5", base, append), credentials, md5));
     uploads.add(upload(String.format("%s%s.sha1", base, append), credentials, sha1));
 
-    if (gpgSign) {
-      uploads.add(upload(String.format("%s%s.asc", base, append), credentials, sign(item)));
-      uploads.add(upload(String.format("%s%s.md5.asc", base, append), credentials, sign(md5)));
-      uploads.add(upload(String.format("%s%s.sha1.asc", base, append), credentials, sign(sha1)));
+    SigningMethod signingMethod = signingMetadata.getSigningMethod();
+    if (signingMethod.equals(SigningMethod.GPG)) {
```
shs96c, line 207: We've lost the code path where there is no signing to be done

private/tools/java/com/github/bazelbuild/rules_jvm_external/maven/MavenPublisher.java
```diff
@@ -174,10 +203,27 @@ private static CompletableFuture<Void> upload(
     uploads.add(upload(String.format("%s%s.md5", base, append), credentials, md5));
     uploads.add(upload(String.format("%s%s.sha1", base, append), credentials, sha1));
 
-    if (gpgSign) {
-      uploads.add(upload(String.format("%s%s.asc", base, append), credentials, sign(item)));
-      uploads.add(upload(String.format("%s%s.md5.asc", base, append), credentials, sign(md5)));
-      uploads.add(upload(String.format("%s%s.sha1.asc", base, append), credentials, sign(sha1)));
+    SigningMethod signingMethod = signingMetadata.getSigningMethod();
```
shs96c, line 206: Given how this class is used, it feels like it might be an interface with three implementations (in-memory, gpg, and "not signing") and two methods ("sign" and "isSigning"). That might make this a little clearer

private/tools/java/com/github/bazelbuild/rules_jvm_external/maven/MavenPublisher.java
```diff
@@ -434,4 +480,171 @@ public boolean hasRequestMetadataOnly() {
     @Override
     public void refresh() {}
   }
+
+  /**
+   * This method is based on the following Gradle code:
```
shs96c, line 485: Thank you for adding this!

shs96c (commented):
Thank you for your patience waiting for this review. It's much appreciated, and I love the idea. I've made a few suggestions for changes.

private/tools/java/com/github/bazelbuild/rules_jvm_external/maven/MavenPublisher.java
ThomasCJY: Seems like the existing code are all using `Boolean.parseBoolean`. I prefer to keep using the same api and tackle it as a follow up in case there's any edge case.

private/tools/java/com/github/bazelbuild/rules_jvm_external/maven/MavenPublisher.java
ThomasCJY: The current implementation is 
```
if (signingMethod.equals(SigningMethod.GPG)) {
    // do somthing
} else if (signingMethod.equals(SigningMethod.PGP)) {
    // do something
}

return CompletableFuture.allOf(uploads.toArray(new CompletableFuture<?>[0]));
```

where signing method is defined in 
```
      if (gpgSign) {
        this.signingMethod = SigningMethod.GPG;
      } else if (useInMemoryPgpKeys) {
        this.signingMethod = SigningMethod.PGP;
      } else {
        this.signingMethod = SigningMethod.NONE;
      }
```

Therefore, we still keeps the no-signing path as before where `signingMethod = SigningMethod.NONE`

private/tools/java/com/github/bazelbuild/rules_jvm_external/maven/MavenPublisher.java
```diff
@@ -434,4 +480,171 @@ public boolean hasRequestMetadataOnly() {
...
+    PGP,
+    NONE
+  }
+
+  private static class SigningMetadata {
+    private final String signingKey;
+    private final String signingPassword;
+    private final SigningMethod signingMethod;
+
+    public SigningMetadata(Boolean gpgSign, Boolean useInMemoryPgpKeys, String signingKey, String signingPassword) {
```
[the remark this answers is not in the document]
ThomasCJY: Are you referring to `SigningMetadata` class? I created it mainly to avoid extending the param list of this upload method and other similar methods. 
```
              upload(
                  repo,
                  credentials,
                  coords,
                  String.format("-%s.%s", classifier, ext),
                  artifact,
                  signingMetadata));
```

private/tools/java/com/github/bazelbuild/rules_jvm_external/maven/MavenPublisher.java
ThomasCJY: This should be doable. However, seems to be a bit overkill due to the following reasons:
1. The change scope will be much larger than the issue this PR was originally planning to address and needs a more thorough test plan.
1. There's no dependency injection support in this module, and most of the signing logic are defined inside static method, which diminish the benefits of defining interface/implementation.
1. Technically speaking, there's only 2 sub-types (gpg, pgp) and most likely we won't add any new signing support in the future. Make not-signing to apply `sign` method along with 2 other use cases also seems counter-intuitive.

ThomasCJY:
@Wyverald @shs96c @jin ptal

private/tools/java/com/github/bazelbuild/rules_jvm_external/maven/MavenPublisher.java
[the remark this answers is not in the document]
shs96c: I mean the use of `Boolean` instead of `boolean`. We never handle the `null` case, so not using the primitive type means that we could get a `NullPointerException` if we unbox `useInMemoryPgpKeys` and `gpgSign` in this method. Using the primitive type prevents this.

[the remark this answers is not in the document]
ThomasCJY: Nice catch, updated

private/tools/java/com/github/bazelbuild/rules_jvm_external/maven/MavenSigning.java
```diff
@@ -0,0 +1,245 @@
...
+    private static PGPPrivateKey createPrivateKey(PGPSecretKey secretKey, String password) {
+        try {
+            PBESecretKeyDecryptor decryptor = new BcPBESecretKeyDecryptorBuilder(new BcPGPDigestCalculatorProvider()).build(password.toCharArray());
+            return secretKey.extractPrivateKey(decryptor);
+        } catch (PGPException e) {
+            throw new RuntimeException("Could not extract private key", e);
+        }
+    }
+
+    protected static Path gpg_sign(Path toSign) throws IOException, InterruptedException {
```
shs96c, line 174: Quirky method naming in java, but I'll fix in a follow up PR.

shs96c (approved):
LGTM. Thank you for adding this!

PR #1192, opened 2024-07-08
Remove hardcoded javadocopts

vinnybod:
I have a situation where I need to disable `-use`, but it was hardcoded into JavadocJarMaker.java`.

I changed the way javadoc options work by making the `javadocopts` attribute completely replace the default options instead of adding to them.

This is not completely backwards compatible. I thought about making it backwards compatible by adding some other flag like `override_default_javadocopts`, but wasn't sure how people would feel about adding additional attributes.

Open to changing this implementation or closing the PR if its not valuable enough.

### DIFF
--- a/private/rules/javadoc.bzl
+++ b/private/rules/javadoc.bzl
@@ -8,6 +8,15 @@ _JavadocInfo = provider(
     },
 )
 
+_default_javadocopts = [
+    "-notimestamp",
+    "-use",
+    "-quiet",
+    "-Xdoclint:-missing",
+    "-encoding",
+    "UTF8",
+]
+
 def generate_javadoc(
         ctx,
         javadoc,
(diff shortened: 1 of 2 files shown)


private/rules/javadoc.bzl
```diff
@@ -107,7 +118,7 @@ javadoc = rule(
         "javadocopts": attr.string_list(
             doc = """javadoc options.
             Note sources and classpath are derived from the deps. Any additional
-            options can be passed here.
+            options can be passed here. If nothing is passed, a default list of options is used.
```
jin, line 121: Can you interpolate the list itself into this docstring?

vinnybod: @jin I made your suggested change. I had to open a new PR due to constraints in my previous fork.
https://github.com/bazelbuild/rules_jvm_external/pull/1212

vinnybod:
See https://github.com/bazelbuild/rules_jvm_external/pull/1212

PR #1195, opened 2024-07-10
Add a `maven.artifact` tag example into the bzlmod readme.

jin:
There was a Google-internal query about how the install and artifact tags interact with each other, and when to use one or the other.


shs96c (approved):
LGTM

PR #1197, opened 2024-07-10
Provide mechanism for freezing deps for other rulesets to consume

shs96c:


shs96c:
This isn't going to be a simple fix. Abandoning since we should just move the world to `bzlmod`

PR #1203, opened 2024-07-24
Handle Bzlmod qualified repo names in coursier unpinned repo implementation

plobsing:
Under Bzlmod, repo names are qualified. Whereas a simple workspace with a single
`@maven` repo would have before had an unpinned repo named `@unpinned_maven`, now
this repo is named something like `@rules_jvm_external~~maven~unpinned_maven`.
    
Using `str.rpartition` by the `~` delimiter enables us to consistently
get the unqualified name in both cases:
    
    >>> "rules_jvm_external~~maven~unpinned_maven".rpartition("~")
    ('rules_jvm_external~~maven', '~', 'unpinned_maven')
    >>> "unpinned_maven".rpartition("~")
    ('', '', 'unpinned_maven')

After this change, the lockfile pinning script, `@maven//:pin` will use the local
machine's Coursier cache (`~/Library/Caches/Coursier` on MacOS) even under
Bzlmod. A populated cache makes version resolutions substantially faster to obtain.

### DIFF
--- a/private/rules/coursier.bzl
+++ b/private/rules/coursier.bzl
@@ -137,6 +137,9 @@ def _is_file(repository_ctx, path):
 def _is_directory(repository_ctx, path):
     return repository_ctx.which("test") and repository_ctx.execute(["test", "-d", path]).return_code == 0
 
+def _is_unpinned(repository_ctx):
+    return repository_ctx.attr.name.startswith("unpinned_")
+
 def _shell_quote(s):
     # Lifted from
     #   https://github.com/bazelbuild/bazel-skylib/blob/6a17363a3c27dde70ab5002ad9f2e29aff1e1f4b/lib/shell.bzl#L49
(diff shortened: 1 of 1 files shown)


Wyverald:
Please try not to rely on the canonical repo name format, which is [unstable](https://bazel.build/external/module#repository_names_and_strict_deps:~:text=Note%20that%20the%20canonical%20name%20format%20is%20not%20an%20API%20you%20should%20depend%20on%20and%20is%20subject%20to%20change%20at%20any%20time). In fact, we're looking to change the `~` to `+` to address https://github.com/bazelbuild/bazel/issues/22865. Instead, you could explicitly pass in the "unpinned-ness" of a repo as a boolean attribute.

plobsing:
> Please try not to rely on the canonical repo name format, which is [unstable](https://bazel.build/external/module#repository_names_and_strict_deps:~:text=Note%20that%20the%20canonical%20name%20format%20is%20not%20an%20API%20you%20should%20depend%20on%20and%20is%20subject%20to%20change%20at%20any%20time). In fact, we're looking to change the `~` to `+` to address [bazelbuild/bazel#22865](https://github.com/bazelbuild/bazel/issues/22865). Instead, you could explicitly pass in the "unpinned-ness" of a repo as a boolean attribute.

Fair enough. I wasn't really a fan of the string munging going on, but ignored it to get the smallest patch that would solve my issue. Your way does seem cleaner to me, if we want to do that, its here: https://github.com/bazelbuild/rules_jvm_external/pull/1204 .
Demoting this PR to draft while that one is in flight (will close if it merges).

PR #1204, opened 2024-07-24
Track whether a repo is pinned or unpinned by explicit attr

plobsing:
Doing this by performing string interpretation on the repo name is a bit
brittle. And it works less well under Bzlmod where the unpinned repo
name is going to look something like
`rules_jvm_external~~maven~unpinned_maven`.

After this change, the lockfile pinning script, `@maven//:pin` will use the local
machine's Coursier cache (`~/Library/Caches/Coursier` on MacOS) even under
Bzlmod. A populated cache makes version resolutions substantially faster to obtain.

### DIFF
--- a/private/extensions/maven.bzl
+++ b/private/extensions/maven.bzl
@@ -412,6 +412,7 @@ def _maven_impl(mctx):
                 # created from the maven_install.json file in the coursier_fetch
                 # invocation after this.
                 name = "unpinned_" + name if repo.get("lock_file") else name,
+		pinned_repo_name = name if repo.get("lock_file") else None,
                 user_provided_name = name,
                 repositories = repo.get("repositories"),
                 artifacts = artifacts_json,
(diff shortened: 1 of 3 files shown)


kolloch:
I tested (but did not review in depth) your changes and it does speed things up signifincatly.

Thank you.

private/extensions/maven.bzl
```diff
@@ -412,6 +412,7 @@ def _maven_impl(mctx):
                 # created from the maven_install.json file in the coursier_fetch
                 # invocation after this.
                 name = "unpinned_" + name if repo.get("lock_file") else name,
+		pinned_repo_name = name if repo.get("lock_file") else None,
```
Wyverald, line 415: could you fix up the indent here?

Wyverald (approved):

private/extensions/maven.bzl
plobsing: Whoops, didn't see that a tab got in here somehow. Good catch. Fixed.

PR #1207, opened 2024-07-26
handle transitive dependencies of packaging=pom

### LINKED ISSUE #1206: a Maven artifact with a transitive dependency on an artifact with packaging=pom does not fully work out of the box
I've tested this with rules_jvm_external 6.2. If I have a `maven_install` which imports a Maven artifact which in turn depends on a packaging=pom type dependency, such as

```
maven_install(
    artifacts = [
        "com.foobar:thing-that-uses-javamoney-moneta:jar:1.2.3",
    ],
    ...
)
```

and the pom.xml for that project in turn looks like:

```xml
  <dependencies>
    ...
    <dependency>
      <groupId>org.javamoney</groupId>
      <artifactId>moneta</artifactId>
      <type>pom</type>
    </dependency>
```

RJE ends up generating an `external/maven/BUILD` with the following target:

```
jvm_import(
	name = "com_foobar_thing_that_uses_javamoney_moneta,
	jars = ["com/foobar/thing-that-uses-javamoney-moneta/1.2.3/...jar"],
	deps = [
		...        
		":org_javamoney_moneta",
	],
        ...
)
```

however `external/maven/BUILD` does not have a target defined for `org_javamoney_moneta`.

Building a target which has a `dep` on `@maven//:com_foobar_thing_that_uses_javamoney_moneta` will yield an error like:

```
$ bazel build //...
ERROR: /private/var/tmp/_bazel_mattbrown/24f8f5a0eeb6b27b916bd8a74f8ebe77/external/maven/BUILD:31168:11: in deps attribute of jvm_import rule @@maven//:com_foobar_thing_that_uses_javamoney_moneta: rule '@@maven//:org_javamoney_moneta' does not exist
ERROR: /private/var/tmp/_bazel_mattbrown/24f8f5a0eeb6b27b916bd8a74f8ebe77/external/maven/BUILD:31168:11: Analysis of target '@@maven//:com_foobar_thing_that_uses_javamoney_moneta' failed
```

---

There is a simple workaround for this, but its a bit non-obvious to setup and it seems like something that RJE could handle on its own. The workaround is to add the packaging=pom artifact to `maven_install`:

```diff
 maven_install(
     artifacts = [
+        "org.javamoney:moneta:1.4.2",
         "com.foobar:thing-that-uses-javamoney-moneta:jar:1.2.3",
     ],
```

note that the packaging must be left out of the string, specifying it as `"org.javamoney:moneta:pom:1.4.2"` yields the original error above.

mattnworb:
closes #1206


mattnworb:
I should probably add a test showing this actually fixes the issue 😀

mattnworb:
This should be ready for review now. I'm not well-versed in testing strategies for rulesets so I tried to mimic what was going on with other tests, and would appreciate any feedback on how to better follow the practices for other tests (for example I couldn't decide if the `@transitive_dependency_with_type_of_pom` repo I added should be pinned or not).

MODULE.bazel
```diff
@@ -707,6 +707,16 @@ dev_maven.install(
...
 )
 
+# https://github.com/bazelbuild/rules_jvm_external/issues/1206
+dev_maven.install(
+    name = "transitive_dependency_with_type_of_pom",
+    # an arbitrary artifact which depends on org.javamoney:moneta:pom
+    artifacts = [
+        # https://github.com/quarkiverse/quarkus-moneta/blob/2.0.0/runtime/pom.xml#L16-L21
+        "io.quarkiverse.moneta:quarkus-moneta:2.0.0",
+    ],
```
shs96c, line 717: Can you please add a lock file for this?

shs96c (approved):
LGTM. Thank you for the patch, and for your patience with the review!

Issue #1154, opened 2024-05-29
java_export generated pom.xml does not include <classifier> property


vinnybod:
I am encountering this with a `java_export` that contains `io.netty:netty-tcnative-boringssl-static`. I can provide a repo example if it helps.

To reproduce, create a `java_export` with any/all of the following artifacts:
```
"@maven//:io_netty_netty_tcnative_boringssl_static_linux_aarch_64",
"@maven//:io_netty_netty_tcnative_boringssl_static_linux_x86_64",
"@maven//:io_netty_netty_tcnative_boringssl_static_osx_aarch_64",
"@maven//:io_netty_netty_tcnative_boringssl_static_osx_x86_64",
"@maven//:io_netty_netty_tcnative_boringssl_static_windows_x86_64",
```

The generated pom.xml contains the group, artifact, scope, and version, but is missing the classifier. What I would expect:
```
<dependency>
  <artifactId>netty-tcnative-boringssl-static</artifactId>
  <classifier>linux-aarch_64</classifier>
  <groupId>io.netty</groupId>
  <scope>runtime</scope>
  <version>2.0.61.Final</version>
</dependency>
```

The `maven_coordinates` tag does contain the classifier, but the `unpack_coordinates` function does not account for it: https://github.com/bazelbuild/rules_jvm_external/blob/master/private/rules/maven_utils.bzl#L7
```
maven_coordinates=io.netty:netty-tcnative-boringssl-static:jar:linux-aarch_64:2.0.61.Final
```

I'd be happy to put up a PR with a bit of guidance. The maven coordinate string in the javadoc linked above seems to follow a different syntax than what is present in the tag.
`groupId:artifactId[:type[:scope]]:version` while the tag looks to be `groupId:artifactId[:type[:classifier]]:version` without a scope.

PR #1208, opened 2024-07-30
Add tests for pom generation

vinnybod:
This adds tests for various pom generation scenarios.

---
Original context:
This fixes https://github.com/bazelbuild/rules_jvm_external/issues/1154

It seems like a bug that scope was being considered in the maven coordinates. I couldn't find any examples of that being the case. But it is common for `group:artifact:type:classifier:version`.

Demo of the issue: [vinnybod/bazel_java_example@classifier-pom](https://github.com/vinnybod/bazel_java_example/tree/classifier-pom?rgh-link-date=2024-07-30T18%3A21%3A17Z)

Because bom generation was the only thing depending on sending the scope as part of the GAV, I refactored `maven_utils.generate_pom` so `versioned_dep_coordinates` and `unversioned_dep_coordinates` are expected to already be `unpacked` and `_maven_dependencies_bom` sets the type/scope itself. I also added tests.


vinnybod:
I am seeing that this is failing the bom generation tests. WIll take a look.

private/rules/maven_utils.bzl
```diff
@@ -1,10 +1,10 @@
 def unpack_coordinates(coords):
     """Takes a maven coordinate and unpacks it into a struct with fields
-    `groupId`, `artifactId`, `version`, `type`, `scope`
+    `groupId`, `artifactId`, `version`, `type`, `classifier`
```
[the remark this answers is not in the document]
vinnybod: > since I know that are plenty of folks out there who use this format for their test scoped dependencies

Can you provide an example of how `unpack_coordinates` is used in those cases?

I do think `groupId:artifactId[:type[:classifier]]:version` is the "correct" way to do things, but if there's concerns about keeping things compatible, I'd like to see how its being used. I think I covered all the use cases *within* the `rules_jvm_external` repo.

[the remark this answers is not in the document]
shs96c: I've got a version of `unpack_coordinates` that preserves the existing format and also supports the Gradle "external dependency" format [in this PR](https://github.com/bazelbuild/rules_jvm_external/pull/1232/files#diff-28225bfdb29cfcfd891252f2c3fdc998f5acd86912756acae7780659b5009fb1).

It's currently used to help parse the `artifacts` and `boms` from `maven_install`.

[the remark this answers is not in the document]
vinnybod: I'll wait for that PR to land, then I will reimplement this in top of it.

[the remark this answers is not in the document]
shs96c: The PR landed!

private/lib/coordinates.bzl
```diff
@@ -5,7 +5,7 @@ def unpack_coordinates(coords):
 
     Assumes `coords` is in one of the following syntaxes:
      * groupId:artifactId[:type[:scope]]:version
-     * groupId:artifactId[:version][@classifier][:type]
+     * groupId:artifactId[:version][:classifier][@type]
```
vinnybod, line 8: I think this was a typo according to https://docs.gradle.org/current/dsl/org.gradle.api.artifacts.dsl.DependencyHandler.html#N1737E

private/rules/maven_utils.bzl
```diff
@@ -85,13 +92,20 @@ def generate_pom(
     deps = []
     for dep in sorted(versioned_dep_coordinates) + sorted(unversioned_dep_coordinates):
         include_version = dep in versioned_dep_coordinates
-        unpacked = _unpack_coordinates(dep)
+        split = dep.split(":")
+        if len(split) == 5:
+            gradle_format = split[0] + ":" + split[1] + ":" + split[4] + ":" + split[3] + "@" + split[2]
+            print(st)
+            unpacked = _unpack_coordinates(st)
```
vinnybod, line 99: This probably needs to be more robust?

shs96c:
It looks like the test failures are genuine.

private/rules/maven_utils.bzl
```diff
@@ -85,13 +92,20 @@ def generate_pom(
     deps = []
     for dep in sorted(versioned_dep_coordinates) + sorted(unversioned_dep_coordinates):
         include_version = dep in versioned_dep_coordinates
-        unpacked = _unpack_coordinates(dep)
+        split = dep.split(":")
+        if len(split) == 5 and split[3] not in ["import", "compile", "runtime", "test", "provided", "system"]:
```
vinnybod, line 96: This seems hacky.

private/rules/pom_file.bzl
```diff
@@ -1,6 +1,6 @@
 load("@rules_java//java:defs.bzl", "JavaInfo")
 load(":has_maven_deps.bzl", "MavenInfo", "calculate_artifact_jars", "has_maven_deps")
-load(":maven_utils.bzl", "determine_additional_dependencies", "generate_pom")
+load(":maven_utils.bzl", "determine_additional_dependencies", "generate_pom", "unpack_coordinates")
```
shs96c, line 3: This appears to be used. Could you please run `bazel run scripts:format`?

shs96c (changes requested):
I think we're almost there :)

private/rules/maven_utils.bzl
vinnybod: I had to convert it to the gradle format for it to be parsed correctly. If I just pass it in as-is, then `unpack_coordinates` is putting the `classifier` into the `<scope>` tag.
So maybe this hardcoded check for `"import", "compile", "runtime", "test", "provided", "system"` should go into the `unpack_coordinates`?

I feel like I might be missing something obvious if you are saying the heuristic check is already in place.

vinnybod:
@shs96c It looks like https://github.com/bazel-contrib/rules_jvm_external/pull/1279 fixes my original complaint 😄 :
>It seems like a bug that scope was being considered in the maven coordinates. I couldn't find any examples of that being the case. But it is common for group:artifact:type:classifier:version.

I merged master into this branch and now all it does is add some new tests.

vinnybod:
Closing because I added these tests into https://github.com/bazel-contrib/rules_jvm_external/pull/1336

PR #1210, opened 2024-08-04
Derive jvm_import debug label from explicit attribute

plobsing:
The delimiter used in Bzlmod for repo names has never been stable and
will be changing soon.[^1] String-parsing to derive shorter,
user-friendly names for labels will be broken by this change and is
brittle in general. Instead, passing the right values down from the
contexts where the true values are known is preferred.
    
[^1]: https://github.com/bazelbuild/bazel/issues/23127

### DIFF
--- a/private/rules/jvm_import.bzl
+++ b/private/rules/jvm_import.bzl
@@ -14,10 +14,10 @@ def _jvm_import_impl(ctx):
     if len(ctx.files.jars) != 1:
         fail("Please only specify one jar to import in the jars attribute.")
 
-    label = "@{workspace_name}//{package}:{name}".format(
+    label = "@{repo}//{package}:{name}".format(
         name = ctx.label.name,
         package = ctx.label.package,
-        workspace_name = ctx.attr.user_provided_repo_name,
+        repo = ctx.attr.user_provided_repo_name,
     )
 
     injar = ctx.files.jars[0]
--- a/private/rules/jvm_import.bzl
+++ b/private/rules/jvm_import.bzl
@@ -14,13 +14,10 @@ def _jvm_import_impl(ctx):
     if len(ctx.files.jars) != 1:
         fail("Please only specify one jar to import in the jars attribute.")
 
-    # With `bzlmod` enabled, workspace names end up being `~` separated. For the
-    # user-visible workspace name, we need the final part of the name
-    visible_name = ctx.label.workspace_name.rpartition("~")[2]
     label = "@{workspace_name}//{package}:{name}".format(
         name = ctx.label.name,
         package = ctx.label.package,
-        workspace_name = visible_name,
+        workspace_name = ctx.attr.user_provided_repo_name,
     )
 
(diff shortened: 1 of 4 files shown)


private/rules/jvm_import.bzl
```diff
@@ -14,13 +14,10 @@ def _jvm_import_impl(ctx):
     if len(ctx.files.jars) != 1:
         fail("Please only specify one jar to import in the jars attribute.")
 
-    # With `bzlmod` enabled, workspace names end up being `~` separated. For the
-    # user-visible workspace name, we need the final part of the name
-    visible_name = ctx.label.workspace_name.rpartition("~")[2]
     label = "@{workspace_name}//{package}:{name}".format(
```
Wyverald, line 17: in the interest of keeping names up to date, let's replace uses of `workspace_name` here with `repo_name` (or just `repo`)

private/rules/maven_install.bzl
```diff
@@ -123,6 +123,7 @@ def maven_install(
             # created from the maven_install.json file in the coursier_fetch
             # invocation after this.
             name = name if maven_install_json == None else "unpinned_" + name,
+            user_provided_name = name,
```
jin, line 126: Instead of adding another parameter, why not derive this from `pinned_repo_name`, and if it's `None` or not an attr in the case of `pinned_coursier_fetch`, use `name`?

plobsing: Inside a repo rule,`repository_ctx.attr.name` is the fully-qualified repo name; it is no longer the shortname passed when declaring the repo. Using that value yields the following test failure:

```
INFO: From Testing //tests/unit/manifest_stamp:diff_manifest_test:
==================== Test output for //tests/unit/manifest_stamp:diff_manifest_test:
2c2
< Target-Label: @manifest_stamp_testing//:com_google_guava_guava
---
> Target-Label: @_main~maven~manifest_stamp_testing//:com_google_guava_guava
FAIL: files "tests/unit/manifest_stamp/guava_MANIFEST.MF.golden.unix" and "tests/unit/manifest_stamp/guava_MANIFEST.MF" differ. 
```

IIUC, the preferred way to deal with this is to pass a second string-valued attr, that gets the same value as `name`, at the declaration site; since this attr is user-declared, it doesn't have any platform magic associated with it - it keeps the same value into the repo rule.

So, the way I see it, we have a need for two user-declared attrs: the shortname of this repo, and the shortname of the assocaited pinned repo (in the case of unpinned repos). I could be wrong, and maybe there's some way to get by without yet another attr here, but I couldn't work out how to do it - the magic associated with the `name` attr interferes with what we're trying to do.

shs96c:
I've opened #1224 to help make it easier to get the visible name of a repo.

private/rules/maven_install.bzl
plobsing: Neat. Should we go with that change instead? Happy to close this if we do; I only care about the result.

PR #1212, opened 2024-08-06
Remove hardcoded javadocopts

vinnybod:
I have a situation where I need to disable `-use`, but it was hardcoded into `JavadocJarMaker.java`.

I changed the way javadoc options work by making the `javadocopts` attribute completely replace the default options instead of adding to them.

This is not completely backwards compatible. I thought about making it backwards compatible by adding some other flag like `override_default_javadocopts`, but wasn't sure how people would feel about adding additional attributes.

Open to changing this implementation or closing the PR if its not valuable enough.

Note: This is a copy of https://github.com/bazelbuild/rules_jvm_external/pull/1192 . I had to reopen the PR from a different Github fork because of new constraints in the previous one.


jin:
I think this is reasonable for folks to override the default instead of the rule always adding hardcoded defaults. @shs96c wdyt?

private/rules/javadoc.bzl
```diff
@@ -65,12 +74,14 @@ def _javadoc_impl(ctx):
     # `None` https://github.com/bazelbuild/bazel/issues/10170). For this
     # reason we allow people to set javadocopts via the rule attrs.
 
+    javadocopts = ctx.attr.javadocopts if ctx.attr.javadocopts else _default_javadocopts
```
shs96c, line 77: Why not just set the default value of the attribute to be `_default_javadocopts`? Then you can continue to use `ctx.attr.javadocopts` throughout.

private/rules/javadoc.bzl
```diff
@@ -8,6 +8,15 @@ _JavadocInfo = provider(
     },
 )
 
+_default_javadocopts = [
```
shs96c, line 11: This is a constant, so please name in all caps.

shs96c (commented):
Thank you for your patience with this review. 

One small change requested, but LGTM. Could you also please run `./scripts/generate_docs.sh` since you're updating the stardoc too.

private/rules/javadoc.bzl
vinnybod: That makes sense to me

vinnybod:
After making the suggested changes and merging master, some tests are failing. will have to look into it later

shs96c:
@vinnybod, please tag me when you'd like me to have another look at this.

vinnybod:
@shs96c all tests are now passing, this is ready for review

jin (approved):

PR #1224, opened 2024-08-27
Extract utility methods for handling bzlmod mangled names into one place

### LINKED ISSUE #1219: buildozer message incorrect with --incompatible_use_plus_in_repo_names
When bazel is run with --incompatible_use_plus_in_repo_names, repository names change, and the buildozer message that is emitted by rules_jvm_external is incorrect:

```
 ** Please add the following dependencies:
  @rules_jvm_external++maven+maven//:io_rest_assured_json_path @rules_jvm_external++maven+maven//:org_apache_groovy_groovy to //src/path/to:target
 ** You can use the following buildozer command:
buildozer 'add deps @rules_jvm_external++maven+maven//:io_rest_assured_json_path @rules_jvm_external++maven+maven//:org_apache_groovy_groovy' //src/path/to:target
```

See: https://github.com/bazelbuild/bazel/issues/23127

This is on windows, with rules_jvm_external 6.2 and bazel 7.3.1

shs96c:
This allows us to cope with `--incompatible_use_plus_in_repo_names`

Closes #1219


cheister (approved):

private/rules/jvm_import.bzl
```diff
@@ -8,15 +8,14 @@
...
 load("//settings:stamp_manifest.bzl", "StampManifestProvider")
 
 def _jvm_import_impl(ctx):
     if len(ctx.files.jars) != 1:
         fail("Please only specify one jar to import in the jars attribute.")
 
-    # With `bzlmod` enabled, workspace names end up being `~` separated. For the
-    # user-visible workspace name, we need the final part of the name
-    visible_name = ctx.label.workspace_name.rpartition("~")[2]
+    visible_name = to_visible_name(ctx.label.workspace_name)
```
Wyverald, line 18: @fmeum can this usage of `to_visible_name` be avoided? the label is added as _part_ of a string arg.

Wyverald (approved):
I'm still uneasy about these assumptions being used, but better that they work with `+` too than if they don't.

shs96c:
> I'm still uneasy about these assumptions being used, but better that they work with `+` too than if they don't.

Is there a better way of doing this? The two tasks we're trying to accomplish are:

1. Figure out the visible name. Turns out we need more than a `print` to do this since we're embedding a string.
2. Determine the human-readable owning repo name of a dependency.

private/lib/bzlmod.bzl
```diff
@@ -0,0 +1,32 @@
...
+    if "+" in name:
+        return name.rpartition("+")[-1]
+
+    if "~" in name:
+        return name.rpartition("~")[-1]
+
+    return name
+
+def get_file_owner_repo_name(name):
+    """Convert a name (typically from `File.owner.workspace_name`) to the owning repo
```
fmeum, line 22: This doesn't return the name of a repo, it gives you the name of the module that owns the repo that owns the file.

Note that this can have surprising effects: If my `better_repo_rules` module offers a `better_http_archive` module extension, then the reported owner of files in such archives will always be `better_repo_rules`, not the module that specified tags for the extension.

@Wyverald What do you think of exposing a `module_name` field on `Label`? That wouldn't require any repo unmapping, which is nice.

private/rules/jvm_import.bzl
fmeum: (can't comment on the lines below)

```patch
-progress_message = "Stamping the manifest of %s" % ctx.label,
+progress_message = "Stamping the manifest of %{label}",
```

```patch
-args.add_all(["--manifest-entry", "Target-Label:{target_label}".format(target_label = label)])
+args.add("--manifest-entry", label, format = "Target-Label:%s")
```

Both should give you apparent labels with Bazel 7. The are also more memory efficient as the formatted string isn't retained.

.bazelversion
```diff
@@ -1 +1 @@
-7.1.0
+7.3.0
```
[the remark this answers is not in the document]
shs96c: Done

examples/bzlmod/.bazelversion
```diff
@@ -1 +1 @@
-7.1.0
+7.3.0
```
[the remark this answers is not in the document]
shs96c: Done

private/lib/bzlmod.bzl
shs96c: I've updated the method name and comment.

private/rules/jvm_import.bzl
shs96c: Done.

private/rules/maven_project_jar.bzl
```diff
@@ -20,9 +21,7 @@ def _strip_excluded_workspace_jars(jar_files, excluded_workspaces):
             workspace_name = owner.workspace_name
 
             # bzlmod module names use ~ as a separator
```
fmeum, line 23: This is outdated now.

fmeum (approved):

PR #1226, opened 2024-09-03
Upgrade rules_java to 7.10.0

comius:
rules_java 7.10.0 work with Bazel 7 down to Bazel 6.2.0. This makes it possible to load symbols like JavaInfo from rules_java also on older Bazel versions (previous Java releases didn't have Java/common/java_info.bzl file). A single rules_java version for multiple Bazel releases should also simplify things.

Remove the flag --release, because it breaks on the combination of Bazel 6.x and rules_java 7.10.0, most likely because of incompatible changes in the turbine. The flag is not necessary, because the release is already controlled --java_release_version flag.

PR #1231, opened 2024-09-04
Test release version

comius:


private/tools/java/com/github/bazelbuild/rules_jvm_external/ByteStreams.java
```diff
@@ -23,7 +23,7 @@ public static byte[] toByteArray(InputStream source) throws IOException {
   public static void copy(InputStream source, OutputStream sink) throws IOException {
     byte[] buffer = new byte[COPY_BUFFER_SIZE];
 
-    for (int read = source.read(buffer); read != -1; read = source.read(buffer)) {
+    for (var read = source.read(buffer); read != -1; read = source.read(buffer)) {
```
shs96c, line 26: Our users still expect Java 11 to work. Please revert this.

shs96c (changes requested):
This looks like a mash-up of #1226 and something else. Can we please keep this PR focused on one thing?

private/tools/java/com/github/bazelbuild/rules_jvm_external/ByteStreams.java
shs96c: I stand corrected. `var` was in Java 10. This should be fine.

PR #1232, opened 2024-09-04
Make our rules output `PackageInfo` where possible

shs96c:
The Bazel community is heading towards using `@rules_license` as the basis for SBOM generation. Providing `PackageInfo` for the dependencies we import from Maven provides a part of the puzzle of how to make this work.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -12,6 +12,7 @@ bzl_library(
     srcs = [
         ":defs.bzl",
         ":specs.bzl",
+        "@rules_license//:docs_deps",
     ],
     visibility = [
         # This library is only visible to allow others who depend on
--- a/private/dependency_tree_parser.bzl
+++ b/private/dependency_tree_parser.bzl
@@ -28,6 +28,7 @@ load(
     "strip_packaging_and_classifier",
     "strip_packaging_and_classifier_and_version",
 )
+load("//private/lib:coordinates.bzl", "unpack_coordinates")
 
 def _genrule_copy_artifact_from_http_file(artifact, visibilities):
     http_file_repository = escape(artifact["coordinates"])
@@ -121,7 +122,7 @@ def _generate_target(
     #
     is_dylib = False
     if packaging == "jar":
-        target_import_string.append("\tjars = [\"%s\"]," % artifact_path)
+        target_import_string.append("\tjar = \"%s\"," % artifact_path)
         if srcjar_paths != None and target_label in srcjar_paths:
             target_import_string.append("\tsrcjar = \"%s\"," % srcjar_paths[target_label])
     elif packaging == "aar":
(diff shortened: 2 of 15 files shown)


shs96c:
This will cause the generated build files to do one of two things. For `jvm_import` rules, we can now generate a `PackageInfo` directory, and the metadata we need is added to that. For all other rules, we generate a `package_info` target and associate with it with the actual targets being generated. We end up with something like:

```starlark
jvm_import(
        name = "example1",
        jar = "path/to/some.jar",
        maven_coordinates = "com.example:package:1.0.0",
        maven_url = "https://repo1.maven.org/maven2/com/example/package/1.0.0/package-1.0.0.jar",
)

package_info(
        name = "example2_package_info",
        package_name = "com.example:package2:aar:1.0.0",
        package_url = "https://repo1.maven.org/maven2/com/example/package2/1.0.0/package-1.0.0.aar",
        package_version = "1.0.0",
)

aar_import(
        name = "example2",
        aar = "path/to/some/thing/else.aar",
        applicable_licenses = [":example2_package_info"],
)
```

The `java_export` rule also generates a `PackageInfo`, so should be picked up by SBOM generators too.

BUILD
```diff
@@ -12,33 +12,7 @@ bzl_library(
...
-        "//private/rules:maven_project_jar.bzl",
-        "//private/rules:maven_publish.bzl",
-        "//private/rules:maven_utils.bzl",
-        "//private/rules:pin_dependencies.bzl",
-        "//private/rules:pom_file.bzl",
-        "//private/rules:urls.bzl",
-        "//private/rules:v1_lock_file.bzl",
-        "//private/rules:v2_lock_file.bzl",
-        "//settings:stamp_manifest.bzl",
+        "@rules_license//:docs_deps",
```
aiuto, line 15: Sigh. The circular loop of skylib StarlarkLibraryInfo being needed in a rule set that wants not deps is a problem. This is a case where bazel needs a conditional.

What I need is a module extension to detect that skylib is loaded already, and if so, create a local repo that provide bzl_library, either pointing to the real one or a fake one.

That makes more sense in my head than it probably does on paper.

aiuto (commented):
This generally sounds right.
Do you need rules_license in MODULE.bazel?

shs96c:
[We've already added `rules_license` to the `MODULE.bazel` file](https://github.com/bazelbuild/rules_jvm_external/pull/1232/files#diff-6136fc12446089c3db7360e923203dd114b6a1466252e71667c6791c20fe6bdcR23-R26)

comius:
@jin friendly ping

alexeagle:
Is there any way to observe the result by running one of the aspects in rules_license to gather the PackageInfo? Or is more wiring needed for that?

I'm trying to add similar support for other languages but rules_license has no end-user docs I can find.

private/dependency_tree_parser.bzl
```diff
@@ -121,7 +122,7 @@ def _generate_target(
     #
     is_dylib = False
     if packaging == "jar":
-        target_import_string.append("\tjars = [\"%s\"]," % artifact_path)
```
cheister, line 124: This is generating the jvm_import statement right?  I don't see where jvm_import can take `jar` instead of `jars` ? https://github.com/bazelbuild/bazel/blob/master/src/main/starlark/builtins_bzl/common/java/java_import.bzl#L264

cheister: nvm, this is for a different java_import that we define

cheister (approved):

private/rules/jvm_import.bzl
```diff
@@ -8,13 +8,26 @@
...
 
 load("@rules_java//java:defs.bzl", "JavaInfo")
+load("@rules_license//rules:providers.bzl", "PackageInfo")
+load("//private/lib:coordinates.bzl", "to_external_form", "to_purl", "unpack_coordinates")
+load("//private/lib:urls.bzl", "scheme_and_host")
 load("//settings:stamp_manifest.bzl", "StampManifestProvider")
 
 def _jvm_import_impl(ctx):
-    if len(ctx.files.jars) != 1:
+    if not ctx.attr.jar and not ctx.attr.jars:
```
jin, line 17: nit: this should probably be an independent change.

shs96c: I'll break it out.

shs96c: https://github.com/bazelbuild/rules_jvm_external/pull/1242

PR #1235, opened 2024-09-04
Make our extension implementation and tag classes public

shs96c:
This allows other people to extend them as necessary, which is useful for meta-rulesets that aggregate other rulesets.


gibfahn (approved):

jin:
Can you provide a bit more context? Are these rulesets re-exporting `rules_jvm_external` functionality?

jin:
Thanks for the additional context! Do you think it will be helpful to provide documentation on how users can do that in the README, or even better, code examples that we can put into CI?

jin (approved):

PR #1182, opened 2024-06-26
Add support for resources in javadoc jars and make variable substitution in javadoc arguments

vinnybod:
There's two use cases I'm trying to account for here.

1. I need to add resources to the javadoc jar such as a `LICENSE` file.
2. I need to send a `--define` variable to the javadoc generator so that the `maven_version` variable I have set will appear in the `<title>` of the generated html. rules_jvm_external already supports the same type of substitution on the `coordinates` attribute.

```
java_export(
    name = "lib",
    maven_coordinates="com.example:lib:$(maven_version)",
    ...
    javadocopts=["-windowtitle", "$(maven_version)"],
    doc_resources = ["//:LICENSE"]
)
```

PR #1238, opened 2024-09-07
Add support for resources in javadoc jars and make variable substitution in javadoc arguments

vinnybod:
Note: This is a dupe of https://github.com/bazelbuild/rules_jvm_external/pull/1182 . I had to reopen the PR from a different GH org.

---

There's two use cases I'm trying to account for here.

1. I need to add resources to the javadoc jar such as a `LICENSE` file.
2. I need to send a `--define` variable to the javadoc generator so that the `maven_version` variable I have set will appear in the `<title>` of the generated html. rules_jvm_external already supports the same type of substitution on the `coordinates` attribute.

```
java_export(
    name = "lib",
    maven_coordinates="com.example:lib:$(maven_version)",
    ...
    javadocopts=["-windowtitle", "$(maven_version)"],
    doc_resources = ["//:LICENSE"]
)
```


private/rules/javadoc.bzl
```diff
@@ -33,15 +34,27 @@ def generate_javadoc(
...
         args.add("-linkoffline")
         args.add(dep_info.url)
         args.add(dep_info.element_list.dirname)
         inputs.append(dep_info.element_list)
-    args.add_all(javadocopts)
+
+    processed_opts = []
+    for opt in javadocopts:
+        if "$(" in opt:
+            processed_opt = ctx.expand_make_variables("javadocopts", opt, ctx.var)
```
shs96c, line 52: Is there any harm in just expanding the make variables unconditionally? In the common case, I'd expect it to be a no-op, and it makes the code a little easier to read.

tests/com/github/bazelbuild/rules_jvm_external/BUILD
```diff
@@ -105,3 +105,14 @@ java_test(
         artifact("org.hamcrest:hamcrest"),
     ],
 )
+
+java_library(
+    name = "ZipUtils",
```
shs96c, line 110: `zip_utils`. We save CamelCase for `java_binary` targets.

shs96c (changes requested):
Two small nits, and then I'll merge. Thank you for re-opening!

private/rules/javadoc.bzl
vinnybod: Seems to work!

vinnybod:
@shs96c Made the suggested changes, this is ready for another review. Thanks!

shs96c (approved):
LGTM. Thank you for making the changes!

PR #1239, opened 2024-09-07
skip non-class files in the exclusions processing for MergeJars

vinnybod:
Note: This is a dupe of https://github.com/bazelbuild/rules_jvm_external/pull/1125 . I had to reopen the PR from a different GH org.

---

This is sort of an addendum to https://github.com/bazelbuild/rules_jvm_external/issues/1106 and https://github.com/bazelbuild/rules_jvm_external/pull/1107
In the previous PR, I added an exception for `COPYRIGHT` and `NOTICE` since it was the same kind of use case as the existing `LICENSE` exception.

The issue that I am having is that I have files that should be present in all of the jars being published in a repo that are not `COPYRIGHT`, `NOTICE`, or `LICENSE`. For example a `project-version.properties` file that contains info about the build.

Changing the `readExcludedFileNames` function to only process `*.class` files allows for these files to propagate to the final jar. It also still passes all existing test cases.

The only test for this functionality is specifically looking at `.class` files.https://github.com/bazelbuild/rules_jvm_external/blob/master/tests/com/github/bazelbuild/rules_jvm_external/jar/MergeJarsTest.java#L221

I am open to other solutions, but this seems to have fixed a lot of issues for my use case without breaking any existing functionality that is being tested for.


shs96c (approved):
Let's see what happens. LGTM.

PR #1240, opened 2024-09-07
Add sha256 and sha512 to MavenPublisher

vinnybod:
As part of a migration to Bazel, I need the artifacts I am publishing with Bazel to match what we were getting with Gradle as closely as I can. I noticed that Gradle is publishing SHA-256/SHA-512 already by default.


shs96c (approved):
Interesting. Thank you. LGTM

PR #1242, opened 2024-09-11
Allow `jvm_import` to accept a `jar` parameter

shs96c:
We only allow a single `srcjar`, and the `jars` parameter is already checked to ensure the length is 1.

This also deprecates the original `jars` parameter.

### DIFF
--- a/private/rules/jvm_import.bzl
+++ b/private/rules/jvm_import.bzl
@@ -9,12 +9,23 @@
 
 load("@rules_java//java:defs.bzl", "JavaInfo")
 load("//settings:stamp_manifest.bzl", "StampManifestProvider")
+load(":maven_utils.bzl", "unpack_coordinates")
 
 def _jvm_import_impl(ctx):
-    if len(ctx.files.jars) != 1:
+    if not ctx.attr.jar and not ctx.attr.jars:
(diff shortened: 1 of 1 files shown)


private/rules/jvm_import.bzl
```diff
@@ -55,6 +66,10 @@ def _jvm_import_impl(ctx):
         progress_message = "Creating compile jar for %s" % ctx.label,
     )
 
+    additional_providers = []
```
jin, line 69: unrelated?

private/rules/jvm_import.bzl
```diff
@@ -91,6 +109,12 @@ jvm_import = rule(
         "neverlink": attr.bool(
             default = False,
         ),
+        "maven_coordinates": attr.string(
```
jin, line 112: unrelated?

jin (approved):

private/rules/jvm_import.bzl
shs96c: Yes. Removed.

private/rules/jvm_import.bzl
shs96c: Yes. Removed.

jin (approved):

PR #1257, opened 2024-09-27
Add scala_export

vinnybod:
This adds a `scala_export` macro that works the same way as `java_export` and `kt_jvm_export`.
I'd like to add the `scaladoc` classifier artifact generation in a follow up PR.


shs96c:
I'm wondering whether we should push both the `kotlin_export` and `scala_export` to their respective rulesets. We're introducing a circular dependency between those rulesets and `rules_jvm_external`, and we'd require people who are using these targets to have those rulesets available to them anyway.

Could you please open this PR against `rules_scala`, and I'll do the same for the kotlin rule.

PR #1260, opened 2024-10-03
Add support for publishing maven-metadata.xml

fzakaria:
Maven repositories normally have a maven-metadata.xml file that indicate to the Maven system what versions are available and which is to be considered the latest version.

```xml
<metadata modelVersion="1.1.0">
    <groupId>com.mycompany.app</groupId>
    <artifactId>my-app</artifactId>
    <versioning>
        <latest>1.0</latest>
        <release>1.0</release>
        <versions>
            <version>1.0</version>
        </versions>
        <lastUpdated>20200731090423</lastUpdated>
    </versioning>
</metadata>
```

At Confluent, we use AWS Code Artifactory which does not mark a Maven package as "published" unless a new maven-metadata.xml is uploaded indicating so.

* Add support for reading existing maven-metadata.xml
* Add support for adding the new version to the metadata object
* Add support to upload the file for http & file protocols
* Add small test cases to validate SerDe for Maven metadata object from XML

Co-authored-by: Vince Rose <vrose@confluent.io>
Co-authored-by: Na Lou <nlou@confluent.io>


fzakaria:
@shs96c we added a flag to keep it backwards compatible, but we are of the opinion the default should be true.
Let us know what you think.

private/tools/java/com/github/bazelbuild/rules_jvm_external/maven/MavenPublisher.java
```diff
@@ -150,6 +188,92 @@ private static boolean isSchemeSupported(String repo) {
...
+                                .distinct()
+                                .collect(Collectors.toList())
+                );
+                metadata.setVersioning(versioning);
+
+                Path newMavenMetadataXml = Files.createTempFile("maven-metadata", ".xml");
+                ByteArrayOutputStream os = new ByteArrayOutputStream();
+                new MetadataXpp3Writer().write(os, metadata);
+                Files.write(newMavenMetadataXml, os.toByteArray());
+                return upload(mavenMetadataUrl, credentials, newMavenMetadataXml);
```
fzakaria, line 270: nit: upload could probably take InputStream but I didn't want to change too much of the code; 
Open to doing another pass after on this file if you are open to it.

private/tools/java/com/github/bazelbuild/rules_jvm_external/maven/MavenPublisher.java
```diff
@@ -261,6 +385,60 @@ private static String toHexS(String fmt, String algorithm, byte[] toHash) {
     }
   }
 
+  /**
+   * Attempts to download the file at the given targetUrl.
+   * Valid protocols are: http(s) & file at the moment.
+   */
+  private static CompletableFuture<Optional<String>> download(
```
fzakaria, line 392: nit: could move this to a separate file `Downloaders` to keep this file a bit smaller.

fzakaria (commented):
We tested this locally using:
```
bazel run //tests/integration/pom_file:pom-example-with-runtime-dep.publish --define "maven_repo=file:///tmp/.m2"
```
(Note: we had to have set the `publish_maven_metadata` attr for the target)

Internally at Confluent, we tested the publishing via HTTP to AWS Code Artifactory as well.

shs96c:
The test failures look genuine.

I think the upload of the metadata should be disabled by default. Both Nexus and Artifactory can automatically calculate the maven metadata for you, and it's not always clear that every `publish` should result in the metadata being updated (eg. if you push an release candidate of a given library you may not want all users to update from the stable release)

vinnybod:
This also adds support for publishing to S3 which requires the client to update the maven-metadata.xml 🙂 

@shs96c I addressed the broken test and it is disabled by default.

private/tools/java/com/github/bazelbuild/rules_jvm_external/maven/MavenPublisher.java
```diff
@@ -261,6 +370,102 @@ private static String toHexS(String fmt, String algorithm, byte[] toHash) {
...
+          }
+        });
+  }
+
+  private static CompletableFuture<Optional<String>> httpDownload(
+      String targetUrl, Credentials credentials) {
+    return CompletableFuture.supplyAsync(
+        () -> {
+          try {
+            HttpTransport httpTransport = new NetHttpTransport();
```
shs96c, line 439: All versions of the JDK that we support have the native HTTP handler. We already use this in the `HttpDownloader` class, which also handles things like using the `.netrc` file if there is one. Perhaps use that? We could do that in a follow-up PR, but what I don't want to happen is for us to have a proliferation of http clients in this repo.

private/tools/java/com/github/bazelbuild/rules_jvm_external/maven/MavenPublisherTest.java
```diff
@@ -0,0 +1,74 @@
+package com.github.bazelbuild.rules_jvm_external.maven;
+
+import static org.junit.Assert.*;
```
shs96c, line 3: Looks like you need to run `bazel run scripts:format`

shs96c (approved):
LGTM, though I'm worried about the new http client we've pulled in.

shs96c:
Could you please run `bazel run scripts:format`, or otherwise expand the wildcard imports?

vinnybod:
I removed the extra http client dependency. Tests are passing, but I need to do a bit more end to end testing since I just merged ~4 months of commits from master.

private/tools/java/com/github/bazelbuild/rules_jvm_external/maven/MavenPublisher.java
```diff
@@ -305,6 +378,83 @@ private static String toHexS(String fmt, String algorithm, byte[] toHash) {
     }
   }
 
+  /**
+   * Attempts to download the file at the given targetUrl. Valid protocols are: http(s), file, and
+   * s3 at the moment.
+   */
+  private static CompletableFuture<Optional<String>> download(String targetUrl) {
```
shs96c, line 385: The schemes here aren't the same as `SUPPORTED_SCHEMES`. Is that deliberate?

private/tools/java/com/github/bazelbuild/rules_jvm_external/maven/MavenPublisher.java
```diff
@@ -305,6 +378,83 @@ private static String toHexS(String fmt, String algorithm, byte[] toHash) {
...
+    if (targetUrl.startsWith("http")) {
```
shs96c, line 386: Why not do a `URI.create(targetUrl)` and do a switch on the scheme? You don't want to use `URL` because that does a DNS lookup in its constructor, but `URI` should be safe.

private/tools/java/com/github/bazelbuild/rules_jvm_external/maven/MavenPublisherTest.java
```diff
@@ -0,0 +1,74 @@
...
+            + "    <versioning>\n"
+            + "        <latest>1.0</latest>\n"
+            + "        <release>1.0</release>\n"
+            + "        <versions>\n"
+            + "            <version>1.0</version>\n"
+            + "        </versions>\n"
+            + "        <lastUpdated>20200731090423</lastUpdated>\n"
+            + "    </versioning>\n"
+            + "</metadata>";
+    Metadata metadata =
```
shs96c, line 31: I'm curious. This looks like it's checking that the maven library works, but the test names suggests that this is about `MavenPublisher`.

private/tools/java/com/github/bazelbuild/rules_jvm_external/maven/MavenPublisherTest.java
```diff
@@ -0,0 +1,74 @@
...
+    assertEquals(metadata.getArtifactId(), "my-app");
+
+    Versioning versioning = metadata.getVersioning();
+    assertEquals(versioning.getLatest(), "1.0");
+    assertEquals(versioning.getLastUpdated(), "20200731090423");
+  }
+
+  @Test
+  public void testSerialization() throws Exception {
+    String expected =
```
shs96c, line 45: This is fragile. If you do want to test that the third party library is working, it'd be better to make assertions about the parsed XML that's returned, rather than the exact string.

private/tools/java/com/github/bazelbuild/rules_jvm_external/maven/MavenPublisher.java
vinnybod: It's a subset. Open to having someone else add the rest as needed.

private/tools/java/com/github/bazelbuild/rules_jvm_external/maven/MavenPublisherTest.java
vinnybod: We could just remove these tests, they were used early on in development just to make sure the library worked how we expected.

shs96c (approved):
LGTM. Thank you!

PR #1265, opened 2024-10-09
Do not use or recommend `@//` labels with Bzlmod

fmeum:
With Bzlmod, only the root module can refer to its own repo via `@//`. Instead, modules should just use repo-relative labels to refer to themselves, which works even if they aren't the root module.

Avoids this warning with Bazel 8:
```
WARNING: The module extension @@rules_jvm_external+//:extensions.bzl%maven produced an invalid lockfile entry because it referenced @@[unknown repo '' requested from @@rules_jvm_external+]. Please report this issue to its maintainers.
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -177,7 +177,7 @@ If you are using `bzlmod`, define an `install` tag in your root
 maven.install(
     name = "rules_jvm_external_deps",
     repositories = ["https://mycorp.com/artifacts"],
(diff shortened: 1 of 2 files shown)


jin (approved):

mortenmj:
This fixed https://github.com/bazel-contrib/rules_jvm_external/pull/1265

PR #1266, opened 2024-10-11
fix: Deduplicate deps while amending

honnix:
To address https://github.com/bazel-contrib/rules_jvm_external/pull/1251#issuecomment-2407255281


honnix:
Does this change make sense? Can someone help review it? Thanks.

jin:
(responding in this PR because this came first, but #1269 also appears to solve the same thing)

cc @shs96c who wrote the parent PR

Can you add comments like in #1269, and regression build tests by adding the triggering artifacts in https://github.com/bazel-contrib/rules_jvm_external/issues/1267 to https://github.com/bazel-contrib/rules_jvm_external/blob/1202665f8e233f43325396af209f4eecbc1587f6/MODULE.bazel#L501 and https://github.com/bazel-contrib/rules_jvm_external/blob/1202665f8e233f43325396af209f4eecbc1587f6/tests/unit/build_tests/BUILD#L7?

private/rules/coursier.bzl
```diff
@@ -963,6 +965,12 @@ def rewrite_files_attribute_if_necessary(repository_ctx, dep_tree):
 
     return dep_tree
 
+def is_dep(jar_path, deps):
+    for dep in reversed(deps):
```
jin, line 969: why reversed?

private/rules/coursier.bzl
```diff
@@ -963,6 +965,12 @@ def rewrite_files_attribute_if_necessary(repository_ctx, dep_tree):
 
     return dep_tree
 
+def is_dep(jar_path, deps):
```
jin, line 968: I think you can inline this function.

private/rules/coursier.bzl
honnix: Good catch. It's a leftover from an optimization I tried. I will remove it.

private/rules/coursier.bzl
honnix: Hmm, doesn't seem to be easier to read. After some tweak, this PR would be more or less like #1269 anyway, so maybe we can take it from there?

honnix:
I added a few more commits, but please feel free to take #1269 instead and cherry pick useful commits from this PR.

honnix:
Shall we move forward with this? I'm fine either this PR or #1269. Thanks.

jin (approved):

jin:
Thanks, sorry for the delay.

PR #1268, opened 2024-10-17
Remove android_extensions.bzl load

keith:
This has been removed from bazel 8.x, and is included in rules_android
now

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -6,7 +6,7 @@ module(
 
 bazel_dep(
     name = "rules_android",
-    version = "0.1.1",
+    version = "0.5.1",
 )
 bazel_dep(
     name = "bazel_features",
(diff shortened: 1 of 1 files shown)


MODULE.bazel
```diff
@@ -40,10 +40,6 @@ bazel_dep(
     repo_name = "io_bazel_stardoc",
 )
 
-# Remove this once rules_android has rolled out official Bzlmod support
-remote_android_extensions = use_extension("@bazel_tools//tools/android:android_extensions.bzl", "remote_android_tools_extensions")
-use_repo(remote_android_extensions, "android_gmaven_r8", "android_tools")
```
keith, line 45: not sure yet if we need to `use_repo` these still

shs96c:
Those failures look legitimate. 

We support "N-2" bazel LTS releases, which once Bazel 8 ships will mean that we want to continue supporting Bazel 6 workspace-based builds. I'm assuming that this change will continue to work with Bazel 6 once we land this?

ahumesky:
The failures in CI are because the Starlark rules in rules_android 0.5.1 rely on the natively-defined Android providers, which are available only with `--experimental_google_legacy_api`. The upcoming version 0.6.0 of rules_android uses Starlark-defined providers, except for ProguardSpecInfo (because it's in the java rules) which is unguarded from  `--experimental_google_legacy_api` in Bazel 7.4.0 and Bazel 8

There's some related discussion around compatibility in 
https://github.com/bazel-contrib/rules_jvm_external/issues/1270#issuecomment-2438271500

PR #1269, opened 2024-10-19
Ensure no duplicates after pom jar unification

### LINKED ISSUE #1267: Depending on both POM and JAR of the same maven artifact causes symlink error at artifacts pinning
# Description

After updating to `6.4`, when trying to pin artifacts, bazel reported the following error:

```
ERROR: An error occurred during the fetch of repository 'rules_jvm_external~~maven~unpinned_maven':
	File "external/rules_jvm_external~/private/rules/coursier.bzl", line 1077, column 84, in _coursier_fetch_impl
		artifact.update({"file": _relativize_and_symlink_file_in_coursier_cache(repository_ctx, artifact["file"], coursier_cache_path)})
	File "external/rules_jvm_external~/private/rules/coursier.bzl", line 231, column 31, in _relativize_and_symlink_file_in_coursier_cache
		repository_ctx.symlink(absolute_path, repository_ctx.path(artifact_relative_path))
Error in symlink: java.io.IOException: Could not create symlink <skipped> (File exists)
```

After investigation, I found that 2 of my dependencies were transitively depending on the same maven artifact, one with explicit `jar` packaging and one without (default to `pom`), specifically:
- `org.jetbrains.kotlinx:kotlinx-coroutines-core:1.7.3`
- `org.jetbrains.kotlinx:kotlinx-coroutines-core:jar:1.7.3` (note the `:jar`)

And the introduction of 8388befc8504ac9df779fad8940ec4a112e16fe0, causes a duplicate entry in the `dep_tree` since both entries for the `pom` and the `jar` coordinates of the same artifact will both have their file point the the same `.jar`.

https://github.com/bazel-contrib/rules_jvm_external/blob/8388befc8504ac9df779fad8940ec4a112e16fe0/private/rules/coursier.bzl#L955-L958

cerisier:
The same artifact can being depended on via pom and jar at different places in the tree. In such case, we deduplicate it so that 2 entries do not reference the same file, which will otherwise lead in symlink error because of existing file down the road.

Fixes #1267

### DIFF
--- a/private/rules/coursier.bzl
+++ b/private/rules/coursier.bzl
@@ -940,6 +940,7 @@ def rewrite_files_attribute_if_necessary(repository_ctx, dep_tree):
     # `file` attributes if necessary.
     # https://github.com/bazelbuild/rules_jvm_external/issues/1250
     amended_deps = []
+    deps_files = []
     for dep in dep_tree["dependencies"]:
         if not dep.get("file", None):
             amended_deps.append(dep)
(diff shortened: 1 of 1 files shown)


honnix:
FWIW, #1266 also addresses the same issue.

jin:
Thanks for this PR! I've merged https://github.com/bazel-contrib/rules_jvm_external/pull/1266 as that PR came first.

PR #1271, opened 2024-10-24
Upgrade rules_java and add rule loads

ted-xie:
* Upgrade to the latest rules_java 8.1.0
* Add loads for Java rules

NOTE: The loads and rules_java version unforutunately are tightly related. The paths for the bzl files for Java rules changed between rules_java 7 and 8.

### DIFF
--- a/private/rules/coursier.bzl
+++ b/private/rules/coursier.bzl
@@ -38,6 +38,7 @@ _BUILD = """
 
 load("@bazel_skylib//:bzl_library.bzl", "bzl_library")
 load("@rules_license//rules:package_info.bzl", "package_info")
+load("@rules_java//java:defs.bzl", "java_binary", "java_library", "java_plugin")
 load("@rules_jvm_external//private/rules:pin_dependencies.bzl", "pin_dependencies")
 load("@rules_jvm_external//private/rules:jvm_import.bzl", "jvm_import")
 {aar_import_statement}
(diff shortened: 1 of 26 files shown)


ted-xie:
*For some more context on the rule loads:* This PR is supposed to make rules_jvm_external work with Bazel 8, which removes all native Java rule loads.

Most users will be able to upgrade to Bazel 8 without manually adding all of these loads for Java rules. The exception is rule repository maintainers -- I'm not sure of the specifics, but it looks like rule reposotiries (rules_java, rules_android, etc) are special-cased, and autoloads from other rule repositories are disabled.

I was motivated to write this PR when trying to get rules_android to work with Bazel 8. Because I'm working inside a rule repo, it looks like transitive repo deps are not autoloaded -- for example, rules_jvm_external's use of java_library/java_import rules.

*For more context on why rules_java 8.1.0:* rules_java v8 changed where symbols such as java_binary and JavaInfo are loaded from. The old paths to various .bzl files seem to no longer exist, or at least are no longer visible.

jin (approved):

jin:
Wait, did you forget to update MODULE.bazel?

ted-xie:
Ah, my main repository already had rules_java bumped to 7.12.2 so I forgot I had to update it in RJE too. I see you already updated RJE's MODULE.bazel in 3ea96498725cb21e3ac18490cc3f3b75db2259ca. I think that should be sufficient.

PR #1273, opened 2024-10-28
Bump rules_java 7.12.2 and .bazelversion 7.4.0

jin:
..to unblock Bazel 8 and the rules_android decoupling from it.


jin:
@shs96c @cheister we'll need to cut a patch containing this and a new rules_android version to unblock Bazel 8 + rules_android's decoupling from Bazel. 

@ahumesky will follow up with the new rules_android version to be included. It will also maintain compatibility back to Bazel 6 by dropping the requirement on `--experimental_google_legacy_apis`.

Wyverald (approved):

meteorcloudy (approved):

PR #1274, opened 2024-10-30
Prepare for v6.5

jin:
Also update rules_java to 7.12.2. It was accidentally left out in https://github.com/bazel-contrib/rules_jvm_external/pull/1273. 

With the rules_java bump for Bazel 6, it brings in additional dependencies on rules_cc and protobuf. cc @comius 

cc @ted-xie @ahumesky


shs96c (approved):
LGTM

PR #1278, opened 2024-11-05
Allow root module's override tags to take precedence over the overridees from transitive deps.

jin:
Occasionally, a `bazel_dep` can introduce a `maven.overrides` tag that aliases an artifact to another label. This may not be what the root module desires, and so we should allow the root module's overrides to take precedence over any transitive overrides, but at the caution of the root module owner.

Concretely, this is a problem encountered by the Bazel project itself, where a `grpc-java` transitive dep overrides `@maven//:com_google_protobuf_protobuf_java` as an alias to `@@protobuf+//:protobuf_java`, which was not desirable.

This PR is validated by a test that contains a root override from okhttp to javapoet, which takes precedence over a dependency's override from okhttp to a poison pill label.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -732,6 +732,25 @@ dev_maven.install(
     ],
 )
 
+
+dev_maven.install(
+    name = "root_module_can_override",
+    artifacts = ["com.squareup:javapoet:1.11.1"],
+)
+
+bazel_dep(name = "transitive_module_can_override", version = "0.0.0")
+local_path_override(
+    module_name = "transitive_module_can_override",
+    path = "tests/integration/override_targets/module",
+)
+
+dev_maven.override(
+    # This override demonstrates that this root module's override takes precedence over that transitive override definition.
+    # Use something absurd for testing, like overriding okhttp3 to javapoet.
+    coordinates = "com.squareup.okhttp3:okhttp",
+    target = "@root_module_can_override//:com_squareup_javapoet",
+)
+
 # Where there are file locks, the pinned and unpinned repos are listed
 # next to each other. Where compat repositories are created, they are
 # listed next to the repo that created them. The list is otherwise kept
@@ -815,6 +834,7 @@ use_repo(
     "starlark_aar_import_test",
     "starlark_aar_import_with_sources_test",
     "strict_visibility_testing",
+    "root_module_can_override",
 
     # Repo with compat repos
     "com_google_http_client_google_http_client_gson",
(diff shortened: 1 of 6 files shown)


meteorcloudy:
Thanks! Is it possible for the root module to just remove the override instead of adding a new one? Not sure how to do it properly.

/cc @shs96c

jin:
It's possible, but removing it instead of aliasing it to something else might cause the transitive bazel_dep to break?

shs96c (approved):
Makes sense

meteorcloudy:
> It's possible, but removing it instead of aliasing it to something else might cause the transitive bazel_dep to break?

So I guess, if the root module wants to revert it back to the jar from maven, they can still do that with an override, right?

jin:
> > It's possible, but removing it instead of aliasing it to something else might cause the transitive bazel_dep to break?
> 
> 
> 
> So I guess, if the root module wants to revert it back to the jar from maven, they can still do that with an override, right?

Yep! That's the idea. The root owner can override it back to the original (or any other target) if they are sure that the transitive dep is ok with it. But outright removing it will probably break the transitive dep.

meteorcloudy (approved):
Thanks!

MODULE.bazel
```diff
@@ -732,6 +732,25 @@ dev_maven.install(
     ],
 )
 
+
+dev_maven.install(
+    name = "root_module_can_override",
+    artifacts = ["com.squareup:javapoet:1.11.1"],
+)
+
+bazel_dep(name = "transitive_module_can_override", version = "0.0.0")
```
shs96c, line 741: This should be marked as a `dev_dependency` so that consumers of us don't try and look up `transitive_module_can_override` in the BCR.

shs96c (changes requested):
The `bzlmod` example test is failing because the new dep is missing in the BCR.

sumeetgajjar:
Hi @jin - Thanks for raising this fix.
I realized the hard way when trying to use the pre-built binaries for gRPC proto generation that this was an issue.

Qq - Are you planning to push the fix for the failing test?
If you are busy, I offer to push the addition test, fix the commit to this branch to unblock, and get this change released :)

jin:
@sumeetgajjar Yes, my bandwidth is limited right now, please feel free to fork this PR and get it submitted.

sumeetgajjar:
> @sumeetgajjar Yes, my bandwidth is limited right now, please feel free to fork this PR and get it submitted.

Sounds good, will push it before https://github.com/bazel-contrib/rules_jvm_external/issues/1379

PR #1284, opened 2024-11-22
Allow round-tripping of the v2 lock file format in java

shs96c:
There have been several occasions where I need to get the information from the lock file in a reusable way. Having this available from within `rules_jvm_external`'s own classes would be most useful


tests/com/github/bazelbuild/rules_jvm_external/resolver/lockfile/V2LockFileTest.java
```diff
@@ -0,0 +1,151 @@
...
+
+  @Test
+  public void shouldRoundTripM2Local() {
+    V2LockFile lockFile = new V2LockFile(repos, Set.of(), Set.of());
+    Map<String, Object> rendered = lockFile.render();
+    rendered.put("m2local", true);
+
+    V2LockFile roundTripped =
+        V2LockFile.create(
+            new GsonBuilder().setPrettyPrinting().serializeNulls().create().toJson(rendered));
```
jin, line 78: Use the `roundTrip` helper?

private/tools/java/com/github/bazelbuild/rules_jvm_external/resolver/lockfile/V2LockFile.java
```diff
@@ -12,32 +12,175 @@
...
+    return infos;
+  }
+
+  public Set<Conflict> getConflicts() {
+    return conflicts;
+  }
+
+  @SuppressWarnings("unchecked")
+  public static V2LockFile create(String from) {
+    URI m2LocalUri = Paths.get(USER_HOME.value()).resolve(".m2/repository").toUri();
```
jin, line 66: nit: make this static and share it with `render`

private/tools/java/com/github/bazelbuild/rules_jvm_external/resolver/DependencyInfo.java
```diff
@@ -81,6 +82,18 @@ public SortedMap<String, SortedSet<String>> getServices() {
     return services;
   }
 
+  @Override
+  public String toString() {
```
jin, line 86: nit: use `MoreObjects#toStringHelper` which handles the formatting for you.

jin (approved):

private/tools/java/com/github/bazelbuild/rules_jvm_external/resolver/DependencyInfo.java
shs96c: Done.

private/tools/java/com/github/bazelbuild/rules_jvm_external/resolver/lockfile/V2LockFile.java
shs96c: Done

tests/com/github/bazelbuild/rules_jvm_external/resolver/lockfile/V2LockFileTest.java
shs96c: We amended the generated JSON by hand by adding `m2local` to it, so we can't use the `roundTrip` helper here. We _could_ have added the `M2_LOCAL_URI` to the list of repos when we create the lockfile instance, but I don't wanted to be sure we were honouring the `m2local` value in the rendered map properly.

PR #1291, opened 2024-11-26
Fix docs generation for artifacts with compile only dependencies

vinnybod:
Fixes https://github.com/bazel-contrib/rules_jvm_external/issues/1241
A detailed description of the issue, minimal repo, and explanation of the fix can be found at https://github.com/vinnybod/bazel_java_example/tree/javadocs-srcs-issue?tab=readme-ov-file#solution


shs96c (approved):
LGTM

PR #1293, opened 2024-12-05
Allow package exclusions and inclusions in javadocs

vinnybod:
There are times where developers want to exclude specific packages from being generated in the javadocs such as internal code. This is commonly offered as a feature of javadoc generation in both maven and gradle.

An exact package can be provided to remove the first level of files, or a `.*` to remove all subpackages under it.

```
java_export(
  ...
  doc_excluded_packages = [
    "com.example.internal.*"
  ]
)
```

```
java_export(
  ...
  doc_included_packages = [
    "com.example.processor.*"
  ],
  doc_excluded_packages = [
    "com.example.processor.internal.*"
  ]
)
```


private/rules/kt_jvm_export.bzl
```diff
@@ -70,11 +70,8 @@ def kt_jvm_export(
 
     javadocopts = kwargs.pop("javadocopts", None)
 
-    # ensure that the kotlin-stdlib is included in deploy_env
-    if KOTLIN_STDLIB not in deploy_env:
-        updated_deploy_env = deploy_env + [KOTLIN_STDLIB]
-    else:
-        updated_deploy_env = deploy_env
+    updated_excluded_workspaces = {name: None for name in excluded_workspaces}
```
vinnybod, line 73: `deploy_env` gets added back into the docs so we were generating docs for kotlin stdlib which seems wrong. Changed to excluded_workspace to keep it out of the docs, which seems like what was actually intended.
This is also how we handle `scala_export` in confluent's fork of rules_jvm_external.

vinnybod:
Hm, only windows is failing. not sure why

PR #1299, opened 2024-12-20
[auth][netrc] Empty account fallback to login

stevebarrau:
With this change the following netrc would get the username set, it used to set an **empty username**:

```
machine artifacts.example.org
  login bob
  password superSecretPa$$word
```


shs96c (approved):
LGTM. Thank you for the patch!

jin (approved):

PR #1302, opened 2024-12-22
Use rules_android_ndk

utzcoz:
When I try to execute `REPIN=1 bazel run @maven//:pin`, the `android_ndk_repository` tells me that it can't recognize `RELEASE.TXT` file or `platforms` directory when exposing latest Android Ndk with `ANDROID_NDK_HOME`.

After investigation, if we want to use Android NDK rc22+, we need to use `rules_android_ndk`, and this module has been released with version 1.0.2, also included into bazel central registry. So this CL just switchs to use `rules_android_ndk` latest stable version, and adds necessary rules_cc that `rules_android_ndk` requires. With it, I can run repin command to generate extra changes for `maven_install.json`.

I am not familiar with bzlmod and rules_jvm_external requirements for external dependencies very much , so please tell me if I did something wrong.

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -6,6 +6,7 @@ tasks:
       # https://github.com/bazelbuild/rules_jvm_external/pull/316
       COURSIER_CACHE: /tmp/custom_coursier_cache
       REPIN: 1
+      ANDROID_NDK_HOME: /opt/android-ndk-r25b
     shell_commands:
       - bazel run @regression_testing_coursier//:pin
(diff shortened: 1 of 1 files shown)


jin:
I'm pretty sure we don't use the NDK for developing the project or even for downstream users who download aar artifacts -- or at least, we _shouldn't_ incur this dependency. It's an incidental complexity, and I think the right fix is to remove the dep altogether.

https://github.com/bazel-contrib/rules_jvm_external/pull/1305 is an attempt.

Issue #916, opened 2023-06-11
Duplicate maven repositories when importing bazel_deps that use maven.install


aarontsharp:
When using `maven.install` via bzlmod in the root module, any `bazel_dep` which also uses `maven.install` (with the default repository name) can create duplicate artifacts/conflicts.

For instance, when including a bazel_dep for [protobuf 21.7](https://registry.bazel.build/modules/protobuf) , the following message is logged:

`The maven repository 'maven' is used in two different bazel modules, originally in '' and now in 'protobuf'`

This happens because protobuf does a `maven.install` using `maven` as the repository name [pinning various common dependencies](https://github.com/bazelbuild/bazel-central-registry/blob/6c6cdc13950a0d6dc9f5a7e7ead07d4cfef0db4a/modules/protobuf/21.7/MODULE.bazel#L24)

This is maybe benign in and of itself, but it can create artifact conflicts. For example, pinning different versions of `mockito`/`guava`/etc to the root module, the following messages are logged:
```
Found duplicate artifact versions
    com.google.guava:guava has multiple versions 32.0.0-jre, 31.1-jre
    org.mockito:mockito-core has multiple versions 5.3.1, 4.3.1
```

It's possible to work around this by using a separate `name` for `maven.install`, but it is quite annoying (since you need to set `repository_name` everywhere) and likewise incredibly difficult to understand where these duplicates were introduced. In my case, I wasn't even directly importing `protobuf` in my root module, but transitively through another bazel_dep, so this was incredibly confusing.

Is there a better way to understand and resolve these conflicts?

meteorcloudy:
@shs96c Are `maven.install` with the same `name` resolved together or separately? How do you recommend to address the warning message?

/cc @Wyverald

Ubehebe:
I just ran into this too, and I think it's specifically a problem with the protobuf module. Other modules take care to instantiate their maven artifacts with different names to avoid polluting the default "maven" name:
- rules_jvm_external itself [uses the name](https://github.com/bazelbuild/rules_jvm_external/blob/master/MODULE.bazel#L24-L25) rules_jvm_external_deps
- rules_kotlin [uses the name](https://github.com/bazelbuild/rules_kotlin/blob/master/MODULE.bazel#L37-L38) kotlin_rules_maven.

Is there a good reason for protobuf should make its internal java dependencies visible in this way? If not, we should ask the protobuf module maintainers to change it.

meteorcloudy:
[...]

Issue #708, opened 2022-06-13
Documentation for resolving `Found duplicate artifact versions`


aryeh-looker:
Would be great to document best-practices for resolving duplicate artifact versions

https://github.com/bazelbuild/rules_jvm_external/blob/master/coursier.bzl#L649

desh-woes:
+1 I am currently not sure how to go about resolving duplicate artifacts.

shs96c:
Duplicate artefacts are listed when the same artefact is in the list given to `maven_install`. Quite often, this is because you're pulling in a constant defined in another repo and including that in your `maven_install`'s `artifacts` parameter.

AAverin:
Does it work with having multiple android version installed? How can I resolve it?
```
rules_jvm_external/coursier.bzl:646:18: Found duplicate artifact versions
    com.google.guava:guava has multiple versions 31.1-jre, 29.0-android, 31.0.1-android
```

vorburger:
> Quite often, this is because you're pulling in a constant defined in another repo and including that in your maven_install's artifacts parameter.

A typical case of this is:

    com.google.guava:guava has multiple versions 32.1.3-android, 33.0.0-jre

if you have (something) like:

```starlark
maven_install(
    artifacts = IO_GRPC_GRPC_JAVA_ARTIFACTS + [
        "junit:junit:4.13.2",
        "com.google.guava:guava:33.0.0-jre",
```

Because [`IO_GRPC_GRPC_JAVA_ARTIFACTS` has `com.google.guava:guava:32.1.3-android`](https://github.com/grpc/grpc-java/blob/38f968fafb46371fe43a08ae6e320a552bc3acf2/repositories.bzl#L23C5-L23C44).

Here is one way to fix this, [as I have done here](https://github.com/enola-dev/enola/commit/49bd4a0d3c51a03bc9e53d8aabb598bf9718e112):

```starlark
# This is required so that we can use duplicate_version_warning (below) and fixes
# the "com.google.guava:guava has multiple versions 32.1.3-android, 33.0.0-jre" problem.
IO_GRPC_GRPC_JAVA_ARTIFACTS_WITHOUT_GUAVA = [item for item in IO_GRPC_GRPC_JAVA_ARTIFACTS if item != "com.google.guava:guava:32.1.3-android"]

maven_install(
    artifacts = IO_GRPC_GRPC_JAVA_ARTIFACTS_WITHOUT_GUAVA + [
[...]

PR #1303, opened 2024-12-30
Document how to deal with 'Found duplicate artifact versions'

vorburger:
Re. #708 & #916

This [did the trick](https://github.com/enola-dev/enola/commit/ba3e78717fd97a9f96b39a6b208ce625281a0950), at least for me.

### DIFF
--- a/docs/bzlmod.md
+++ b/docs/bzlmod.md
@@ -13,6 +13,7 @@ See the `/examples/bzlmod` folder in this repository for a complete, tested exam
 First, you must enable bzlmod.
 Note, the Bazel team plans to enable it by default starting in version 7.0.
 The simplest way is by adding this line to your `.bazelrc`:
+
 ```
 common --enable_bzlmod
 ```
@@ -22,8 +23,14 @@ setting the `version` to the latest one available on https://registry.bazel.buil
 
 ```starlark
 bazel_dep(name = "rules_jvm_external", version = "...")
-maven = use_extension("@rules_jvm_external//:extensions.bzl", "maven")
-maven.install(
+your_maven = use_extension("@rules_jvm_external//:extensions.bzl", "maven")
+your_maven.install(
+    # NB: Do *ALWAYS* specify a name, and NEVER use the (default) name = "maven";
+    # this is because it's not "namespaced" (or "scoped") like you might think it
+    # really should be... use the default name *WILL* cause conflicts with other modules;
+    # notably grpc-java! See e.g. https://github.com/bazel-contrib/rules_jvm_external/issues/916
+    # and https://github.com/bazel-contrib/rules_jvm_external/issues/708 for more background.
+    name = "YOUR_maven",
     artifacts = [
         # This line is an example coordinate, you'd copy-paste your actual dependencies here
         # from your build.gradle or pom.xml file.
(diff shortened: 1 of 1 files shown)


shs96c:
Here's the mental model that I use.

Projects that are expected to be non-root modules (eg. rulesets and libraries) should define a `name` that's unique to them. Projects that are expected to always be root modules should use the default name, since that's the most convenient.

The key benefit of this is that non-root modules that know that they absolutely must contribute to the default resolution can add things to the default namespace, and it'll work nicely. At the same time, most users won't need to think about which namespace their dependencies are coming from (since we always default to `maven`)

As a concrete example, `contrib_rules_jvm`, which require some dependencies in order to work should define its own namespace because it's expected to be a non-root module. I'd suggest that both `protobuf` and `grpc-java` fall into this category as well, since there are well-known maven coordinates that people can import into their dependencies if they need to access that tooling in their own code.

OTOH, in the Selenium project (which we always expect to be a root module), we use the default namespace.

So, for the majority of users, I suspect that the default namespace will be just fine, but we should emphasise that projects which are expected to be non-root modules shouldn't be using `maven`.

vorburger:
From what (little, I admit) I understand, as-is currently not all _"non-root modules (eg. rulesets and libraries"_ comply with what's outlined above (notably gRPC & Proto?), yet - correct? Ergo, the documentation as-is currently (without something like this) isn't as helpful as it could be - agreed?

What I could do is amend this PR by adding something based on the comment above - would that be helpful?

docs/bzlmod.md
```diff
@@ -22,8 +23,14 @@ setting the `version` to the latest one available on https://registry.bazel.buil
 
 ```starlark
 bazel_dep(name = "rules_jvm_external", version = "...")
-maven = use_extension("@rules_jvm_external//:extensions.bzl", "maven")
-maven.install(
+your_maven = use_extension("@rules_jvm_external//:extensions.bzl", "maven")
+your_maven.install(
+    # NB: Do *ALWAYS* specify a name, and NEVER use the (default) name = "maven";
```
ejona86, line 28: Please don't tell people to workaround things and then cause me extra grief as users get duplicate (and not identical) classes in their class path.

This all stems from java_proto_library. Its entire existence it has brought in java_library()s from `@com_google_protobuf`. That means you need overrides to prevent Maven Central dependencies bringing in a _different_ copy of protobuf and put it also on your classpath. Today you can't use the protobuf Maven Central binaries.

Reusing the "maven" name is on purpose and is the only functional way today. Also, since protobuf depends on Maven Central binaries itself, you need a single resolution with the application. See https://stackoverflow.com/questions/54119248/how-to-upgrade-a-bazel-aware-librarys-maven-dependency where I tried to get this all working before maven_install existed. With both maven_install() and maven.install() we need a single resolution of versions to take place.

Note that that is all very separate from rules needing maven.install() for their own processing. That should indeed be a different name because those dependencies won't ever be part of the user's classpath.

ejona86: https://github.com/grpc/grpc-java/issues/12149 is an example of the pain this language would cause.

shs96c:
@vorburger, apologies for taking so long with this. Would you mind taking a look at #1388 and #1393 and seeing whether they address the points you wanted in this PR?

PR #1305, opened 2025-01-02
Remove dep on android ndk.

jin:
We're not using any NDK utility in RJE, and we shouldn't be in the business of doing so. The only android dep is aar_import, so it should be clear to drop our dependency on the NDK.

Also remove the aar_import-based test for `override_targets`: we already have sufficient coverage using the opentelemetry/java_library test case in the same `tests/integration/override_targets/BUILD` file. 

The SDK is still needed for `//tests/unit/aar_import` targets.


jin:
cc @ahumesky

shs96c (approved):
LGTM, though I'm a little surprised that this means we remove the test in `override_targets`

PR #1306, opened 2025-01-04
Add support for using Maven BOMs with Coursier

protocol7:
Coursier now support using Maven BOMs when resolving. This upgrades Coursier to 2.1.24 and enables using BOMs leveraging the existing `boms` attribute on `maven_install`.

Note that Coursier has a slight performance regression since adding support for BOMs, see https://github.com/coursier/coursier/issues/3190.

Fixes https://github.com/bazel-contrib/rules_jvm_external/issues/1289


protocol7:
Had to disable dependencies from https://packages.confluent.io as the repository seems to be broken at the moment.

Not sure about the remaining CI failure, as it looks to be unrelated to this change:
```
2025-01-04 15:34:59 UTC	WARNING: Couldn't auto load rules or symbols, because no dependency on module/repository 'rules_android' found. This will result in a failure if there's a reference to those rules or symbols.
2025-01-04 15:35:00 UTC	ERROR: Traceback (most recent call last):
2025-01-04 15:35:00 UTC		File "/var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/ec321eb2cc2d0f8f91b676b6d4c66c29/external/+maven+regression_testing_coursier/BUILD", line 48, column 11, in <toplevel>
2025-01-04 15:35:00 UTC			aar_import(
2025-01-04 15:35:00 UTC	Error in aar_import: Couldn't auto load 'aar_import' from '@rules_android//rules:rules.bzl'.
```

vinnybod:
@protocol7 What seems to be the issue with the packages.confluent.io repository? Was it intermittent?

protocol7:
> @protocol7 What seems to be the issue with the packages.confluent.io repository? Was it intermittent?

Thanks for the nudge, it seems to have been intermittent, so I now restored the test artifact from the repo.

protocol7:
Updated to Coursier 2.1.24 which fixes the last issue we've found in our internal use of the BOM support.

shs96c (approved):
LGTM. Really nice to see that BOMs have come for the `coursier` resolver too. Thank you for doing this!

peakschris:
This is causing a regression in our use of rules_jvm_external - we cannot upgrade to the latest release because coursier does not exist on mirror.bazel.build.

https://github.com/bazel-contrib/rules_jvm_external/issues/1329

PR #1308, opened 2025-01-14
Fix collection of Maven runtime deps

fmeum:
A transitive dependency of a `maven_project_jar` is a runtime dependency if and only if each path to it from the root goes through at least one `runtime_deps` edge. Before this commit, every transitive dependency that was reachable from the root via at least one path with a `runtime_deps` edge was considered to be a runtime dependency, even if other paths require it as a compile dependency. This fixes a regression introduced by https://github.com/bazel-contrib/rules_jvm_external/commit/7b0abdc591b9b0b0dbdfd62bf9d98928b0efe6a2.

Also show a helpful error when users attempt to move a dep from `runtime_deps` to `deps` on a `java_export` target with no `srcs`.

Work towards https://github.com/CodeIntelligenceTesting/jazzer/issues/919

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -52,7 +52,8 @@ tasks:
       - "//..."
   ubuntu2204_latest:
     platform: ubuntu2204
-    bazel: latest
+    # TODO: Revert to latest after fixing incompatibilities with Bazel 8.0.0
+    bazel: 7.x
     environment:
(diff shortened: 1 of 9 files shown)


fmeum:
CI failures look unrelated

shs96c (approved):
LGTM. Checked that the changes have exploded the size of generated POM files, and it looks good. Thank you, @fmeum!

PR #1311, opened 2025-01-16
Rework how we determine coordinate formats

shs96c:
The assumption that versions begin with numbers is fatally flawed in reality. Fortunately, where we're trying to disambiguate, most of the time, it's safe to assume that people mean a version number rather than some other format.

### DIFF
--- a/private/coursier_utilities.bzl
+++ b/private/coursier_utilities.bzl
@@ -16,24 +16,9 @@
 # download them if it's not asked to to resolve "eclipse-plugin".
 
 load("//:specs.bzl", "parse")
+load("//private/lib:coordinates.bzl", _SUPPORTED_PACKAGING_TYPES = "SUPPORTED_PACKAGING_TYPES")
 
-SUPPORTED_PACKAGING_TYPES = [
-    "jar",
-    "json",
-    "aar",
-    "bundle",
-    "eclipse-plugin",
-    "exe",
-    "orbit",
-    "test-jar",
-    "hk2-jar",
-    "maven-plugin",
-    "scala-jar",
-    "dylib",
-    "so",
-    "dll",
-    "pom",
-]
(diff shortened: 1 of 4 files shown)


cheister:
Can you add a regression tests that cover one of the artifacts used in the issue?
```
# https://github.com/bazelbuild/rules_jvm_external/issues/1290
"com.google.apis:google-api-services-calendar:v3-rev20241101-2.0.0",
```

tests/unit/coordinates_test.bzl
```diff
@@ -114,6 +115,17 @@ def _multiple_formats_impl(ctx):
 
 multiple_formats_test = unittest.make(_multiple_formats_impl)
 
+def _unusual_version_format_impl(ctx):
+    env = unittest.begin(ctx)
+
+    unpacked = unpack_coordinates("group:artifact:FY21R16")
```
cheister, line 121: Should we also add a test for a version that doesn't have any numbers?

private/coursier_utilities.bzl
```diff
@@ -46,23 +33,42 @@ PLATFORM_CLASSIFIER = [
...
+    # We want to modify some of the values
+    unpacked_struct = unpack_coordinates(coord)
+    unpacked = {} | struct_to_dict(unpacked_struct)
+
+    if unpacked.get("classifier", None) in ["sources", "native"]:
+        unpacked["classifier"] = None
+
+    # We add "pom" into SUPPORTED_PACKAGING_TYPES here because "pom" is not a
+    # packaging type that Coursier CLI accepts.
+    if unpacked.get("packaging", None) in SUPPORTED_PACKAGING_TYPES + ["pom"]:
```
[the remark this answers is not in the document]
shs96c: Removed. Nice catch. Thank you

tests/unit/coordinates_test.bzl
shs96c: Done

cheister (approved):

PR #1314, opened 2025-01-19
Update protobuf to 29.3 (bzlmod only)

joca-bt:


shs96c (approved):
LGTM. Thanks for the comment explaining why we're leaving the workspace-based build alone!

PR #1319, opened 2025-02-06
Give the coursier pin alias public visibility to match others

erikkerber:
Ran into this when using the alias'd `:pin` in a `rules_multirun` command. matches visibility of other pins.


shs96c (approved):
Good point, well made. LGTM

erikkerber:
This is correct though maybe not quite yet _useful_. In putting this into rules_multirun I realized also need to do a few less than aesthetic things such as exposing `:unsorted_deps.json` view filegroup.

I'll use that as a local patch, and if others come across this need or if I find a more elegant way to do it I'll make a PR for that too.

PR #1322, opened 2025-02-07
Fix coordinate parsing when version does not start with a digit

cheister:
Fixes https://github.com/bazel-contrib/rules_jvm_external/issues/1321 where coursier could not resolve an artifact with a version that did not start with a digit


shs96c:
Would you prefer this to #1311?

daniel-b2c2:
My two cents, this (#1322) breaks the "principle of least astonishment".  :( 

"Everything works as you'd expect unless the version doesn't have a digit in it" 😬

cheister:
Closing in favor of https://github.com/bazel-contrib/rules_jvm_external/pull/1311

PR #1323, opened 2025-02-07
Use Bazel's JDK to run coursier

cheister:
Addresses https://github.com/bazel-contrib/rules_jvm_external/issues/445 and  https://github.com/bazel-contrib/rules_jvm_external/issues/1046 by taking the workaround in https://github.com/bazel-contrib/rules_jvm_external/issues/445 and applying it as the default before looking at JAVA_HOME.

I also added a way to give the JVM settings for both coursier and maven if a custom cacerts file was needed.  This can be set by doing
```
build --repo_env=JDK_JAVA_OPTIONS=-Djavax.net.ssl.trustStore=<path-to-cacerts>
```
in `.bazelrc` 
Based on [JDK_JAVA_OPTIONS](https://docs.oracle.com/en/java/javase/11/tools/java.html#GUID-3B1CE181-CD30-4178-9602-230B800D4FAE__USINGTHEJDK_JAVA_OPTIONSLAUNCHERENV-F3C0E3BA)

### DIFF
--- a/private/rules/coursier.bzl
+++ b/private/rules/coursier.bzl
@@ -110,6 +110,7 @@ pin_dependencies(
     fetch_sources = {fetch_sources},
     fetch_javadocs = {fetch_javadocs},
     lock_file = {lock_file},
+    jvm_flags = {jvm_flags},
     visibility = ["//visibility:public"],
 )
 """
@@ -257,6 +258,14 @@ def _get_aar_import_statement_or_empty_str(repository_ctx):
         return ""
 
 def _java_path(repository_ctx):
+    # Allow setting an env var to keep legacy JAVA_HOME behavior
+    use_java_home = repository_ctx.os.environ.get("RJE_COURSIER_USE_JAVA_HOME")
+
+    if use_java_home == None:
+        embedded_java = "../bazel_tools/jdk/bin/java"
+        if _is_file(repository_ctx, embedded_java):
+            return repository_ctx.path(embedded_java)
+
     java_home = repository_ctx.os.environ.get("JAVA_HOME")
     if java_home != None:
         return repository_ctx.path(java_home + "/bin/java")
@@ -693,6 +702,7 @@ def generate_pin_target(repository_ctx, unpinned_pin_target):
             boms = repr(repository_ctx.attr.boms),
             artifacts = repr(repository_ctx.attr.artifacts),
             excluded_artifacts = repr(repository_ctx.attr.excluded_artifacts),
+            jvm_flags = repr(repository_ctx.os.environ.get("JDK_JAVA_OPTIONS")),
             repos = repr(repository_ctx.attr.repositories),
             fetch_sources = repr(repository_ctx.attr.fetch_sources),
             fetch_javadocs = repr(repository_ctx.attr.fetch_javadoc),
@@ -1511,6 +1521,7 @@ coursier_fetch = repository_rule(
     },
     environ = [
         "JAVA_HOME",
+        "JDK_JAVA_OPTIONS",
         "http_proxy",
         "HTTP_PROXY",
         "https_proxy",
--- a/private/rules/pin_dependencies.bzl
+++ b/private/rules/pin_dependencies.bzl
@@ -15,7 +15,7 @@ load("//private/rules:coursier.bzl", "compute_dependency_inputs_signature")
 
 _TEMPLATE = """#!/usr/bin/env bash
 
-{resolver_cmd} --argsfile {config} --resolver {resolver} --input_hash '{input_hash}' --output {output}
+{resolver_cmd} --jvm_flags={jvm_flags} --argsfile {config} --resolver {resolver} --input_hash '{input_hash}' --output {output}
 """
 
 def _stringify_exclusions(exclusions):
@@ -79,6 +79,7 @@ def _pin_dependencies_impl(ctx):
             resolver_cmd = ctx.executable._resolver.short_path,
             resolver = ctx.attr.resolver,
             output = "$BUILD_WORKSPACE_DIRECTORY/" + ctx.attr.lock_file,
+            jvm_flags = ctx.attr.jvm_flags,
         ),
         is_executable = True,
     )
@@ -119,6 +120,9 @@ pin_dependencies = rule(
         "lock_file": attr.string(
             doc = "Location of the generated lock file",
             mandatory = True,
+        ),
+        "jvm_flags": attr.string(
+            doc = "JVM flags to pass to resolver",
         ),
         "_resolver": attr.label(
             executable = True,
--- a/README.md
+++ b/README.md
@@ -47,6 +47,7 @@ Table of Contents
     - [Fetch and resolve timeout](#fetch-and-resolve-timeout)
     - [Ignoring empty jars](#ignoring-empty-jars)
     - [Duplicate artifact warning](#duplicate-artifact-warning)
+    - [Provide JVM options for artifact resolution](#provide-jvm-options-for-artifact-resolution)
     - [Provide JVM options for Coursier with `COURSIER_OPTS`](#provide-jvm-options-for-coursier-with-coursier_opts)
     - [Resolving issues with nonstandard system default JDKs](#resolving-issues-with-nonstandard-system-default-jdks)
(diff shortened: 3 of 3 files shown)


cheister:
If this looks like an acceptable approach I can update the documentation as well.

private/rules/coursier.bzl
```diff
@@ -257,11 +258,15 @@ def _get_aar_import_statement_or_empty_str(repository_ctx):
         return ""
 
 def _java_path(repository_ctx):
-    java_home = repository_ctx.os.environ.get("JAVA_HOME")
-    if java_home != None:
-        return repository_ctx.path(java_home + "/bin/java")
-    elif repository_ctx.which("java") != None:
-        return repository_ctx.which("java")
+    embedded_java = "../bazel_tools/jdk/bin/java"
```
shs96c, line 261: Does this work with Java 8? I thought the location of the java binary was somewhere under `rules_java` now?

private/rules/coursier.bzl
```diff
@@ -257,11 +258,15 @@ def _get_aar_import_statement_or_empty_str(repository_ctx):
...
+    if _is_file(repository_ctx, embedded_java):
+        return repository_ctx.path(embedded_java)
+    else:
```
shs96c, line 264: We've already returned at this point if `embedded_java` is a file, so the `else` is unnecessary and introduces extra indentation.

private/rules/coursier.bzl
```diff
@@ -257,11 +258,15 @@ def _get_aar_import_statement_or_empty_str(repository_ctx):
...
+        java_home = repository_ctx.os.environ.get("JAVA_HOME")
+        if java_home != None:
+            return repository_ctx.path(java_home + "/bin/java")
+        elif repository_ctx.which("java") != None:
```
shs96c, line 268: I see we did the same thing here, but I don't think it makes things clearer. *shrug*

private/rules/pin_dependencies.bzl
```diff
@@ -120,6 +121,9 @@ pin_dependencies = rule(
             doc = "Location of the generated lock file",
             mandatory = True,
         ),
+        "jvm_flags": attr.string(
```
shs96c, line 124: A `string_list` would better model what you're doing here, and would match how we set jvm args in the regular java rules.

shs96c (commented):
The general approach seems fine, but it might be fragile.

private/rules/coursier.bzl
cheister: It worked when I ran `REPIN=1 bazel run @regression_testing_coursier//:pin` with Bazel 8.0.1.  Adding RJE_VERBOSE=1 showed that the coursier tasks were running with
```
.../bazel_tools/jdk/bin/java -noverify -jar ...
```
I also deleted the bazel cache directory and ran it from a clean install and it also worked.

private/rules/pin_dependencies.bzl
cheister: True, but is there value in parsing the `JDK_JAVA_OPTIONS` env var into a string_list only to immediately re-concatenate it so we can pass it to the `--jvm_flags` option of the java wrapper script?

private/rules/coursier.bzl
cheister: ok, removed the else

cheister:
Since this would change the default java running coursier do we want to provide an option for people to opt back to using JAVA_HOME if they need to?

private/rules/coursier.bzl
srdo: Is this JDK the minimized (jlinked) JDK Bazel bundles, or is it something else (e.g. a downloaded separate one)?

I'm asking because I think the minimized JDK may not contain all normal JDK modules, and so using it to run Coursier could be problematic down the road if Bazel drops modules from their dependency list that Coursier happens to need.

https://github.com/bazelbuild/bazel/blob/d2c2ad1c657d517ae421a562028e26c345eaf12f/src/minimize_jdk.sh#L64

When I check the module list for this JDK in a local Bazel install, it looks like it's been stripped down.

```
$ jimage list bazel_tools/jdk/lib/modules | grep "Module:"
Module: java.base
Module: java.compiler
Module: java.instrument
Module: java.logging
Module: java.management
Module: java.naming
Module: java.security.sasl
Module: java.sql
Module: java.transaction.xa
Module: java.xml
Module: jdk.crypto.ec
Module: jdk.management
Module: jdk.unsupported
```
as compared to a full JDK:

```
$ jimage list modules | grep "Module:"
Module: java.base
Module: java.compiler
Module: java.datatransfer
Module: java.desktop
Module: java.instrument
Module: java.logging
Module: java.management.rmi
Module: java.management
Module: java.naming
Module: java.net.http
Module: java.prefs
Module: java.rmi
Module: java.scripting
Module: java.se
Module: java.security.jgss
Module: java.security.sasl
Module: java.smartcardio
Module: java.sql.rowset
Module: java.sql
Module: java.transaction.xa
Module: java.xml.crypto
Module: java.xml
Module: jdk.accessibility
Module: jdk.attach
Module: jdk.charsets
Module: jdk.compiler
Module: jdk.crypto.cryptoki
Module: jdk.crypto.ec
Module: jdk.dynalink
Module: jdk.editpad
Module: jdk.graal.compiler.management
Module: jdk.graal.compiler
Module: jdk.hotspot.agent
Module: jdk.httpserver
Module: jdk.incubator.vector
Module: jdk.internal.ed
Module: jdk.internal.jvmstat
Module: jdk.internal.le
Module: jdk.internal.opt
Module: jdk.internal.vm.ci
Module: jdk.jartool
Module: jdk.javadoc
Module: jdk.jcmd
Module: jdk.jconsole
Module: jdk.jdeps
Module: jdk.jdi
Module: jdk.jdwp.agent
Module: jdk.jfr
Module: jdk.jlink
Module: jdk.jpackage
Module: jdk.jshell
Module: jdk.jsobject
Module: jdk.jstatd
Module: jdk.localedata
Module: jdk.management.agent
Module: jdk.management.jfr
Module: jdk.management
Module: jdk.naming.dns
Module: jdk.naming.rmi
Module: jdk.net
Module: jdk.nio.mapmode
Module: jdk.random
Module: jdk.sctp
Module: jdk.security.auth
Module: jdk.security.jgss
Module: jdk.unsupported.desktop
Module: jdk.unsupported
Module: jdk.xml.dom
Module: jdk.zipfs
```

shs96c:
@cheister, that's probably a good idea. The neatest way would be a [configurable build attribute](https://bazel.build/docs/configurable-attributes#faq-select-platforms), but an env variable would be simpler.

private/rules/coursier.bzl
cheister: This is true, but CI should catch any problems with the minimized JDK when coursier is updated in the future and we'll leave the option to run with java from JAVA_HOME as an option for now.

srdo: :+1: The risk is probably still there that CI doesn't exercise all the coursier code so missing class errors could be lurking, but might not be an issue in practice. 

Sorry for butting in, but I'm wondering why rules_jvm_external can't find Java via a toolchain, like other Java-using rulesets do, instead of trying to use the embedded JDK?

cheister:
> @cheister, that's probably a good idea. The neatest way would be a [configurable build attribute](https://bazel.build/docs/configurable-attributes#faq-select-platforms), but an env variable would be simpler.

Yeah, going for the env var since I'm hoping we can remove it later

blockjon-dd:
Thank you so much for working on this PR.

I'm at the brink of installing the JDK on all of our Bazel CI and developer machines. However, it seems like once this PR merges, I could avoid doing that.

Therefore, I feel like I have to ask when this PR might merge. If the answer is really soon, that would be amazing. Otherwise, I will start installing the JDK in the global systems.

srdo:
@blockjon-dd If you need it urgently, you could just point Bazel at this branch and use this immediately?

shs96c (approved):
Ship it!

hypdeb:
For anyone else finding this and pulling their hair over `Unable to run coursier...`: adding a reference to `rules_jvm_external` to your `MODULE.bazel` is not properly reflected until you run a `bazel clean --expunge`.

PR #1327, opened 2025-02-11
Update bazel_deps and fix building for some examples

cheister:
Fixing the build for the java-export and kt_jvm_export examples and added them to the build.  Also updated the deps used by the protobuf-java example


.bazelci/examples.yml
```diff
@@ -1,45 +1,23 @@
 ---
 tasks:
-  simple-linux:
```
cheister, line 3: There is a lot of noise in this file because I sorted the entries when I added the java-export and kt_jvm_export examples.

examples/java-export/.bazelversion
```diff
@@ -1 +1 @@
-7.1.0
+7.4.1
```
shs96c, line 1: 7.5.0 is the latest stable 7.x release

examples/java-export/src/main/java/com/github/bazelbuild/rulesjvmexternal/example/export/Main.java
```diff
@@ -7,8 +7,8 @@
 public class Main {
 
   public static void main(String[] args) {
-    var fromGuava = ImmutableList.of("Hello", "World!");
-    var fromJre = List.of("Hello", "World!");
+    List<String> fromGuava = ImmutableList.of("Hello", "World!");
```
shs96c, line 10: The lists returned by `List.of` are immutable. We can replace the guava usage with standard JDK here

shs96c (approved):
If you have time, could you bump the Bazel version to 7.5.0, please?

examples/java-export/src/main/java/com/github/bazelbuild/rulesjvmexternal/example/export/Main.java
cheister: I thought the point of the example was to use both Guava and JRE Lists and compare them?

examples/java-export/.bazelversion
cheister: Ok, I'll update all bazel versions from 7.4.1 to 7.5.0

PR #1332, opened 2025-02-24
Allow `:outdated` to handle empty versions by skipping that artifact

shs96c:
This also switches us from a prebuilt artifact to calling the regular `java_binary`, which makes things a bit simpler to manage.


cheister:
Would it be better to make this 2 PRs?  And is there a test case we can add for the missing version?

cheister:
> Would it be better to make this 2 PRs? And is there a test case we can add for the missing version?

nvm, it looks like the PR is meant to be replaced by https://github.com/bazel-contrib/rules_jvm_external/pull/1338 ?

PR #1336, opened 2025-02-26
Implement different scope behavior for generated pom files

vinnybod:
Background: https://github.com/bazel-contrib/rules_jvm_external/issues/1335

TL;DR - https://github.com/bazel-contrib/rules_jvm_external/pull/1113 and https://github.com/bazel-contrib/rules_jvm_external/pull/1291 regressed in 6.7

The scopes seem counter-intuitive, but "scope" in Maven and Bazel are different, and I think this is the right path forward.
Bazel -
   `deps` are available at compile time and runtime, and are not leaked to dependents
   `runtime_deps` are available at runtime, but not compile time, and are not leaked to dependents
   `exports` leaked to dependents as compile and runtime dependencies
Maven -
   `compile` means the dependency is available at compile time and runtime, and is leaked to dependents
   `runtime` means the dependency is available at runtime, but not compile time. To promote good build hygiene,
                    this should be the default scope for the generated pom.xml because it means that we don't
                    leak our dependencies to dependents.

So putting all this together, this is the mapping:
Bazel `deps` -> Maven `runtime`
Bazel `runtime_deps` -> Maven `runtime`
Bazel `exports` -> Maven `compile`


vinnybod:
Keeping as a WIP to wait for discussion and to give some more time for me to test against our internal repos.

fmeum:
Can't we just replace the `compile_deps` with `export_deps`? I don't think the former are used for anything else, I introduced them in my PR.

vinnybod:
Yes, that makes sense. I'll make that change.
I am also seeing some issues with javadoc generation in 6.7 in our internal repos, so I am looking into whether that is related to these changes.

vinnybod:
Unfortunately it looks like https://github.com/confluentinc/rules_jvm_external/pull/31 has also regressed in 6.7
I'll be sure to include tests next time 😅

tests/unit/runtime_scope/runtime_scope_test.bzl
```diff
@@ -16,12 +16,11 @@ def _runtime_scope_test_impl(ctx):
...
         "example:runtime_dep_leaf:1.0.0",
     ], sorted(tut[MavenInfo].maven_deps.to_list()))
+
     asserts.equals(env, [
         # keep sorted
         "example:exported_leaf:1.0.0",
-        "example:leaf:1.0.0",
-        "example:runtime_dep_and_dep_leaf:1.0.0",
-    ], sorted(tut[MavenInfo].maven_compile_deps.to_list()))
+    ], sorted(tut[MavenInfo].maven_export_deps.to_list()))
```
vinnybod, line 23: Not sure if these tests are still needed, or we could rename it to `export_scope`

tests/unit/javadocs/BUILD
```diff
@@ -0,0 +1,29 @@
+load("@rules_java//java:defs.bzl", "java_library")
+load("//:defs.bzl", "java_export")
+
+# the apps-docs target previously failed with this setup.
+# now it succeeds.
+# todo: how to wrap this in a test target?
```
vinnybod, line 6: Could use some guidance on this.

fmeum: Just add an `alias` to it, that will build it by default.

tests/unit/runtime_scope/runtime_scope_test.bzl
fmeum: Let's keep it, but rename and adapt it to the new semantics.

private/rules/maven_utils.bzl
```diff
@@ -97,7 +98,22 @@ def generate_pom(
         include_version = dep in versioned_dep_coordinates
         unpacked = _unpack_coordinates(dep)
 
-        new_scope = "compile" if dep in versioned_compile_dep_coordinates_set else "runtime"
+        # This seems counter-intuitive, but "scope" in Maven and Bazel are different.
```
shs96c, line 101: Thank you for adding this comment. It's really useful.

shs96c (approved):
LGTM. @vinnybod, I'll wait until you close the comments from @fmeum before merging, but thank you for adding the tests!

fmeum (approved):

vinnybod:
@shs96c just fyi this is ready to go

PR #1342, opened 2025-03-13
Fix a fatal error when printing verbose logs when depedencies contain exclusions

vinnybod:
When using `RJE_VERBOSE=1`, the resolution fails when logging this message because it was expecting two parameters in `.format`


shs96c (approved):
Nice catch. Thank you!

PR #1353, opened 2025-04-12
Make rules_java_external run without bash on Windows

crt-31:
With this PR, rules_java_external can now run without Bash on Windows. This makes it way better for Windows users, and helps address hermeticity issues when bazel rules use bash. 

This should address this comment:
https://github.com/bazel-contrib/rules_jvm_external/blob/d8af22108bd8b353a226140570008231f2921931/private/rules/coursier.bzl#L306

Overview:
- I added a new rule bat_binary(), which is a rule I have used successfully in other projects. It should handle runfiles and subcalls properly.
- The main functionality that needed to be adapted to bat_binary were: pin, outdated, and maven_publish. Also deps.bzl now uses skylib's copy_file() instead of a genrule.
- The urls.bzl was changed to make Windows URL correctly.
- CI now has a windows_bashless configuration to build and test in an environment without bash. (There is still a windows_withbash configuration that allows rje to use bash in //tests/...)

So to keep rules_java_external bashless on windows moving forward,  we need to watch where we use sh_binary(), and genrule() to ensure there's a bashless equivelent functionality.  Tests however can still use bash, (but bazel_run_tests should be able to run bashless on windows).

### DIFF
--- a/.bazelci/examples.yml
+++ b/.bazelci/examples.yml
@@ -102,12 +102,15 @@ tasks:
     working_directory: examples/kt_jvm_export
     build_targets:
       - "//..."
-  kotlin-jvm-export-windows:
-    name: "kt_jvm_export example"
-    platform: windows
-    working_directory: examples/kt_jvm_export
-    build_targets:
-      - "//..."
+
+  #this is failing since rules_kotlin v2.1.3 due to bug.
+  # Uncomment this once https://github.com/bazelbuild/rules_kotlin/issues/1300 is fixed
+  #kotlin-jvm-export-windows:
+  #  name: "kt_jvm_export example"
+  #  platform: windows
+  #  working_directory: examples/kt_jvm_export
+  #  build_targets:
+  #    - "//..."
   pom-file-generation-linux:
     name: "POM file generation example"
     platform: ubuntu1804
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -99,12 +99,8 @@ tasks:
       - bazel run @regression_testing_coursier//:pin
       - bazel run @regression_testing_maven//:pin
       - bazel run @maven_install_in_custom_location//:pin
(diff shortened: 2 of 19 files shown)


shs96c:
I'm on vacation at the moment, but will review this properly once I return. Until then, I'm not sure why this is needed, as I thought bash was a requirement for the regular Bazel install. Is there a discussion somewhere about rolling back that requirement?

crt-31:
> I thought bash was a requirement for the regular Bazel install. Is there a discussion somewhere about rolling back that requirement?

My understanding is Bazel reliance on Bash was removed in Bazel 1.0.0. https://github.com/bazelbuild/bazel/issues/4319.

shs96c:
Apologies for the delay getting back to you. I _think_ there's a fairly widespread assumption that Bazel has a unix shell available to it, but I'm asking around in the community and from the project leads to figure out what the right direction to go in is.

shalupov:
2 cents from a Bazel user (rolling out Bazel for >500 developers, some of them using Windows), it could be an assumption that Bazel has a unix shell available, but for us we have nothing on developer's machine and not very eager to install stuff for everyone. Hermeticity is a way better for us.

shs96c:
While I appreciate this PR and the idea behind it, we aren't in a position to ensure that it continues to work. While it would be useful for Windows users not to need a unix shell available, I think it's inescapable for any moderately sized repo.

Because of this, I'm closing this PR, but please feel free to reopen if you can suggest some ways to ensure that the Windows-specific code paths are updated and maintained in sync with the unix ones, given that none of the committers of this repo have access to a Windows machine.

shalupov:
@shs96c Sorry for the late reply. The easiest way to write scripts IF you already have a jvm, is to call `bin/java` (*) actually. It's a very cross platform way to write scripts once. Using of unix shell is not hermetic therefore regardless of other rulesets it's not recommended.

(*) https://openjdk.org/jeps/330

PR #1359, opened 2025-05-08
[java_export]: Smarter handling of duplicate file names

shs96c:
In #1239 we essentially allowed anything that wasn't a class file to be included in the jar more than once. While this was useful in some cases it meant that if we pulled in dependencies generated by (eg) an aspect, such as when using protobufs and packing the original proto files, this can lead to unwanted to duplicates.

We now allow users more control of what to include, though we always include files that are often legally required (eg. the LICENSE file) which should strike a balance between flexibility and Doing The Right Thing by default.


cheister (approved):

shs96c:
The failing CI pipeline fails for all the incoming PRs. I don't think it's related.

PR #1365, opened 2025-05-19
Add documentation for maven.override when using bzlmod

cheister:
Small documentation update to explain in bzlmod one should use maven.override instead of override_targets

### DIFF
--- a/README.md
+++ b/README.md
@@ -845,10 +845,11 @@ In this case, once pinning is complete, guava `23.3-jre` will be selected.
 
 ### Overriding generated targets
 
-You can override the generated targets for artifacts with a target label of your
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -865,6 +866,15 @@ maven_install(
 )
 ```
 
+When you are using bzlmod you can override the generated target with
+```
+maven.override(
+    name = "maven",
+    coordinates = "com.google.guava:guava",
+    target = "@//third_party/guava:guava",
```
shs96c, line 874: Nit: when using `bzlmod`, you don't need the leading `@` in labels.

shs96c (approved):
LGTM

PR #1369, opened 2025-05-26
Update maven-metadata.xml when publishing locally

AustinSchuh:
When publishing to a file:// URL, the resulting repository isn't directly usable without updating maven-metadata.xml.  gradle does this automatically, let's make Bazel do it too.


shs96c (approved):
LGTM, though I'm sure there's an "official" maven api for generating the metadata xml.

PR #1375, opened 2025-06-11
Write exclusions to pom.xml

vinnybod:
This change adds “exclusions” to pom files in two different ways. The results of which are combined for the generated pom.xml.

**per java_export target**

This is useful for when an exclusion is required for a target that is built within the repo.

```python
java_export(
  name="my_dep",
  ...
  maven_coordinates="com.example:my_dep",
  deps=["log4j:log4j"]
)

java_export(
  name="my_lib",
  ...
  maven_coordinates="com.example:my_lib"
  exclusions={
    ":my_dep": ["log4j:log4j"]
  }
)
```

**maven.artifact declarations**

Passes through exclusions defined by the user on the `maven.artifact` to the lock file, which then appear on the generated pom.xml.

```python
# bzlmod
maven.artifact(
    artifact = "guava",
    exclusions = ["com.google.errorprone:error_prone_annotations"],
    group = "com.google.guava",
    version = "31.1-jre",
)

# WORKSPACE
maven_install(
    name = "pom_exclusions_testing",
    artifacts = [
        maven.artifact(
            artifact = "guava",
            exclusions = [
                maven.exclusion(
                    artifact = "error_prone_annotations",
                    group = "com.google.errorprone",
                ),
            ],
            group = "com.google.guava",
            version = "31.1-jre",
        ),
    ],
)
```

Exclusions set by `excluded_artifacts` are not passed along to the `pom.xml`.

----

Some decisions that we discussed but am still open to discussing:

* Does it make to propagate the `excluded_artifacts`?
We decided no on this because it seems like the wrong thing to do. If I for example add `excluded_artifacts=["log4j:log4j"]`, I want to completely exclude it from being brought into my repo via the coursier resolver. But that doesn't necessarily mean I want to add this exclusing to **every single** dependency entry in the pom.xml.

* Is it okay that we essentially ignore how Coursier outputs its exclusions?
I think yes. The current codebase wasn't doing anything with that information at that point. You can see in the new generated lockfiles, all that's changed is a new `exclusions` section.


shs96c:
@vinnybod, this seems like a really good idea. Can you please ensure it works with the `maven` resolver too?

shs96c:
The test failures look related too.

vinnybod:
@shs96c If you have as chance can you help me narrow down what might be happening in the 2 jvm_import_test failures? I can see that the url for the artifact is a different repository than what is expected, but I'm not sure what could have changed there.

RE: Maven, I'll take a look. Hopefully it's not too much extra work.

shs96c:
The tests look like they're failing because the order of the repositories has changed....

JohnnyMorganz:
Excited to see this happening!

We have a local fork that handles `maven.install` exclusions as well, which also parses tags from targets as part of `has_maven_deps` aspect for `MavenInfo`. However, we were able to skip the indirection via the lockfile, and generate those tags directly. The relevant input is still accessible when `generate_imports` is called in `coursier.bzl`, using `repository_ctx.attr.attrifacts`, so we construct a list of artifact exclusions using that instead. Not saying you should change your approach though, I think this way is neat too - just wanted to mention an alternative.

> Does it make to propagate the excluded_artifacts?

We are quite 50/50 on this one. On the one hand, indeed it ends up injecting an exclusion for every single dependency regardless of if it is actually part of the dependency tree or not, which is a bit ugly. On the other hand, I think it is technically more correct for us to do so. In our case, we export artifacts from our Bazel repository that are consumed downstream by other Maven repositories at our company. If we don't include these exclusions, then technically the exported POMs don't truly represent the dependency tree that Bazel is actually using.

Consider the case where `excluded_artifacts = ["log4j:log4j"]`, and we *don't* explicitly exclude `log4j:log4j` from the individual artifacts. We then bring in `some.group:external-dep` that had a published dependency on `log4j`, but that was removed in Bazel due to our global exclusions list. If we then publish `my_lib` which depends on `external-dep`, the exported pom does not include the `log4j:log4j` exclusion that Bazel has, and then log4j ends up on the classpath for our downstream Maven consumers when it shouldn't.

vinnybod:
@JohnnyMorganz 
>We are quite 50/50 on this one. On the one hand, indeed it ends up injecting an exclusion for every single dependency regardless of if it is actually part of the dependency tree or not, which is a bit ugly. On the other hand, I think it is technically more correct for us to do so

I'd be happier to hide this behavior behind a flag, given that we both came to the same conclusion independently that it is a bit more controversial. If we do so, I would prefer to have that change in a follow up PR.

vinnybod:
I have updated this PR to include support for the gradle resolver.

I am running into issues trying to repin the android example lockfiles, even after running https://github.com/bazel-contrib/rules_jvm_external/blob/d61367cc18c6178838655446dbcdcb76c1bf4b7d/README.md#installing-the-android-sdk-on-macos , but otherwise the tests are all passing.

vinnybod:
I updated this to skip the lockfile and its way less code

private/artifact_utilities.bzl
```diff
@@ -1,73 +1,31 @@
 #
-# Utilites for working with artifacts
+# Utilities for working with artifacts
 #
 
-load("//:specs.bzl", "utils")
-
-def deduplicate_and_sort_artifacts(dep_tree, artifacts, excluded_artifacts, verbose):
+def deduplicate_and_sort_artifacts(dep_tree, artifacts, verbose):
```
shs96c, line 5: I don't think this is de-duplicating any more, which seems less than ideal

cheister:
I'm not following why `private/artifact_utilities.bzl`  and  `tests/unit/artifact_utilities_test.bzl` have changed?  If you undo the changes to both of these files then all of your tests still pass.  These two files only handle resolving duplicate dependencies returned from coursier, how do the duplicated coursier artifacts affect adding exclustions to exported pom files?

cheister:
Also, do you mind squashing the commits for the PR?

tests/custom_maven_install/from_files_install.json
vinnybod, line 1: not 100% sure why this needed to be repinned.

tests/bazel_run_tests.sh
```diff
@@ -339,10 +339,6 @@ function test_gradle_metadata_is_resolved_correctly_for_jvm_artifact {
   expect_log "@regression_testing_gradle//:com_squareup_okio_okio_jvm"
 }
 
-function test_gradle_versions_catalog {
```
cheister, line 342: Was this deletion intentional?

vinnybod: I think that some of these changes were brought in accidentally when rebasing/squashing. I will review.

vinnybod:
There's now errors from the lockfiles. I tried running the same commands locally and they pass.

cheister:
LGTM.  Thanks for following up on the changes

cheister (approved):

PR #1378, opened 2025-06-18
Allow the same coordinate to be overridden in different repos

cheister:
Allow the same coordinate to be overridden in different maven.install targets


cheister:
Should fix https://github.com/bazel-contrib/rules_jvm_external/issues/1180

PR #1381, opened 2025-06-19
Allow root module's override tags to take precedence over the overridees from transitive deps.

sumeetgajjar:
#1278 author's [bandwidth is limited at the moment](https://github.com/bazel-contrib/rules_jvm_external/pull/1278#issuecomment-2954772387); As suggested by the OG author, this PR forks the same and addresses the review comments to get these changes merged.

---

Occasionally, a `bazel_dep` can introduce a `maven.overrides` tag that aliases an artifact to another label. This may not be what the root module desires, and so we should allow the root module's overrides to take precedence over any transitive overrides, but at the caution of the root module owner.
 
Concretely, this is a problem encountered by the Bazel project itself, where a `grpc-java` transitive dep overrides `@maven//:com_google_protobuf_protobuf_java` as an alias to `@@protobuf+//:protobuf_java`, which was not desirable.
 
This PR is validated by a test that contains a root override from okhttp to javapoet, which takes precedence over a dependency's override from okhttp to a poison pill label.


sumeetgajjar:
The CI's failing for [kt_jvm_export example](https://buildkite.com/bazel/rules-jvm-external-examples/builds/4821#01978597-fd11-40fc-967c-4a493917bb6e), however, it's been failing for quite some time now, so I'm assuming it's a non-issue.

tests/integration/override_targets/BUILD
```diff
@@ -69,3 +69,20 @@ sh_test(
...
+    name = "root_module_can_override",
+    expression = "deps(@root_module_can_override//:com_squareup_okhttp3_okhttp)",
+    opts = [
+        "--nohost_deps",
+        "--noimplicit_deps",
+    ],
+    scope = ["@root_module_can_override//:com_squareup_okhttp3_okhttp"],
+)
+
+sh_test(
```
shs96c, line 83: Elsewhere we've used a `diff_test` to check this kind of thing. There's an example of how to ignore the test when not using `bzlmod` [here](https://github.com/bazel-contrib/rules_jvm_external/pull/1373/files#diff-b7f7f72456bab30f131cdd91dd097e0f9555b53963ad19c635b5693687621022R119-R125) It would be nice to continue using the same approach here, if only for consistency in our testing.

MODULE.bazel
```diff
@@ -775,6 +775,26 @@ dev_maven.install(
...
+    artifacts = ["com.squareup:javapoet:1.11.1"],
+)
+
+bazel_dep(name = "transitive_module_can_override", version = "0.0.0", dev_dependency = True)
+local_path_override(
+    module_name = "transitive_module_can_override",
+    path = "tests/integration/override_targets/module",
+)
+
+dev_maven.override(
```
shs96c, line 790: Can the comment please mention the test this is used in?

shs96c (commented):
Love it, and it's sorely needed. Thank you for putting this together. Just a couple of nits, and then I think this is good to go!

tests/integration/override_targets/BUILD
sumeetgajjar: Makes sense.
This would require refactoring the OG tests added by @jin 
Will push the changes by tomorrow.

PR #1390, opened 2025-06-26
Allow coursier artifacts to have empty versions provided by BOMs and don't inlcude them in outdated

cheister:
Fix the error
```
ERROR: rules_jvm_external/specs.bzl:259:65: An error occurred during the fetch of repository '_main~maven~unpinned_coursier_resolved_with_boms':
	File "rules_jvm_external/private/rules/coursier.bzl", line 1072, column 38, in _coursier_fetch_impl
		dep_tree = make_coursier_dep_tree(
	File "rules_jvm_external/private/rules/coursier.bzl", line 844, column 48, in make_coursier_dep_tree
		coordinates = utils.artifact_coordinate(a)
	File "rules_jvm_external/specs.bzl", line 259, column 65, in _artifact_to_coord
		return artifact["group"] + ":" + artifact["artifact"] + ":" + artifact["version"] + classifier + type
Error: unsupported binary operation: string + NoneType
```
when using BOMs with coursier and not specifiying a version for an artifact.

Also don't check artifacts that get their versions from BOMs when running outdated


shs96c (approved):
LGTM

PR #1392, opened 2025-07-01
Add dll, dylib and so types to maven package mappings

cheister:


cheister:
> Ooof. I worry slightly that this is an unbounded list, but this is definitely helpful to people. LGTM

This at least covers the cases from https://github.com/bazel-contrib/rules_jvm_external/blob/master/private/dependency_tree_parser.bzl#L95

PR #1395, opened 2025-07-03
Fix resolution of Android/AAR artifacts with Gradle resolver

smocherla-brex:
The AAR integration is broken right now and it turns out AAR resolution in the dependency graph is broken, and for some reason I didn't notice it. This is discussed in https://github.com/bazel-contrib/rules_jvm_external/pull/1394 as well.

Essentially, the default `runtimeClasspath` configuration only resolves artifacts for the JVM platform with gradle. If we want to resolve artifacts for other platforms like Android, we either need to configure the android plugin (this is a real rabbit hole with the gradle tooling API I do not recommend) or add additive configurations to resolve them and add to the single dependency graph we produce. I do the latter here with a detached configuration which gradle provides (which is a clean slate). We only apply the detached configuration to dependencies that couldn't be resolved by `runtimeClasspath`, and then add nodes to the dependency graph from it. After that, we collect other artifacts like AARs from this configuration's ArtifactView. Finally, I had to handle cycles in the formed graph because for some reason we encounter cycles with this approach. After this fix, I'm able to resolve AAR artifacts properly - I uncommented the test case and repinned the lockfile.

As you can see, we have the right target here

```
bazel query @regression_testing_gradle//:androidx_compose_foundation_foundation_layout_android --output=build
# /private/var/tmp/_bazel_smocherla/536b5c8af4e86fad57fc24f39380ac8a/external/_main~maven~regression_testing_gradle/BUILD:392:11
aar_import(
  name = "androidx_compose_foundation_foundation_layout_android",
  tags = ["maven_coordinates=androidx.compose.foundation:foundation-layout-android:1.5.0-beta01@aar", "maven_repository=https://maven.google.com", "maven_sha256=4984eb1b708b9060fb6e70455cf245f7b9604909fca7895c15ebfd8bfd678309", "maven_url=https://maven.google.com/androidx/compose/foundation/foundation-layout-android/1.5.0-beta01/foundation-layout-android-1.5.0-beta01.aar"],
  applicable_licenses = ["@regression_testing_gradle//:androidx_compose_foundation_foundation_layout_android_package_info"],
  deps = ["@regression_testing_gradle//:androidx_annotation_annotation", "@regression_testing_gradle//:androidx_compose_runtime_runtime", "@regression_testing_gradle//:org_jetbrains_kotlin_kotlin_stdlib_common"],
  aar = "@regression_testing_gradle//:androidx/compose/foundation/foundation-layout-android/1.5.0-beta01/foundation-layout-android-1.5.0-beta01.aar",
)
# Rule androidx_compose_foundation_foundation_layout_android instantiated at (most recent call last):
#   /private/var/tmp/_bazel_smocherla/536b5c8af4e86fad57fc24f39380ac8a/external/_main~maven~regression_testing_gradle/BUILD:392:11 in <toplevel>

```

This solution also offers a good template to resolve Kotlin Multplatform artifacts in the resolver which I hope to get to soon.


cheister:
Thanks for the PR!  LGTM.  

Do you mind uncommenting the `"androidx.compose.foundation:foundation-layout:1.5.0-beta01"` dep in the regression_test_gradle declared in the WORKSPACE file as well?  

And checking the formatting by running `bazel run //scripts:format` ?

PR #1408, opened 2025-07-08
[ci] Fix bzlmod failures caused by `protobuf` update

shs96c:


examples/bzlmod/MODULE.bazel
```diff
@@ -3,6 +3,9 @@ module(
     version = "6.0",
 )
 
+# Required because `protobuf` contributes to the default namespace
+bazel_dep(name = "protobuf", version = "29.0", repo_name = None)
```
meteorcloudy, line 7: `repo_name = None` won't work with Bazel 7 though, so if you want to keep it here, you need to disable Bazel 7 for the test module in bcr presubmit.

shs96c: In the Bazel 7 world, we only support Bazel 7.6.1, which does support `repo_name = None`

PR #1411, opened 2025-07-10
Fetch transitive POMs in gradle resolver

### LINKED ISSUE #1409: When using gradle resolver, some dependencies are missing from the final lock_file (okhttp -> okio)
Hi:

Thanks for the new resolver 🙏 I tried it out and found out that in some cases dependencies are not properly resolved. I did put together a small sample project, just follow the links in the description below:

* When adding a [dependency](https://github.com/acecilia/gradle-to-bazel/blob/main/okio-missing/MODULE.bazel#L22) to `okhttp 4.12.0`
* Looking at the POM file in [here](https://repo1.maven.org/maven2/com/squareup/okhttp3/okhttp/4.12.0/okhttp-4.12.0.pom), I can see it has 2 dependencies: `kotlin-stdlib-jdk8` and `okio`
* During [verbose pinning](https://github.com/acecilia/gradle-to-bazel/blob/main/okio-missing/Makefile#L1), I can see `okio` being downloaded
* But after pinning the dependencies, inside the lockfile `okio` is missing. See [here](https://github.com/acecilia/gradle-to-bazel/blob/main/okio-missing/maven_install.json#L95)

smocherla-brex:
Fixes #1409, also https://github.com/bazel-contrib/rules_jvm_external/issues/909 is tangentially related (note that this PR doesn't solve all multiplatform cases, but it should fix a bug where we weren't downloading or pinning all transitive KMP artifacts for the JVM platform).

https://repo1.maven.org/maven2/com/squareup/okio/okio/3.15.0/ `okio` is a transitive dep of `okhttp`. It also turns out to be a Kotlin Multiplatform dependency with gradle metadata like

```
 {
      "name": "jvmRuntimeElements-published",
      "attributes": {
        "org.gradle.category": "library",
        "org.gradle.jvm.environment": "standard-jvm",
        "org.gradle.libraryelements": "jar",
        "org.gradle.usage": "java-runtime",
        "org.jetbrains.kotlin.platform.type": "jvm"
      },
      "available-at": {
        "url": "../../okio-jvm/3.15.0/okio-jvm-3.15.0.module",
        "group": "com.squareup.okio",
        "module": "okio-jvm",
        "version": "3.15.0"
      }
    },
```

basically pointing to the JVM artifact to be downloaded. The JAR for `okio` is basically a placeholder  https://repo1.maven.org/maven2/com/squareup/okio/okio/3.15.0/okio-3.15.0.jar (it's not actually a valid jar) but right now we don't download it:
- because gradle doesn't give us that as an artifact (because it's not in the gradle metadata)
- we should get the POM file - but right now we fetch only the roots/POMs for artifacts specified by the user, and we request the POMs explicitly by their extension, but that won't fetch the transitive POMs. As a result, in the current logic we fail to fetch some transitive artifacts which are Kotlin Multiplatform/gradle metadata based dependencies as they are essentially children (`okio-jvm`, `okio-js`,...) of the base `okio` dependency .
- To fix this, I've updated the logic to fetch the JARS for the base multiplatform coordinates like `okio` and fetch the POM for it, and indirectly its JAR. Even if this JAR is not useful in itself, it will ensure we get the children artifacts like `okio-jvm` and make them available in the final graph so that it can be used. This approach will help with other KMP dependencies as well.

I'm open to other approaches, but I updated the regression tests and the existing tests pass with this approach.


shs96c (approved):
LGTM. I think. Thank you for adding the extra tests!

PR #1412, opened 2025-07-11
Add support for pinning snapshots in gradle resolver

acecilia:
Hi 👋 

This PR adds support for snapshot dependencies in the gradle resolver. It uses the repository timestamp of the snapshot as a version.

Related: https://github.com/bazel-contrib/rules_jvm_external/pull/974


private/tools/java/com/github/bazelbuild/rules_jvm_external/resolver/remote/Downloader.java
```diff
@@ -84,11 +84,7 @@ public DownloadResult download(Coordinates coords) {
       return result;
     }
 
-    // Are we dealing with a packaging dep? Download the `pom.xml` and check
-    String originalTarget = coords.toRepoPath();
-    String pomName = String.format("%s-%s.pom", coords.getArtifactId(), coords.getVersion());
-    String pom = Paths.get(originalTarget).getParent().resolve(pomName).toString();
-
+    String pom = coords.setExtension("pom").toRepoPath();
```
acecilia, line 87: Not sure why the special handling of POM files above 🤔

tests/custom_maven_install/regression_testing_gradle_install.json
```diff
@@ -243,6 +243,37 @@
...
+        "jar": "cbfc3906b19b8f55dd7cfd6dfe0aa4532e834250d7f080bd8d211a3e246b59cb"
+      },
+      "version": "1.0.3"
+    },
+    "com.google.guava:guava": {
+      "shasums": {
+        "jar": "79abe7953803b20d1deb5f36283d4d640c7aacbdb1ddd7601aae369b1633e903"
+      },
+      "version": "999.0.0-HEAD-jre-SNAPSHOT",
+      "version_revision": "999.0.0-HEAD-jre-20250623.150948-114"
```
acecilia, line 263: Here is now the snapshot version pinning looks like, including the timestamp

acecilia:
Thanks for the feedback. Let me ask for some more context to understand a bit better your thoughts 😄 

> Snapshots aren't always stored with a version. Sometimes, the -SNAPSHOT.jar is just silently replaced

By "version", do you mean "timestamp"? If yes, is there maybe a public snapshot you could share that I could use during development, and also to add tests?

> We should have a consistent story for handling snapshots throughout the ruleset

What do you mean here, that snapshots should be handled by all resolvers? If yes, I agree. I was planning to look into the other resolvers after merging this PR. Is that okey, or would you prefer do do it differently?

> When we create the lockfile, omit the sha256 of any snapshot jar that doesn't have a fixed version (likely still contains SNAPSHOT.jar in the file name). That means that if the version is updated in place, builds won't break unexpectedly.

Let me expand a bit on my comment [here](https://github.com/bazel-contrib/rules_jvm_external/issues/901#issuecomment-3061649472). How would this work in bazel?
* Bazel uses `http_file` rule in [here](https://github.com/bazel-contrib/rules_jvm_external/blob/d61367cc18c6178838655446dbcdcb76c1bf4b7d/private/extensions/download_pinned_deps.bzl#L38) to download the artifacts. That rule has a sha256 property that is optional, but the [docs](https://bazel.build/rules/lib/repo/http#http_file-sha256) mention that _It is a security risk to omit the SHA-256 as remote files can change. At best omitting this field will make your build non-hermetic_ - we would be accepting these risks when adding support for such kind of snapshots
* For `http_file` rules without sha256, bazel would keep the downloaded artifacts in the cache, and will never try to update them. How do you envision that someone would trigger a snapshot update in this case? My understanding is that gradle caches snapshots but still fetches them once a day, but that is not the case for bazel
* My current understanding is that time-stamping snapshots is the default behaviour for maven repositories. I acknowledge that when searching in google "maven snapshots without timestamp" there are some ways to do it mentioned (example [here](https://maven.apache.org/plugins-archives/maven-deploy-plugin-2.6/examples/disabling-timestamps-suffix.html) and [here](https://stackoverflow.com/questions/1243574/how-to-stop-maven-artifactory-from-keeping-snapshots-with-timestamps)), but I also observed that those mentions are usually +10 years old. Furthermore, when trying to find more recent resources about how to do this, I found [this](https://stackoverflow.com/a/70653852) from 3 years ago that mentions that _Support for uniqueVersion was deprecated in Maven 2.x, and removed entirely in Maven 3_, which is confirmed by maven documentation in [here](https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=2852668#Maven3.xCompatibilityNotes-Non-uniqueSnapshotDeployments). 

This way of integrating snapshots goes against bazel fundamentals + it is deprecated. This is what makes me question wether this ruleset wants/needs to support it. Wdyt?

acecilia:
@shs96c let me know 🙏

acecilia:
Got it thanks for that detailed answer 🙏  Added support for non-versioned snapshots in gradle resolver

tests/bazel_run_tests.sh
```diff
@@ -367,8 +379,10 @@ TESTS=(
   "test_transitive_dependency_with_type_of_pom"
   "test_when_both_pom_and_jar_artifact_are_available_jar_artifact_is_present"
   "test_when_both_pom_and_jar_artifact_are_dependencies_jar_artifact_is_present"
-  # "test_gradle_metadata_is_resolved_correctly_for_aar_artifact"
+  "test_gradle_metadata_is_resolved_correctly_for_aar_artifact"
```
acecilia, line 382: For context about this change see [here](https://github.com/bazel-contrib/rules_jvm_external/pull/1395/files#r2264116068)

private/rules/v2_lock_file.bzl
```diff
@@ -138,6 +139,7 @@ def _get_artifacts(lock_file_contents):
             "group": parts[0],
             "artifact": parts[1],
             "version": data["version"],
+            "version_revision": data.get("version_revision"),
```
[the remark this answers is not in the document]
acecilia: I added `version_revision` separately because I did not consider version and snapshot timestamp to be the same:
* `version`: this is the version obtained by calling `component.getModuleVersion().getVersion()` in [here](https://github.com/bazel-contrib/rules_jvm_external/pull/1412/files)
* `version_revision`: the timestamp (or revision) for a specific version

You can see a sample of how both look like in [here](https://github.com/bazel-contrib/rules_jvm_external/pull/1412/files#r2199375341).

Would you prefer I try to get rid of `version_revision`, merging both concepts into one?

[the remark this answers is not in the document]
acecilia: Hi, found some time to look into this. I tried following the approach you proposed: getting rid of `version_revision` and using existing `version` instead. 

I was unable to make it work because the `version` field and the `version_revision` are needed separately.

This is my understanding of where the difference is important:

* After the lockfile is created, the [Downloader.java](https://github.com/bazel-contrib/rules_jvm_external/blob/1c5cfbf96de595a3e23cf440fb40380cc28c1aea/private/tools/java/com/github/bazelbuild/rules_jvm_external/resolver/remote/Downloader.java#L177) downloads the `jar` files
* The download url for the `jar` files is calculated from the information in the lockfile. The url is passed to the downloader in [here](https://github.com/bazel-contrib/rules_jvm_external/blob/1c5cfbf96de595a3e23cf440fb40380cc28c1aea/private/tools/java/com/github/bazelbuild/rules_jvm_external/resolver/remote/Downloader.java#L194), and is calculated by the [toRepoPath](https://github.com/bazel-contrib/rules_jvm_external/blob/1c5cfbf96de595a3e23cf440fb40380cc28c1aea/private/tools/java/com/github/bazelbuild/rules_jvm_external/Coordinates.java#L126) method
* The URL format is: 
    * `<base_url>/<getGroupId>/<artifactId>/<version>/<artifactId>/<version>` - for non-snapshots
    * `<base_url>/<getGroupId>/<artifactId>/<version>/<artifactId>/<snapshotId>` for versioned snapshots (note that `snapshotId` in the context of this PR is `versionRevision`)

So for the guava snapshots, the download URL is: `https://oss.sonatype.org/content/repositories/snapshots/com/google/guava/guava/999.0.0-HEAD-jre-SNAPSHOT/guava-999.0.0-HEAD-jre-20250623.150948-114.jar` (you can paste it in the browser and you will see it successfully downloads the `jar` file)

If I use a single `version` to store both the snapshot version and the timestamp version, then the URL is incorrect: `https://oss.sonatype.org/content/repositories/snapshots/com/google/guava/guava/999.0.0-HEAD-jre-20250623.150948-114/guava-999.0.0-HEAD-jre-20250623.150948-114.jar`

private/tools/java/com/github/bazelbuild/rules_jvm_external/resolver/gradle/plugin/GradleDependencyModelBuilder.java
```diff
@@ -268,7 +268,17 @@ private GradleResolvedDependency walkResolvedComponent(
       info = new GradleResolvedDependencyImpl();
       info.setGroup(component.getModuleVersion().getGroup());
       info.setName(component.getModuleVersion().getName());
-      info.setVersion(component.getModuleVersion().getVersion());
+      String version = component.getModuleVersion().getVersion();
+      info.setVersion(version);
+
+      // For snapshot dependencies, extract the timestamped version from the component ID
+      // Note: this is unsafe, is there a better way of obtaining the timestamp for an snapshot? I could not find any
+      if (component.getId().getClass().getName().contains("MavenUniqueSnapshotComponentIdentifier")) {
```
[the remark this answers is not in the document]
acecilia: I tried to update it as suggested: using regex to identify snapshots. I could not use the exact regex tho, but very similar

private/rules/coursier.bzl
```diff
@@ -597,7 +597,9 @@ def _pinned_coursier_fetch_impl(repository_ctx):
             http_files.extend([
                 "    http_file(",
                 "        name = \"%s\"," % http_file_repository_name,
-                "        sha256 = \"%s\"," % artifact["sha256"],
+                # sha256 is optional: non-versioned snapshots may not have it
+                # See: https://github.com/bazel-contrib/rules_jvm_external/pull/1412
+                "        sha256 = \"%s\"," % artifact.get("sha256", None),
```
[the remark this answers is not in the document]
acecilia: Updated as suggested, thanks!

private/tools/java/com/github/bazelbuild/rules_jvm_external/resolver/gradle/models/GradleResolvedDependency.java
```diff
@@ -34,6 +34,10 @@ public interface GradleResolvedDependency {
 
   void setVersion(String version);
 
+  String getVersionRevision();
```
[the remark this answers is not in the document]
acecilia: See [here](https://github.com/bazel-contrib/rules_jvm_external/pull/1412#discussion_r2312126617) please

private/rules/v2_lock_file.bzl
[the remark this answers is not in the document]
H5-O5: > Why is this needed? Surely the version is one of:
> 
> 1. A regular version number for non-snapshot deps
> 2. `x.x.x-SNAPSHOT` for snapshot deps that aren't versioned
> 3. A snapshot timestamp for snapshot deps that are versioned
> 
> Looking at the maven code, it doesn't look like their abstractions use `version_revision` either.

I'd like to join the discussion here. I'm in favor of adding a new field.

The url is `<baseUrl>/<groupId>/<artifactId>/<version>/<artifactId>-<version>`, e.g. `https://oss.sonatype.org/content/repositories/snapshots/com/google/guava/guava/999.0.0-HEAD-jre-SNAPSHOT/guava-999.0.0-HEAD-jre-20250623.150948-114.jar`

Currently we do:

```java
    path.append(getGroupId().replace('.', '/'))
        .append("/")
        .append(getArtifactId())
        .append("/")
        .append(getVersion())
        .append("/")
        .append(getArtifactId())
        .append("-")
        .append(getVersion());
```

But this code only works if the version and revision is the same.

If we keep using single field, then it must be guava-999.0.0-HEAD-jre-20250623.150948-114.jar. We'll have to add codes like this:

```
if version looks like a snapshot version, then
  real_version = remove the timestamp and append "-SNAPSHOT"
else
  real_version = just the version
```

And as the codebase evolves, we may need to add such a hack everywhere. and the codebase is harder to maintain. It's just easier to use 2 fields to remember 2 parts of the URL, especially the first part "xxx-SNAPSHOT" and the second part "xxx-TIMESTAMP" are different.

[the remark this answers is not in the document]
acecilia: > I'd like to join the discussion here.

Appreciate it 🙌 

> At first I thought this is unnecessary and maintaining a new field is clearer

Yes, this is my thought also: having to decode `Coordinates.version` in all the places where it is used seems to me an unattainable and error-prone task.

> Upon further look I saw that in many places we have the equivalent of new Coordinates(old_coord.toString()). If we use 2 fields then there is no way to encode the same version-revision information in <groupId>:<artifactId>[:<extension>[:<classifier>][:<version>], but replacing version with revision and to part it on-the-fly can make the serialization and deserialization work as before.

Im sorry, I tried, but I do not understand what you mean here, I think I am missing some context. Could you maybe share an example that helps me grasp a bit better the comment 🙏

acecilia:
I am keen to complete this PR with your guidance, have a look when you have a chance please 🙏

private/tools/java/com/github/bazelbuild/rules_jvm_external/resolver/gradle/GradleResolver.java
```diff
@@ -275,7 +276,8 @@ private void addDependency(
                   childCoordinates.getArtifactId(),
                   extension,
                   childCoordinates.getClassifier(),
-                  childCoordinates.getVersion());
+                  childCoordinates.getVersion(),
```
smocherla-brex, line 279: What I was hoping here was substitute the versionRevision into the `version` field of `Coordinates` instead of the same version (if it was a snapshot one). Then in the downloader, it should try to download the actual revision of the snapshot? That way, you probably still need versionRevision in the gradle plugin not in `Coordinates` itself. I can try testing that query quickly as well. @acecilia

H5-O5: @acecilia I have an implementation of supporting snapshot for maven-resolver. I used the single field as @smocherla-brex suggested. The key is that we store the snapshot version (2025xxxx.xxxxxx-y) in Coordinates, and when downloading, calculate the real url. Have a look at https://github.com/bazel-contrib/rules_jvm_external/compare/master...H5-O5:rules_jvm_external:maven-snapshot and private/tools/java/com/github/bazelbuild/rules_jvm_external/Coordinates.java and it may help you get your MR aligned with the reviewer's request.

smocherla-brex: I took a shot to try to use just the single version field in `Coordinates`, it requires additional handling within Coordinates but it does seem to work https://github.com/smocherla-brex/rules_jvm_external/commit/6b6a293c9340b342226c3c03a5895ff6bd1ca3cb#diff-0884e6cc400f368585972b32d3105196826dc1bdfb803843798f30b24ecc87c1R370-R371 (this is on top of this branch/PR) and making sure we handle it in toRepoPath(). I have not tested everything though. I'm not necessarily opposed to a new field but it could increase the lockfile size potentially and I'll defer to Simon on the decision.

MODULE.bazel
```diff
@@ -764,12 +764,17 @@ dev_maven.install(
         "androidx.annotation:annotation:1.6.0",
         # https://github.com/bazel-contrib/rules_jvm_external/issues/1409
         "com.squareup.okhttp3:okhttp:4.12.0",
+        # Versioned snapshot pinning support: https://github.com/bazel-contrib/rules_jvm_external/pull/1412
+        "com.google.guava:guava:999.0.0-HEAD-jre-SNAPSHOT",
+        # Non-versioned snapshot pinning support: https://github.com/bazel-contrib/rules_jvm_external/pull/1412
+        "org.seleniumhq.selenium:selenium-java:4.34.0-SNAPSHOT",
```
smocherla-brex, line 770: You'll need to port this change here as well https://github.com/bazel-contrib/rules_jvm_external/blob/master/WORKSPACE#L390 in WORKSPACE to resolve the CI failures related to Bazel 6.

acecilia: Thanks added

tests/custom_maven_install/regression_testing_gradle_install.json
```diff
@@ -422,6 +525,114 @@
...
+      "shasums": {
+        "jar": "1fad6e6be7557781e4d33729d49ae1cdc8fdda6fe477bb0cc68ce351eafdfbab"
+      },
+      "version": "1.0.0"
+    },
+    "org.seleniumhq.selenium:selenium-api": {
+      "shasums": {
+        "jar": null
+      },
+      "version": "4.34.0-SNAPSHOT"
```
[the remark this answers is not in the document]
acecilia: Mmm I think this is because is selenium: the snapshot is non-versioned.

This PR contains 2 integrations:
* `guava`: versioned snapshot - with timestamps and `versionRevision`
* `selenium`: non-versioned snapshot - without timestamp nor `versionRevision`

Does it make sense now? Or am I missing something?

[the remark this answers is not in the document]
smocherla-brex: Got it. I was under the assumption that snapshot versions always mapped to a timestamp version. Makes sense.

shs96c:
Apologies for the long delay getting back to you. I've been traveling and sick. I want to review the `version_revision` as it's making me uneasy, and I'd like to be satisfied it's required.

private/tools/java/com/github/bazelbuild/rules_jvm_external/resolver/gradle/GradleResolver.java
H5-O5: @shs96c what's the status of this MR?

private/rules/v2_lock_file.bzl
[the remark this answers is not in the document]
acecilia: Updated as suggested: removed `version_revision` and instead encode the timestamp in the `version` field

private/tools/java/com/github/bazelbuild/rules_jvm_external/resolver/gradle/GradleResolver.java
acecilia: Thanks @shs96c for https://github.com/acecilia/rules_jvm_external/pull/1!

Now I hope I got it. Updated as suggested: removed version_revision and instead encode the timestamp in the version field

private/tools/java/com/github/bazelbuild/rules_jvm_external/resolver/remote/Downloader.java
```diff
@@ -174,6 +181,11 @@ private DownloadResult performDownload(Coordinates coordsToUse, String path) {
     }
 
     if (!downloaded) {
+      // If the only known path was a POM and the jar download failed, return a
+      // POM-only result rather than null so the caller can still use the POM.
+      if (hadPomOnlyKnownPath) {
+        return new DownloadResult(coordsToUse, Set.of(), null, null);
+      }
```
acecilia: Check the commit message in https://github.com/bazel-contrib/rules_jvm_external/pull/1412/changes/c1d50b096ceb9be8b3e43ce073691e6c114c645d for an explanation of this change (is fixing an existing bug introduced in the last master commit)

acecilia:
Ready for another review round

tests/custom_maven_install/regression_testing_gradle_install.json
```diff
@@ -440,11 +446,35 @@
...
+      "shasums": {
+        "jar": "cbfc3906b19b8f55dd7cfd6dfe0aa4532e834250d7f080bd8d211a3e246b59cb"
+      },
+      "version": "1.0.3"
+    },
+    "com.google.guava:guava": {
+      "shasums": {
+        "jar": "9f71d9a756937d784cd175af7e3a1992f088321435b880af02488003da4c43b2"
+      },
+      "version": "999.0.0-HEAD-jre-20260701.164618-384"
```
acecilia: Here is how a timestamped snapshot dependency shows: with the timestamp as part of the version

acecilia:
@shs96c Would appreciate a review when you have a chance

tests/custom_maven_install/regression_testing_gradle_install.json
```diff
@@ -440,11 +446,35 @@
...
+      "shasums": {
+        "jar": "cbfc3906b19b8f55dd7cfd6dfe0aa4532e834250d7f080bd8d211a3e246b59cb"
+      },
+      "version": "1.0.3"
+    },
+    "com.google.guava:guava": {
+      "shasums": {
+        "jar": null
+      },
+      "version": "999.0.0-HEAD-jre-20260828.195817-455"
```
acecilia: Here is how a timestamped snapshot dependency shows: with the timestamp as part of the version

tests/custom_maven_install/non_timestamped_snapshot_testing_gradle_install.json
```diff
@@ -0,0 +1,32 @@
...
+    "com.example:snapshot": 810239419,
+    "repositories": -1669196818
+  },
+  "__RESOLVED_ARTIFACTS_HASH": {
+    "com.example:snapshot": 1149605678
+  },
+  "artifacts": {
+    "com.example:snapshot": {
+      "shasums": {
+        "jar": null
```
acecilia, line 13: Here is how a non-timestamped snapshot dependency shows: with the shasum set to null

tests/custom_maven_install/timestamped_snapshot_testing_gradle_install.json
```diff
@@ -0,0 +1,32 @@
...
+  },
+  "__RESOLVED_ARTIFACTS_HASH": {
+    "com.example:snapshot": -552255756
+  },
+  "artifacts": {
+    "com.example:snapshot": {
+      "shasums": {
+        "jar": "07c82b630621b46fe45aff0e43de98feea968c9cb793f8229b91f12c5ef1e976"
+      },
+      "version": "1.0-20260829.120000-1"
```
acecilia, line 15: Here is how a timestamped snapshot dependency shows: with the timestamp as part of the version

PR #1416, opened 2025-07-14
maven.override: make override alias preserve strict visibility of overridden target

dmivankov:
Motivation is that override alias should represent and replace overridden target coordinates and should be subject to same visibility rules.

And before this change if alias target is public, but overridden target isn't, replacing it with public alias to public target makes it public subtly breaking strict visibility.

Resolves https://github.com/bazel-contrib/rules_jvm_external/issues/1415


shs96c (approved):
Good catch. Thank you! LGTM

PR #1423, opened 2025-07-21
Require `force_version` for root module artifact to take precedence

shs96c:
There was concern expressed by some owners of sub-modules that deliberately contribute to the `maven` repo that the behaviour of implicitly overriding artifacts by merely listing them in the root module would lead to users accidentally overriding those values. Instead, require the root module to use `force_version` to override the specific coordinate, which is more inline with the original intention of `force_version`.


private/extensions/maven.bzl
```diff
@@ -292,7 +301,7 @@ def _deduplicate_artifacts_with_root_priority(root_artifacts, non_root_artifacts
             else:
                 warning_parts.append(maven_module)
 
-        print("WARNING: The following maven modules appear in multiple sub-modules with potentially different versions. " +
+        print("WARNING: The following coordinates from `%s` appear in multiple sub-modules with potentially different versions. " % name +
               "Consider adding one of these to your root module to ensure consistent versions:\n\t%s" %
```
cheister, line 305: Should this message be updated to tell people they'll need force_version=True for the coordinates in the root module?

PR #1444, opened 2025-09-02
Fix issue where MavenPublisher hangs indefinitely

vinnybod:
The previous pull request https://github.com/bazel-contrib/rules_jvm_external/pull/1260 introduced an issue where publishing could hang indefinitely because it integrated the existing `HttpDownloader` into `MavenPublisher`. `HttpDownloader` contained an `EventListener` that was not being closed.

This updates the `HttpDownloader` to be auto-closeable and it closes the listener on close. `MavenPublisher` now uses it with a try-with-resources block.

There are now some end to end tests for MavenPublisher that tests local and http publishing. To make testing easier, we refactored the `MavenPublisher` entrypoint so that there is a `run` function that doesn't need to read from environment variables.


private/tools/java/com/github/bazelbuild/rules_jvm_external/maven/MavenPublisher.java
```diff
@@ -82,101 +86,136 @@
...
+        upload(
+            repo,
+            credentials,
+            coords,
+            "." + getFileExtension(mainArtifactPath),
+            Paths.get(mainArtifactPath),
+            signingMetadata,
+            executor));
+
+    if (!Strings.isNullOrEmpty(extraArtifacts)) {
```
vinnybod, line 188: empty string being passed for extraArtifacts was causing an IndexOutOfBounds exception

private/tools/java/com/github/bazelbuild/rules_jvm_external/maven/MavenPublisher.java
```diff
@@ -254,7 +293,7 @@ private static CompletableFuture<Void> uploadMavenMetadata(
                 versioning.setRelease(coords.version);
                 // This may be needed for SNAPSHOT support
                 String timestamp = new SimpleDateFormat("yyyyMMddHHmmss").format(new Date());
-                versioning.setLastUpdated("20200731090423");
```
vinnybod, line 257: This seemed like a bug. timestamp wasn't being used.

private/tools/java/com/github/bazelbuild/rules_jvm_external/resolver/remote/HttpDownloader.java
```diff
@@ -99,7 +99,7 @@ public static EventListener defaultEventListener() {
     boolean consoleAvailable = System.console() != null;
     if (System.getenv("RJE_VERBOSE") != null) {
       return new PlainConsoleListener();
-    } else if (termAvailable && consoleAvailable) {
+    } else if (termAvailable && consoleAvailable || System.getenv("FORCE_ANSI") != null) {
       return new AnsiConsoleListener();
```
fzakaria, line 103: I wanted to test the AnsiConsoleListener in a test to validate the cleanup; unfortunately Bazel test runner calls `System.exit` making it difficult to test it. [[slack](https://bazelbuild.slack.com/archives/CDCE4DFAP/p1756831399977859)]

private/tools/java/com/github/bazelbuild/rules_jvm_external/maven/MavenPublisher.java
```diff
@@ -82,101 +86,136 @@
...
-          "Repository must be uploaded to via the supported schemes: "
-              + Arrays.toString(SUPPORTED_UPLOAD_SCHEMES));
+    final String repo = System.getenv("MAVEN_REPO");
+    if (Strings.isNullOrEmpty(repo)) {
+      throw new IllegalArgumentException("MAVEN_REPO environment variable must be set");
+    }
+
+    // give friendly warning if true/false not explicitly set as args[3]
+    if (!"true".equalsIgnoreCase(args[3]) && !"false".equalsIgnoreCase(args[3])) {
+      LOG.warn(
```
shs96c, line 112: If the value isn't correct, we should bail. The only way this should be called is from our own code anyway, so being strict is fine.

private/tools/java/com/github/bazelbuild/rules_jvm_external/maven/MavenPublisher.java
```diff
@@ -82,101 +86,136 @@
...
+    if (!isSchemeSupported(repo)) {
+      throw new IllegalArgumentException(
+          "Repository must be accessed via the supported schemes: "
+              + Arrays.toString(SUPPORTED_SCHEMES));
+    }
 
-    try {
-      List<CompletableFuture<Void>> futures = new ArrayList<>();
-      futures.add(upload(repo, credentials, coords, ".pom", pom, signingMetadata));
+    if (!isUploadSchemeSupported(repo)) {
```
shs96c, line 164: Since the upload schemes are a subset of the supported schemes, this means that the difference of those sets can't be used any more with the publisher. `artifactregistry` and `gs` would both be excluded.

Looking at the previous code, this same error existed, but it should be fixed.

private/tools/java/com/github/bazelbuild/rules_jvm_external/maven/MavenPublisher.java
```diff
@@ -82,101 +86,136 @@
...
+      all = all.thenCompose(Void -> uploadMavenMetadata(repo, credentials, coords, executor));
     }
-  }
 
-  private static Path getPathIfSet(String arg) {
-    if (!arg.isEmpty()) {
-      return Paths.get(arg);
-    }
-    return null;
+    all.get(5, MINUTES);
```
shs96c, line 218: Why change this from 30 minutes?

private/tools/java/com/github/bazelbuild/rules_jvm_external/maven/MavenPublisher.java
```diff
@@ -305,22 +345,23 @@ private static CompletableFuture<Void> upload(
...
+      uploads.add(
+          upload(String.format("%s%s.asc", base, append), credentials, gpg_sign(item), executor));
     } else if (signingMethod.equals(MavenSigning.SigningMethod.PGP)) {
       uploads.add(
           upload(
               String.format("%s%s.asc", base, append),
               credentials,
-              in_memory_pgp_sign(
-                  item, signingMetadata.signingKey, signingMetadata.signingPassword)));
+              in_memory_pgp_sign(item, signingMetadata.signingKey, signingMetadata.signingPassword),
```
shs96c, line 363: Note to self: refactor this method name.

private/tools/java/com/github/bazelbuild/rules_jvm_external/maven/MavenSigning.java
```diff
@@ -234,11 +234,15 @@ protected enum SigningMethod {
     NONE
   }
 
-  protected static class SigningMetadata {
-    protected final String signingKey;
-    protected final String signingPassword;
+  public static class SigningMetadata {
+    public final String signingKey;
```
shs96c, line 238: Use IJ to generate getters, please. We shouldn't really be messing with fields directly.

PR #1449, opened 2025-09-17
Expose CreateJar as a rule

vinnybod:
In Confluent's fork of rules_jvm_external, we have a `scala_export` macro that utilizes the `CreateJar` tool. It was decided that the scala_export macro wouldn't be [upstreamed](https://github.com/bazel-contrib/rules_jvm_external/pull/1257) here. So I am reimplementing it in `rules_scala` and it needs this utility for creating the [scaladocs](https://github.com/bazel-contrib/rules_scala/compare/master...confluentinc:rules_scala:vinnybod/upstream-scala-export#diff-aab783120e43fe84f1484d1bd0222296feaf64a3e2314c60aad81a8277a7e0a9R110-R114) jar.

If we don't want to expose this utility publicly in rules_jvm_external, then I can just copy this file to the rules_scala change.

### DIFF
--- a/tests/com/github/bazelbuild/rules_jvm_external/jar/CreateJarTest.java
+++ b/tests/com/github/bazelbuild/rules_jvm_external/jar/CreateJarTest.java
@@ -1,0 +1,49 @@
+package com.github.bazelbuild.rules_jvm_external.jar;
+
+import static com.github.bazelbuild.rules_jvm_external.ZipUtils.readJar;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
+import java.io.File;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.Map;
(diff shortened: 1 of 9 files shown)


tests/com/github/bazelbuild/rules_jvm_external/jar/CreateJarTest.java
```diff
@@ -0,0 +1,49 @@
...
+    Path file1 = tempDir.resolve("file1.txt");
+    Files.writeString(file1, "content1");
+    Path subDir = tempDir.resolve("subdir");
+    Files.createDirectory(subDir);
+    Path file2 = subDir.resolve("file2.txt");
+    Files.writeString(file2, "content2");
+
+    // Create output jar path
+    Path outputJar = temp.newFile("output.jar").toPath();
+    ;
```
vinnybod, line 35:
```suggestion
```

PR #1454, opened 2025-09-29
Added non-conflicting hash for install files

MarconZet:
# Sumary

This commit introduces lock file version 3 with per-artifact hashing instead of a single global hash.

This per-artifact hashing approach can reduce the amount of merge conflicts when multiple people update canonical version in large monorepo.

The code still supports reading v2 lock files - it checks for v3 first, then falls back to v2, then v1. Users with older lock files will see a message to repin.

## Key Changes

1. Lock File Format Change (v2 → v3)

* Before (v2): __INPUT_ARTIFACTS_HASH and __RESOLVED_ARTIFACTS_HASH were single integer values
* After (v3): Both are now dictionaries mapping each artifact coordinate to its individual hash

Example in maven_install.json: 
```jsonc
// Old format 
"__INPUT_ARTIFACTS_HASH": 1994476565, 
"__RESOLVED_ARTIFACTS_HASH": -274973469,
// New format
"__INPUT_ARTIFACTS_HASH": { "com.google.guava:guava": 733518530, "junit:junit": -652553691, "..." }, 
"__RESOLVED_ARTIFACTS_HASH": { "com.google.guava:guava": -1587873388, "..." }
```

2. Hash Computation Changes (private/rules/v3_lock_file.bzl:53-108)

The new _compute_lock_file_hash_v3 function computes individual hashes per artifact that include:

* The artifact's own info (coordinates, SHA sums)
* The repository it came from
* Hashes of all transitive dependencies (dependency-aware hashing)

3. Input Hash Changes (private/rules/coursier.bzl:334-386)

compute_dependency_inputs_signature now returns a dictionary of per-artifact hashes plus backward-compatible v1/v2 signatures.


shs96c:
This is looking really good. I like the idea of only having conflicts if the transitive deps have changed.

private/rules/coursier.bzl
```diff
@@ -331,6 +331,13 @@ def _stable_artifact(artifact):
     keys = sorted(parsed.keys())
     return ":".join(["%s=%s" % (key, parsed[key]) for key in keys])
 
+def _add_to_hash_dictionary(dictionary, artifact, salt):
+    artifact_dict = json.decode(artifact)
+    key = artifact_dict["group"] + ":" + artifact_dict["artifact"]
```
shs96c: You should use the same logic that's in `Coordinates.asKey()` to get a stable key that includes things like the classifier. That's already in `v2_lock_file.bzl` as `to_key`, but if it's needed in more than one place, we should move it to the `coordinates.bzl` library file.

vinnybod:
FWIW, We've been using this at Confluent for a month now and it has been working well.

private/rules/coursier.bzl
MarconZet: hmm, this key has a direct mapping to what appears in the lock file

My idea was, that I don't want things like classifier and packaging, my reason being:
1) It looked ugly in the lock file – when I tried it in my repo, the `__INPUT_ARTIFACTS_HASH` size duplicated with `:sources`. It did not provide any information and was just noise
2) I think that this is at a level, where we don't want to allow for non-conflicting merges

PR #1456, opened 2025-10-05
Bump Robolectric to 4.16

utzcoz:
1. Bump minSdk to 21 as AndroidX bumped it before and dependencies have this restriction.
2. Bump AndroidX test junit version and add correct junit4 dependency version to avoid potential class cast issues when bumping dependencies.


utzcoz:
See https://github.com/robolectric/robolectric-bazel/releases/tag/4.16.

PR #1459, opened 2025-10-06
ci: Enable Local test examples on macOS

utzcoz:
A try to enable local test example running on macOS. Wait CI result.


utzcoz:
I think CI passed is what I mean result. I think it works like other CI job on macOS.

Issue #1477, opened 2025-11-05
Regression in 6.8: Build failures with pom artifacts when fetch_sources = True and lockfile enabled


thirtyseven:
## Issue
When a transitive dependency of a maven artifact is a `pom` and `fetch_sources` and `lock_file` are both enabled, building that maven artifact fails.

### Setup
MODULE.bazel
```
bazel_dep(name = "rules_jvm_external", version = "6.9") # Or 6.8

maven = use_extension("@rules_jvm_external//:extensions.bzl", "maven")
maven.install(
    artifacts = [ 
        "org.opencadc:cadc-util:1.12.10",
    ],  
    fetch_sources = True,
    lock_file = "maven_install.json",
)
use_repo(maven, "maven")
```

### Command
```
bazel build @maven//:org_opencadc_cadc_util
```

### Expected
Build succeeds

### Actual
```
ERROR: no such package '@@[unknown repo 'org_apache_logging_log4j_log4j_3_0_0_beta3' requested from @@rules_jvm_external++maven+maven (did you mean 'org_apache_logging_log4j_log4j_kit_3_0_0_beta3'?)]//file': The repository '@@[unknown repo 'org_apache_logging_log4j_log4j_3_0_0_beta3' requested from @@rules_jvm_external++maven+maven (did you mean 'org_apache_logging_log4j_log4j_kit_3_0_0_beta3'?)]' could not be resolved: No repository visible as '@org_apache_logging_log4j_log4j_3_0_0_beta3' from repository '@@rules_jvm_external++maven+maven'
[...]

PR #1479, opened 2025-11-17
coursier: ignore dependencies with classifier="sources" and no "file"

mattnworb:
when coursier is asked to resolve an artifact that has a transitive dependency on `org.apache.logging.log4j:log4j:3.0.0-beta3` (note the lack of packaging in the [real-world example here][0]) and `fetch_sources=True` is set, coursier will return this in the list of dependencies:

```
{
  "coord": "org.apache.logging.log4j:log4j:jar:sources:3.0.0-beta3",
  "file": null,
  "directDependencies": [],
  "dependencies": []
}
```

in rules_jvm_external 6.8 this will cause errors when building the external repo generated by RJE since RJE will end up handling this dependency by a) generating a `http_file` with an empty list of `urls` and b) emitting a `copy_file` rule in the external repo's BUILD file that refers to the non-existing `http_file` repo from A. See [this comment][1] for a breakdown of why this happens.

PR [#1207](https://github.com/bazelbuild/rules_jvm_external/issues/1207) added `pom` to the list of `SUPPORTED_PACKAGING_TYPES` so that the dependencies of the pom could be aggregated (Maven interprets a dependency on an artifact with packaging=pom as depending on the `<dependencies>` in that pom), but that PR didn't test what happens with `fetch_sources=True` nor did it consider the case like with `org.apache.logging.log4j:log4j:3.0.0-beta3` where the coordinates output by coursier don't mention the packaging at all.

fixes [#1477](https://github.com/bazelbuild/rules_jvm_external/issues/1477)

[0]: https://central.sonatype.com/artifact/org.opencadc/cadc-util/1.12.10
[1]: https://github.com/bazel-contrib/rules_jvm_external/issues/1477#issuecomment-3530832168

### DIFF
--- a/private/rules/coursier.bzl
+++ b/private/rules/coursier.bzl
@@ -1106,9 +1106,6 @@ def _coursier_fetch_impl(repository_ctx):
     # Once coursier finishes a fetch, it generates a tree of artifacts and their
     # transitive dependencies in a JSON file. We use that as the source of truth
     # to generate the repository's BUILD file.
-    #
-    # Coursier generates duplicate artifacts sometimes. Deduplicate them using
-    # the file name value as the key.
     dep_tree = make_coursier_dep_tree(
         repository_ctx,
         artifacts,


MODULE.bazel
```diff
@@ -1019,6 +1028,10 @@ use_repo(
     "policy_pinned_testing",
     "unpinned_policy_pinned_testing",
 
+    # Pinned repo
+    "transitive_dependency_with_type_of_pom",
+    "unpinned_transitive_dependency_with_type_of_pom",
```
shs96c: Nit: we no longer need the `unpinned_` variant since `//transitive_dependency_with_type_of_pom:pin` will work

private/rules/coursier.bzl
```diff
@@ -985,18 +986,35 @@ def make_coursier_dep_tree(
...
+    #    coursier doesn't consistently include pom in the string, for example it
+    #    will return dependencies both like
+    #    "org.javamoney:moneta:pom:sources:1.4.4" and
+    #    "org.apache.logging.log4j:log4j:jar:sources:3.0.0-beta3".
+
     amended_deps = []
     for dep in dep_tree["dependencies"]:
         if not dep.get("file", None):
-            amended_deps.append(dep)
+            if get_classifier(dep["coord"]) == "sources":
```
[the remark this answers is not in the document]
shs96c: I ask because the javadocs jars follow the same pattern as the source jars when we try and download them. I guess we can wait until someone files an issue....

MODULE.bazel
shs96c: I should go through and remove more, but I guess if we add this it'll be one more line :)

shs96c (approved):
LGTM! Let's land this thing :)

tobyh-canva:
Hi @shs96c, do you know if this will be released soon? We're evaluating whether we should upgrade to 6.9 and add this patch, or wait for the 6.10 release. Thanks :)

PR #1481, opened 2025-11-18
Support stripping protobuf-java jars in java_export when using protoc via toolchain

### LINKED ISSUE #1480: protobuf-java runtime is not removed from java_export targets when using aspect-build/toolchains_protoc
## To reproduce:

Set up a Bazel repository according to the https://github.com/aspect-build/toolchains_protoc/ README.
Add a RJE `java_export` target that depends on a `java_protobuf_library`

### Expected: 
com/google/protobuf classes are excluded from jar

### Actual:
com/google/protobuf classes are included in jar

## Additional details
`java_export` appears to have a hardcoded list of workplaces to exclude that normally include protobuf-java, defaulting to `com_google_protobuf` and `protobuf`. However, when using `--incompatible_enable_proto_toolchain_resolution`, the Java protobuf runtime can come from any repo. I have a patch I am working on that will add an optional dependency on the Java Protobuf toolchain in order to obtain the runtime jars and remove them.

## Related issues

https://github.com/bazel-contrib/rules_jvm_external/issues/1377

thirtyseven:
Fixes #1480.

This PR should have tests, but I'm not sure of the best way to add them since ideally we want to run tests/integration/java_export in both toolchain and non-toolchain mode (with and without `--incompatible_enable_proto_toolchain_resolution`) and this doesn't seem easy with the current test suite design. Would like validation about this approach before committing to adding support for this to the test suite.

### DIFF
--- a/private/rules/maven_project_jar.bzl
+++ b/private/rules/maven_project_jar.bzl
@@ -1,3 +1,4 @@
+load("@bazel_skylib//lib:new_sets.bzl", "sets")
 load("@rules_java//java:defs.bzl", "JavaInfo", "java_common")
 load("@rules_license//rules:providers.bzl", "PackageInfo")
 load("//private/lib:bzlmod.bzl", "get_module_name_of_owner_of_repo")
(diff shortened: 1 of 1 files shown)


private/rules/maven_project_jar.bzl
```diff
@@ -252,5 +270,8 @@ single artifact that other teams can download and use.
             default = "@bazel_tools//tools/jdk:current_java_toolchain",
         ),
     },
-    toolchains = ["@bazel_tools//tools/jdk:toolchain_type"],
+    toolchains = [
+        "@bazel_tools//tools/jdk:toolchain_type",
+        config_common.toolchain_type("@rules_java//java/proto:toolchain_type", mandatory = False),
```
shs96c: Since we're dropping Bazel 6 support, I think this is okay.

private/rules/maven_project_jar.bzl
```diff
@@ -60,6 +72,12 @@ def _maven_project_jar_impl(ctx):
         ctx.attr.excluded_workspaces,
     )
 
+    # Strip protobuf runtime
+    artifact_jars = _strip_toolchain_protobuf_runtime(
```
shs96c: If the `java_export` depends on protobuf, we really should add a dependency on `protobuf-java` rather than just stripping it. The question then becomes how to allow the user to set that.

shs96c: Neither do I. Let's leave it for now

PR #1482, opened 2025-11-19
Add slf4j logging implementation for gradle resolver

shs96c:
This prevents a warning being displayed on the console when we start the resolver.


shs96c:
Not needed any more

PR #1488, opened 2025-11-23
Add finegrained control of the visibility of override targets

JonathanPerry651:
This allows us, for example, to override okhttp to point at okhttp_jvm, and provide okhttp as a client-facing api, without having to expose okhttp_jvm directly

### DIFF
--- a/tests/integration/override_targets/verify_visibility.sh
+++ b/tests/integration/override_targets/verify_visibility.sh
@@ -1,10 +1,8 @@
 #!/bin/bash
 set -e
 
-# The genquery output file should contain the target label if it matches the visibility attribute.
-# The file path is passed as the first argument or found in runfiles.
-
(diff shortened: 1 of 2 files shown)


tests/integration/override_targets/verify_visibility.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+set -e
+
+# The genquery output file should contain the target label if it matches the visibility attribute.
+# The file path is passed as the first argument or found in runfiles.
+
+QUERY_OUTPUT=$(find . -name "verify_visibility_query")
```
shs96c: You can use the `args` attribute of `sh_test` to pass the exact location, which will remove the need to execute `find`.

JonathanPerry651: Thanks for the review - I believe I've fixed the PR @shs96c

PR #1492, opened 2025-11-28
Allow an optional index of dep -> class to be created

shs96c:
This provides information that the java gazelle plugin will use to more accurately determine where to load dependencies from.

Note: the index file will be substantially larger than the lock file since it contains far more data (the name of every single top-level class in every dependency)


tests/custom_maven_install/regression_testing_gradle_index.json
```diff
@@ -0,0 +1,139 @@
+{
+  "__AUTOGENERATED_FILE_DO_NOT_MODIFY_THIS_FILE_MANUALLY": "THERE_IS_NO_DATA_ONLY_ZUUL",
+  "classes": {
```
cheister: Why doesn't this include all of the classes from 
```
"androidx.compose.foundation:foundation-layout:1.5.0-beta01",
"androidx.annotation:annotation:1.6.0",
"com.squareup.okhttp3:okhttp:4.12.0",
"com.almworks.sqlite4java:libsqlite4java-linux-i386:1.0.392",
"androidx.fragment:fragment-ktx:1.6.1",
```

cheister: Should we call this `split_package_classes` instead of `classes` then ?

private/tools/java/com/github/bazelbuild/rules_jvm_external/jar/IndexJar.java
```diff
@@ -151,7 +157,31 @@ && isNumericVersion(parts[2])) {
...
   private boolean isNumericVersion(String part) {
     return IS_NUMERIC_VERSION.test(part);
   }
+
+  private boolean isInnerClass(String zipEntryName) {
+    // Inner classes, anonymous classes, and local classes all contain '$' in the class file name
+    // Examples: Outer$Inner.class, Outer$1.class, Outer$1LocalClass.class
+    int lastSlash = zipEntryName.lastIndexOf('/');
+    String fileName = lastSlash == -1 ? zipEntryName : zipEntryName.substring(lastSlash + 1);
+    return fileName.contains("$");
```
honnix, line 185: I don't know the exactly need here, but just want to point out that there are cases like "org.immutables.value.internal.$guava$.annotations" so they are not really inner classes but a relocated packages. We noticed this in our maven_install.json file change after upgrading to 6.10.

honnix: There are also cases like:

```
  org/jooq/meta/firebird/rdb/tables/Rdb$checkConstraints.class
  org/jooq/meta/firebird/rdb/tables/Rdb$fields.class
  org/jooq/meta/firebird/rdb/tables/Rdb$functions.class
  ...

  These aren't actually inner classes — they're top-level jOOQ-generated classes that use $ in the class name to mirror the Firebird system table names (like RDB$CHECK_CONSTRAINTS). But the naive fileName.contains("$") check treats them all as inner classes. Since every
  class in the tables package gets skipped, no package entry is ever added for org.jooq.meta.firebird.rdb.tables.
```

PR #1493, opened 2025-12-02
rules_java: add rules loads to fix autoload on Bazel 9

gergelyfabian:
Autoload is disabled for rules_java by default on Bazel 9. Many of these were already fixed in #1271, but some more are still left.


gergelyfabian:
Removing changes in examples in hope of getting rid of CI failures.

PR #1499, opened 2025-12-07
fix: add string attributes to `amend_artifact` for explicit unset state

JonathanPerry651:
This is a bit foul, but it's the only way I could see of actually passing through an explicit False to any of these attrs.  I think I need this because junit is coming into my repo as testonly, and I want to amend that to False (I have some other deps that directly depend on junit).
If this is the wrong approach for whatever reason, I'm super-open to pushback!

Thank you


JonathanPerry651:
will do thanks

shs96c (approved):
LGTM. Thank you!

shs96c:
The test failures look legitimate to me....

shs96c:
No problem!

PR #1502, opened 2025-12-16
Fix build with Bazel 9

meteorcloudy:
Fix generated via running `buildifier -lint=fix -r ./`

See also https://github.com/bazelbuild/bazel-central-registry/issues/6402


shs96c:
The test failures look genuine.

PR #1503, opened 2025-12-16
fix(pom_file): exclusions referenced before assignment

YavorPaunov:
Under specific circumstances, the exclusions variable inside _pom_file_impl ends up being referenced before assignment. Moving it outside the if block fixes this.

More specifically, this happens when a maven_export has the target parameter set (to point to a jar file) instead of a lib_name. I have also added a test case that triggers this to prevent regression.


shs96c (approved):
LGTM. Nice catch. Thank you

Issue #1517, opened 2026-01-14
The class index file is not supported with the "coursier" resolver


jeffmace:
The new class index file added by #1492 does not work with `resolver = coursier`. It appears that [this condition](https://github.com/bazel-contrib/rules_jvm_external/blob/master/private/rules/coursier.bzl#L799-L801) forces the unpinned script to be used without accounting for the `index_file` attribute.

PR #1519, opened 2026-01-15
Update the maven and coursier resolver tests to create a class index file.

jeffmace:
Validation of the index file for coursier is disabled until #1517 is resolved.


shs96c (approved):
LGTM. Thank you adding the tests.

PR #1523, opened 2026-01-16
Only allow modules specified in known_contributing_modules to contribute artifacts or boms to the root module

cheister:
Instead of allowing any bazel_dep to contribute dependencies to a maven repository just by using the rule, this PR will change the known_contributing_modules to be the allow list of contributing modules for the repo.  If known_contributing_modules is not set then all modules can contribute to the module.  Once it is set then only the modules in the known_contributing_modules will be allowed to contribute artifacts or boms.


shs96c (approved):
LGTM

PR #1528, opened 2026-01-21
Add truststore attrs to maven.install

contagnas:
This change adds first‑class truststore configuration to maven.install.

This is duplicative with/less general than the existing `build --repo_env=JDK_JAVA_OPTIONS=-Djavax.net.ssl.trustStore=<path-to-cacerts>` pattern but as far as I can tell there is not a good way to get that to work when the cacert file is in the repo (or built by the repo) and needs to be a relative path.

A simpler alternative to this would be to expand `%workspace%` variables when handling JDK_JAVA_OPTIONS but this seems more robust and specific.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1058,6 +1058,19 @@ build --repo_env=JDK_JAVA_OPTIONS=-Djavax.net.ssl.trustStore=<path-to-cacerts>
 ```
 can be added to your .bazelrc file if you need to specify custom cacerts for artifact resolution.
 
+If you are using Bzlmod, you can also set truststore-related JVM options directly on `maven.install`:
+
+```python
+maven.install(
+    # ...
+    truststore = "//path/to:cacerts",
+    truststore_password = "//path/to:truststore_password",  # file contains the password
+    truststore_type = "PKCS12",
+)
+```
+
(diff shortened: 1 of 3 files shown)


shs96c:
Why not set the system JDK via the `--server_javabase` option to one that has your trust store already configured? You could also do something like `startup --host_jvm_args=-Djavax.net.ssl.trustStore=/usr/lib/security/cacerts` too. I'm not sure this PR is required. Have I missed something obvious?

contagnas:
> Why not set the system JDK via the `--server_javabase` option to one that has your trust store already configured? You could also do something like `startup --host_jvm_args=-Djavax.net.ssl.trustStore=/usr/lib/security/cacerts` too. I'm not sure this PR is required. Have I missed something obvious?

We have cacerts checked into the repo; this allows referring to that file rather than assuming they'll be present on the machine. 

This also allows (I think, not sure about ordering) using a cacert _built_ in the repo, e.g. with https://github.com/bazel-contrib/rules_distroless/blob/f553dfe82987b5d592bb83f5a6da8f11a51e66fc/distroless/private/java_keystore.bzl#L43

PR #1531, opened 2026-01-22
fix: Do not add coursier opts when run other tools

honnix:
`_generate_java_jar_command` is used in a few other places than calling `coursier`, so adding `COURSIER_OPTS` doesn't make sense.


shs96c (approved):
LGTM. Thank you for the PR

PR #1533, opened 2026-01-22
Skip target generation for artifacts with null shasum

brocollie08:
This is to address issue https://github.com/bazel-contrib/rules_jvm_external/issues/1532


shs96c (changes requested):
I think that this needs a test to prevent the behaviour reverting as we continue to make changes. Could you please add one?

brocollie08:
Could I please get a re-review on this @shs96c

shs96c:
This was superseded by #1591 . Thank you for the PR, and my apologies for not reviewing this sooner.

PR #1543, opened 2026-02-11
Allow boms to be defined in the gradle version catalogues

shs96c:


stevebarrau (approved):
Great improvement. Looking fwd to use this.

cheister (approved):

PR #1548, opened 2026-02-25
Upload primary artifact before uploading hashes in MavenPublisher

thirtyseven:
On rules_jvm_external 6.10, we are continuing to see issues related to publishing artifacts to Artifactory similar to #1544. This PR sequences the uploads in MavenPublisher to ensure that the primary artifact is finished uploaded before its hash files are uploaded. This seems to be needed in order to prevent `Target file to set checksum on doesn't exist` errors.


shs96c (approved):
LGTM

PR #1550, opened 2026-02-27
Fix GitHub search query in README

collinfunk:
The `filename:` query was a part of a legacy implementation of code search. The current equivalent is `path:` [1].

[1] https://github.com/github/docs/issues/24666


shs96c (approved):
Thank you for this! LGTM

Issue #1554, opened 2026-03-03
artifact exclusions do not exclude the artifact in dependencies


escardin:
When adding an artifact as an exclusion, either via `maven.artifact` or `maven.amend_artifact`, exclusions removes the artifact from transitive deps (for runtime), and lists it as a tag so that the `java_export` rule ensures it's listed as an exclusion for the resulting pom.

What it doesn't do, is exclude the artifact as a dependency in the generated `jvm_import` target. This can result in undesirable states when an external dependency graph interacts poorly with bazel.

For example:
```starlark
maven.install(
 boms = [
"com.squareup.misk:misk-bom:2025.11.26.015301-16eeeaa"
]
)
maven.amend_artifact(
    testonly = "true",
    coordinates = "com.squareup.misk:misk-testing",
)

maven.amend_artifact(
    coordinates = "com.squareup.misk:misk-jdbc",
    exclusions = [
        "com.squareup.misk:misk-testing",
    ],
)
```

In this example, misk-jdbc mistakenly depends on misk-testing. misk-testing NEEDS to be marked as testonly, as it rightly depends on other testing dependencies, and should only be used for test targets.

misk-jdbc is needed for non-test targets, and needs a way to exclude misk-testing from its deps, otherwise bazel will refuse to build any target depending on misk-jdbc because it's not also marked testonly.

This can be worked around with an override, but that can be particularly brittle as you now need to keep the override in sync minus the one small change.

I believe that the excluded dependency should actually be excluded from the deps. This is consistent with what one might expect when listing an exclusion, and ensures actual consistent behaviour throughout bazel if the dependency is actually needed but not provided.

PR #1555, opened 2026-03-03
Exclusions listed in the artifact definition should be excluded in deps #1554

escardin:
This should address excluding deps as one might actually expect to happen when listing an exclusion.

I am somewhat worried that it'll break folks, but realistically you would need to allow transitives of your exclusions only as a compile time dependency? I lack imagination for how this could break someone.

Note that this doesn't remove whatever was done previously to exclude dependencies at runtime, but that code should no longer be necessary.

It's not tested against maven or gradle resolvers. I tested it with my own internal repo where I was encountering this bug where I use the coursier resolver.

### DIFF
--- a/private/dependency_tree_parser.bzl
+++ b/private/dependency_tree_parser.bzl
@@ -197,6 +197,13 @@ copy_file(
            get_classifier(dep) in PLATFORM_CLASSIFIER:
             continue
 
+        # If the artifact is in exclusions, exclude it from deps here. It will be listed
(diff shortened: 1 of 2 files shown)


shs96c:
My apologies for not getting to this sooner. I think that #1594 supersedes this, so I'm closing it. Please feel free to let me know if I am mistaken, and thank you for taking the time to prepare a PR for this project! Even if we're slow, we appreciate it.

Issue #1558, opened 2026-03-20
Make rules_jvm_external shell-free


agluszak:
This will improve Windows compatibility (i.e. https://github.com/bazel-contrib/rules_jvm_external/issues/324)

shs96c:
We're trying to avoid compiling java code when pinning as there's no way to use the hermetic java toolchains in a repository rule. If we dropped support for build-time resolutions this would be a significantly easier problem to solve.

Is the current `pin` script problematic on Windows? We've not had reports, but lack of evidence is not necessarily evidence of a lack of an issue.

PR #1559, opened 2026-03-20
Replace pin.sh with a Java pin tool

agluszak:
Work towards #1558


shs96c:
I think the test failures here point to a problem with the approach being taken....

PR #1563, opened 2026-04-07
Upload checksums only after artifact succeeds

rkhachatryan:
CodeArtifact [returns](https://docs.aws.amazon.com/codeartifact/latest/ug/maven-checksums.html#maven-checksum-mismatch) HTTP/404 when uploading checksums for artifacts that haven't been indexed yet. Change the upload flow to wait for the artifact upload to complete before uploading checksums (MD5, SHA1, SHA256, SHA512), ensuring the artifact is available before its checksums are posted.

Checksums are still uploaded in parallel with each other after the artifact succeeds.

### DIFF
--- a/tests/com/github/bazelbuild/rules_jvm_external/maven/MavenPublisherTest.java
+++ b/tests/com/github/bazelbuild/rules_jvm_external/maven/MavenPublisherTest.java
@@ -11,6 +11,8 @@ import java.net.InetSocketAddress;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.util.HashSet;
+import java.util.Set;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import org.junit.Test;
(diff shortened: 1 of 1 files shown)


rkhachatryan:
Hi @jin, @shs96c, @cheister 

I hit this problem when using Bazel + `java_export` from `rules_jvm_external` + CodeArtifact: when uploading checksums, CodeArtifact returns  HTTP/404 for the actual artifact. I think this is caused by parallel uploading of jar and the checksums for it.
I've worked around it temporarily by setting `RJE_MAX_THREADS` to `1` and I think this PR should solve it properly.

Can you please take a look?

(test failures are not related to my changes as far as I can tell - they are also present on other branches)

shs96c (approved):
Nice catch, and what a pain to have had to find out the hard way about. LGTM. Thank you!

shs96c:
You may need to rebase on top of `HEAD` of `master` in order for the Windows CI to pass. My apologies.

rkhachatryan:
Thanks for the review!
I see this issue was already fixed by #1548.
I've updated my PR to keep only the test - if it's needed, otherwise I'd just close it.

IndiTheo:
Oh, it's not only GitLab then. Was about to open a similar issue.

shs96c:
Thank you!

PR #1568, opened 2026-04-23
Prefer getenv to env.get

Andrius-B:
I was having an issue with the value of `REPIN` being cached between invocations and I think using the `rctx/mctx.getenv` is preferrable because it explicitly watches that environment variable, so in case the value of the environment variable changes or it's removed from the client env, the repository gets re-fetched.

https://bazel.build/rules/lib/builtins/repository_ctx#getenv
https://bazel.build/rules/lib/builtins/module_ctx#getenv


shs96c (commented):
LGTM. Thank you for this!

shs96c (approved):

PR #1570, opened 2026-05-07
Track BOMs that manage versionless maven artifacts in the lock file

jeffmace:


jeffmace:
This was opened by accident. There are some more changes still needed along with rebasing the work against top-of-tree.

PR #1574, opened 2026-05-19
Fix MavenPublisher to support POM-only publications

titusfortner:
- Guard the main-artifact upload in `MavenPublisher.run` with `!mainArtifactPath.isEmpty()`, mirroring the `Strings.isNullOrEmpty(extraArtifacts)` check immediately below it.
  - Add `testPublishLocalPomOnly` covering the POM-only publish path.

maven_publish.bzl declares artifact as optional (line 125-127) and encodes "no artifact" as an empty string when generating the publisher script (line 43). maven_bom relies on this — both <name>.publish and <name>-dependencies.publish invoke maven_publish with only a pom (maven_bom.bzl:215, :238), since BOMs are <packaging>pom</packaging> with no jar.

Paths.get("") resolves to the working directory, and Files.readAllBytes(...) on a directory throws java.io.IOException: Is a directory, so every maven_bom publish failed.


shs96c (approved):
LGTM. Thank you!

Issue #1467, opened 2025-10-14
Repositories order and performance regression in 6.8


al-babych-fivetran:
There is an issue with the order of repositories during lock file generation that was introduced in 6.8.
I created a simple repo with an example https://github.com/al-babych-fivetran/rules_jvm_external_issue/commits/main/.

It has two commits (`6.7`, `6.8`).

The `maven.install` configuration has several repositories:
```
repositories = [
      "https://repo1.maven.org/maven2/",
      "https://aaaa.com/maven/",
      "https://cccc.com/maven/",
      "https://bbbb.com/maven/",
      "https://dddd.com/maven/",
  ],
```
The lock file was generated by the `bazel run @maven//:pin` command.
In version 6.7, the order of repositories in the lock file is exactly the same as in the `repositories` option:

<img width="500" height="268" alt="Image" src="https://github.com/user-attachments/assets/9cc33d81-bbef-4ed2-9104-5ace80d2dc64" />

But in version 6.8, they have alphabetical order:

<img width="466" height="260" alt="Image" src="https://github.com/user-attachments/assets/38ad407c-5f9f-4d8b-ac03-f1e81932428b" />

The build tries to download the packages from these sources first.
For this tiny repo, the building with version 6.7 takes ~43s:
```
INFO: Found 1 target...
Target //:example up-to-date:
  bazel-bin/example
  bazel-bin/example.jar
INFO: Elapsed time: 43.344s, Critical Path: 31.71s
INFO: 1252 processes: 9 internal, 1239 darwin-sandbox, 4 worker.
INFO: Build completed successfully, 1252 total actions
```
While the build for version 6.8 has a lot of warnings and took ~127s:
```
...
WARNING: Download from https://bbbb.com/maven/com/amazonaws/aws-java-sdk-accessanalyzer/1.12.732/aws-java-sdk-accessanalyzer-1.12.732-sources.jar failed: class java.io.FileNotFoundException GET returned 404 Not Found
WARNING: Download from https://dddd.com/maven/com/amazonaws/aws-java-sdk-appstream/1.12.732/aws-java-sdk-appstream-1.12.732.jar failed: class com.google.devtools.build.lib.bazel.repository.downloader.UnrecoverableHttpException Unknown host: dddd.com
[...]

PR #1582, opened 2026-06-02
Verify Coursier resolver per-artifact repo membership via HEAD probes

### LINKED ISSUE #349: ability to silence warnings when using multiple repositories

vinnybod:
# Coursier resolver: accurate per-artifact repository membership

Fixes 404-warning noise during `bazel build` when using the Coursier resolver with multiple Maven repositories. Closes #349.

## Problem

The Coursier path through `LockFileConverter` records every artifact as available in *every* configured repository. At fetch time `http_file` tries each URL in order and emits

```
WARNING: Download from https://maven-central-mirror.example.com/.../internal-only-1.0.0.jar failed: GET returned 404 Not Found
```

for every repo that doesn't actually hold the artifact. A real-world monorepo with ~1981 artifacts split across 3 disjoint registries (Maven Central mirror, internal-only artifacts, public artifacts) sees these distributed as `{3: 1981}` in the v3 lock file — every artifact listed against all 3 repos. Each build emits thousands of these warnings.

The Maven and Gradle resolver paths don't have this bug. `Downloader.performDownload()` HEAD-probes every configured repository after the verified primary download and records only those that respond positively.

### Root cause

`coursier.bzl` synthesises a `mirror_urls` field by interpolating each artifact's path into *every* configured repository URL. `LockFileConverter` then prefix-matches those synthetic URLs against the configured repository list:

```java
for (String mirrorUrl : mirrorUrls) {
  for (URI repo : repositories) {
    if (m2local.equals(repo)) continue;
    if (mirrorUrl.startsWith(repo.toString())) {
      repos.add(repo);
    }
  }
}
```

By construction every synthetic URL matches a configured repo, so every artifact gets every repo.

## Fix

Replace the synthetic match in `LockFileConverter.getDependencies()` with two unconditional steps:

1. Record the verified primary repo from Coursier's `url` field.
2. HEAD-probe every other configured repository and record those that respond 2xx.

Same shape as the Maven/Gradle paths. After this change all three resolvers behave the same way and produce equally accurate v3 lock files.

### Parallelism

N-1 HEAD probes per artifact for a ~2000-artifact lock file would serially exceed the default 600s `resolve_timeout`. Per-dep work moves into `toDependencyInfo()` and runs on a fixed-size `ExecutorService` sized by:

- `RJE_MAX_THREADS` env var, or
- `--max-threads` CLI flag,
- defaulting to `max(8, min(32, processors*4))`.

Default higher than the Maven/Gradle path's `min(5, processors)` because HEAD probes are I/O-bound, much cheaper than full downloads, and ~2000 of them must fit inside the 600s `resolve_timeout`.

Note: `RJE_MAX_THREADS` is also read by `MavenPublisher` (default 8 there). Operators tuning one would tune the other. Happy to rename to `RJE_LOCK_CONVERTER_THREADS` if maintainers prefer decoupling.

### Why a new HeadProber instead of reusing HttpDownloader.head?

`LockFileConverter` runs at `repository_rule` time, where `coursier.bzl` invokes the deploy jar via Bazel's embedded jlink JDK (`bazel_tools/jdk`). That JDK is jlink'd to a minimal module set that does **not** include `java.net.http`, so `HttpDownloader.head()` would `NoClassDefFoundError` if loaded there. (The Maven/Gradle paths reuse it because they run at build time with the full configured toolchain.)

A ~200-line `HeadProber` (`Predicate<URI>`) built on `java.net.HttpURLConnection` (`java.base`) handles HTTP/HTTPS and file URIs and reads `~/.netrc` for Basic auth. Tradeoffs vs `HttpDownloader.head`: no 5xx retry, no bearer-token auth — a transient failure drops the mirror from this run's lockfile and is recovered by re-pinning.

### Diagnostic warning

If Coursier's primary `url` doesn't prefix-match any configured `--repo` (stale entry, trailing-slash drift, `/maven` vs `/maven2/`), the lockfile would otherwise record the artifact with no repos and fail at fetch time with no breadcrumb. `LockFileConverter` now warns once per primary host in this case.

## Related

- #349 — the issue this addresses
- #743 — earlier v1 lockfile fix (didn't carry to v3)
- #1447 — earlier attempt
- #1467 — v3 lock file resurfacing
- [confluentinc/rules_jvm_external#81](https://github.com/confluentinc/rules_jvm_external/pull/81) — fork PR where these changes were validated end-to-end against a real-world Maven monorepo

---

Edit: this still needs some edits. The probing ends up getting triggered at build time instead of just at lock time.


shs96c:
I know the AI came up with a plausible reason why you couldn't use the existing classes, but it's not right. Please use the existing infrastructure for this.

vinnybod:
I went down a few paths on this before landing where I did, but I recognize its still not quite a good solution.

By "existing infrastructure" I assume you mean in `coursier.bzl`. This does work. But, the problem I was running into is the unpinned `coursier_fetch` repo rule was still being executed and causing 2000+ 404 warnings (from the probe calls) at build-time.

The Java solution in this draft PR still suffers from the same side-effect. It is only *slightly* better UX in that we can at least suppress the warnings at the expense of more code. But it still hits a timeout and fails the build.

I need to spend a little time reproducing the "unpinned code path at build time" issue. It seems to be a combination of having a `rules_multirun` target that depends on `@maven//:pin` in combination with `bazel-diff`. If I eliminate that, and we stop triggering the "unpinned" code bath on builds, I can reimplement this in our fork in `coursier.bzl` and test it a bit more before re-opening upstream.

---
As a side note, I think the "easy" thing is probably just to try to use the gradle resolver since it doesn’t run as a repo rule or have the same 404 issue, but unfortunately it seems to shift my dependencies too much. I haven't dug into it yet, but it seems like the gradle resolver is ignoring the "pinned" attribute on `install`.

PR #1584, opened 2026-06-09
Enable RJE_UNSAFE_CACHE by default

shs96c:
The shared cache is now used unless RJE_UNSAFE_CACHE is set to "0" or "false". The CLI flag --use_unsafe_shared_cache has been replaced by --use_unique_cache, which opts out of the shared cache.

Maven and Gradle docs updated accordingly. Coursier behaviour is unchanged for now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1032,11 +1032,11 @@ variables are honoured:
 | Environment variable | Meaning                                                                                                                                                        |
 |----------------------|----------------------------------------------------------------------------------------------------------------------------------------------------------------|
 | `RJE_ASSUME_PRESENT` | Prevents the resolver from checking remote repositories to see if a dependency is present, and just assumes it is                                              |
-| `RJE_UNSAFE_CACHE`   | Uses your `$HOME/.m2/repository` directory to speed up dependency resolution. Enabled by default; set to `0` or `false` to disable.                            |
+| `RJE_UNSAFE_CACHE`   | By default, the shared `$HOME/.m2/repository` directory is used to speed up dependency resolution. Set to `0` or `false` to use an isolated cache instead.     |
 
-The unsafe cache option uses your local `$HOME/.m2/repository` as
-a source for dependency resolutions, but will not include any local paths in
-the generated lock file unless the `repositories` attribute contains `m2local`.
+Although your local `$HOME/.m2/repository` is used as a source for
+dependency resolutions, no local paths are included in the generated
+lock file unless the `repositories` attribute contains `m2local`.
 
 The Maven-backed resolver will use credentials stored in a `$HOME/.netrc`
 file when performing dependency resolution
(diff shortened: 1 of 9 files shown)


cheister:
Why are we changing the default to use something that was previously deemed unsafe?

shs96c:
TL;DR: we're making the gradle and maven resolvers act like `coursier` does.

The `coursier` resolver hasn't been paying attention to this for a long time. On macOS, `coursier` caches files in `~/Library/Caches/Coursier/v1`, and because pinning happens via `bazel run` the user's env variables are used. If you've ever needed to repin a repo with a lot of dependencies more than once on the same machine, you'll find it happens suspiciously quickly. That's why.

The `gradle` resolver delegates down to Gradle, which will correctly populate the gradle cache, so it's a safe operation to do there. Similarly, the `maven` resolver uses Maven's own classes for this: it downloads the POMs to the right places. 

We recently landed changes that attempt to use a stable gradle home between invocations (#1473) Because that directory was stable, even though we weren't using the user's cache, we were using the same cache repeatedly.

shs96c:
The reason we viewed the repository cache as "unsafe" is because other tools could (and would) write to and modify it. I don't think that's the right call any more, otherwise a lot of maven and gradle developers would be having a very bad time of it.

README.md
```diff
@@ -1032,9 +1032,9 @@ variables are honoured:
 | Environment variable | Meaning                                                                                                                                                        |
 |----------------------|----------------------------------------------------------------------------------------------------------------------------------------------------------------|
 | `RJE_ASSUME_PRESENT` | Prevents the resolver from checking remote repositories to see if a dependency is present, and just assumes it is                                              |
-| `RJE_UNSAFE_CACHE`   | When set to `1` will use your `$HOME/.m2/repository` directory to speed up dependency resolution                                                               |
+| `RJE_UNSAFE_CACHE`   | Uses your `$HOME/.m2/repository` directory to speed up dependency resolution. Enabled by default; set to `0` or `false` to disable.                            |
```
cheister: Can we re-word this to something like.  The default is to use the shared default directory `$HOME/.m2/repository` for cached artifacts.  If you want to use an isolated cache directory for RJE resolution set to `0` or `false`

private/tools/java/com/github/bazelbuild/rules_jvm_external/resolver/cmd/ResolverConfig.java
```diff
@@ -122,8 +123,8 @@ public ResolverConfig(EventListener listener, String... args) throws IOException
           maxThreads = Integer.parseInt(args[i]);
           break;
 
-        case "--use_unsafe_shared_cache":
-          request.useUnsafeSharedCache(true);
+        case "--use_unique_cache":
```
cheister: maybe use_isolated_cache ?  Not sure what unique means here

private/rules/coursier.bzl
```diff
@@ -1654,6 +1654,12 @@ pinned_coursier_fetch = repository_rule(
         # Use @@// to refer to the main repo with Bzlmod.
         "_workspace_label": attr.label(default = ("@@" if str(Label("//:invalid")).startswith("@@") else "@") + "//does/not:exist"),
     },
+    environ = [
+        "HOME",
```
cheister, line 1658: I'm unclear why all of these now need to be added.  Were tasks using these env vars not working before?

cheister (approved):

private/rules/coursier.bzl
shs96c: Modern Bazel's will now include these in the rule key, so if they change, we'll properly invalidate the existing outputs. The environment variables listed here are all ones that we refer to in our code already.

README.md
shs96c: Good point. Done. Also updated equivalent gradle comment.

private/tools/java/com/github/bazelbuild/rules_jvm_external/resolver/cmd/ResolverConfig.java
shs96c: Done

PR #1589, opened 2026-06-12
feat: Support pluggable dependency download and metadata caching SPIs

JonathanPerry651:
Provides pluggable SPIs for rules_jvm_external's dependency resolution, allowing custom download/caching and authentication implementations (such as integrating with Bazel's remote fetcher/downloader APIs) to be registered via classpath plugins.

Key Changes:
1. Public API Targets & SPIs:
   - Created the //resolver:api target containing public interfaces for MetadataService, DependencyMetadata, and DownloadService.
   - Added a generic SpiLoader utility to unify and validate classpath SPI service resolution (ensuring at most one implementation is loaded per SPI).
   - Exposed custom classpath library loading via the new resolver_plugin_libraries attribute.

2. Aether & Downloader SPI Integration:
   - Replaced Aether's default HTTP transport with a custom HttpDownloaderTransporterFactory, routing all POM, BOM, and metadata lookups through the pluggable DownloadService SPI.
   - Defined the new DownloadService SPI to return null (on download fail) or false (on head check fail) on error instead of throwing checked exceptions, keeping the integration API simple and clean.

3. Inline URL Credentials Handling:
   - Previously, Aether resolved dependency graphs successfully using inline URL credentials (e.g. http://user:pass@host). However, the Downloader (which fetches the final dependency files using HttpDownloader's JDK HttpClient) did not support inline credentials. This created an inconsistent/broken flow where resolution succeeded but downloading failed.
   - To make the behavior consistent, Aether downloads are now routed through the pluggable DownloadService SPI backed by HttpDownloader without inline credentials support (making both resolution and downloading fail consistently). Users must configure credentials via .netrc instead.
   - Deleted the shouldDownloadOverHttpWithAuthenticationPassedInOnRepoUrl test from ResolverTestBase.java since that flow does not work and is explicitly unsupported.

4. Cleanup & Optimization:
   - Removed redundant metadata caches and unused resolver binaries, replacing them with dynamically generated custom_resolver targets.


private/rules/coursier.bzl
```diff
@@ -860,9 +871,11 @@ def generate_pin_target(repository_ctx, unpinned_pin_target):
             lock_file_location = "/".join([package_path, file_name])  # e.g. path/to/some.json
 
         if repository_ctx.attr.resolver == "maven":
-            resolver_target = Label("//private/tools/java/com/github/bazelbuild/rules_jvm_external/resolver/maven:MavenMain")
+            main_class = "com.github.bazelbuild.rules_jvm_external.resolver.maven.MavenMain"
+            resolver_lib = "@rules_jvm_external//private/tools/java/com/github/bazelbuild/rules_jvm_external/resolver/maven:maven-resolver-lib"
```
shs96c: These should still be labels, even if we coerce to a string. That allows people to use a different repo name than `rules_jvm_external`

private/rules/coursier.bzl
```diff
@@ -876,6 +889,9 @@ def generate_pin_target(repository_ctx, unpinned_pin_target):
                 dependency_index_location = "/".join([dep_index_package_path, dep_index_file_name])
 
         return _IN_REPO_PIN.format(
+            main_class = main_class,
+            resolver_lib = resolver_lib,
```
shs96c: And you can use the `repr(str())` trick here to get it to render properly.

private/rules/maven_install.bzl
```diff
@@ -80,6 +81,7 @@ def maven_install(
         is "warn".
       repin_instructions: Instructions to re-pin dependencies in your repository. Will be shown when re-pinning is required.
       ignore_empty_files: Treat jars that are empty as if they were not found.
+      resolver_plugin_libraries: Jars or libraries containing custom MetadataService SPI implementations to add to the resolver classpath.
```
shs96c: Technically speaking, these are resolver deps, but for your use-case you're using them as plugins. Can we pick a more descriptive attribute name

resolver/src/main/java/com/github/bazelbuild/rules_jvm_external/resolver/remote/SpiLoader.java
```diff
@@ -0,0 +1,60 @@
...
+
+import java.util.Iterator;
+import java.util.ServiceLoader;
+import java.util.function.Consumer;
+
+/**
+ * Utility class to dynamically load and resolve SPI (Service Provider Interface) services.
+ * Ensures that at most one custom service implementation is registered on the classpath.
+ */
+public final class SpiLoader {
```
shs96c: We generally avoid declaring classes as `final`, even if they are utility classes.

resolver/src/main/java/com/github/bazelbuild/rules_jvm_external/resolver/remote/SpiLoader.java
```diff
@@ -0,0 +1,60 @@
...
+  public static <T> T load(Class<T> serviceClass, T defaultService, Consumer<T> initializer) {
+    ServiceLoader<T> loader = ServiceLoader.load(serviceClass);
+    Iterator<T> iterator = loader.iterator();
+    if (iterator.hasNext()) {
+      T service = iterator.next();
+      if (iterator.hasNext()) {
+        StringBuilder foundServices = new StringBuilder("[");
+        foundServices.append(service.getClass().getName());
+        foundServices.append(", ");
+        foundServices.append(iterator.next().getClass().getName());
```
shs96c: You can use `Joiner` from guava to get this done nicely

private/tools/java/com/github/bazelbuild/rules_jvm_external/resolver/maven/HttpDownloaderTransporterFactory.java
```diff
@@ -0,0 +1,148 @@
...
+    String scheme = repository.getProtocol();
+    if ("http".equalsIgnoreCase(scheme)
+        || "https".equalsIgnoreCase(scheme)
+        || "file".equalsIgnoreCase(scheme)) {
+      return new HttpDownloaderTransporter(downloadService, repository);
+    }
+    throw new NoTransporterException(repository);
+  }
+
+  private static class HttpDownloaderTransporter implements Transporter {
```
shs96c: It's fine to have top level classes

private/tools/java/com/github/bazelbuild/rules_jvm_external/resolver/remote/HttpDownloader.java
```diff
@@ -246,4 +246,17 @@ private boolean isSuccessful(HttpResponse<?> response) {
   public void close() throws Exception {
     listener.close();
   }
+
+  @Override
+  public Path download(URI uri) {
```
shs96c: Since we've created this custom interface, why not just call the method `get`?

resolver/src/main/java/com/github/bazelbuild/rules_jvm_external/resolver/remote/DependencyMetadata.java
```diff
@@ -0,0 +1,61 @@
...
+  private final Map<String, ? extends Set<String>> services;
+
+  public DependencyMetadata(
+      String sha256,
+      Set<URI> repositories,
+      Set<String> packages,
+      Set<String> classes,
+      Map<String, ? extends Set<String>> services) {
+    this.sha256 = sha256;
+    this.repositories = repositories;
```
shs96c: Since you're returning these from accessor methods, they should really be a copy of the original parameters. It probably doesn't matter here, since we control the code.

resolver/src/main/java/com/github/bazelbuild/rules_jvm_external/resolver/remote/DownloadService.java
```diff
@@ -0,0 +1,54 @@
...
+ * intercept or override caching and authentication during resolution.
+ */
+public interface DownloadService {
+  /**
+   * Downloads the content of the specified URI to a local path.
+   *
+   * @param uri the URI of the resource to download
+   * @return the local Path where the resource was downloaded, or null if the download fails
+   */
+  Path download(URI uri);
```
shs96c: I'd suggest renaming this to `get` to minimise changes in existing code. That will also name the method after the HTTP verb, in line with the `head` method's naming convention.

resolver/BUILD
```diff
@@ -0,0 +1,27 @@
+load("@rules_java//java:java_library.bzl", "java_library")
+load("@rules_jvm_external//:defs.bzl", "artifact")
+
+java_library(
+    name = "api",
```
shs96c: Prefer build files closer to the classes, rather than a package this high up in the tree. Also, I'm not sure why it says about our package selection that the `api` spans two packages.

shs96c (changes requested):

JonathanPerry651:
@shs96c - thank you for your review!  I think I addressed all your comments.  One slightly nasty bit is that I had to expose the coordinates api (since it forms part of the MetadataService signatures).  I was torn between a refactor, moving Coordinates up into the api package, vs an exports attr.  I decided on the latter as it would be less code diffs, but let me know if you disagree/have a different desired approach.

acecilia:
@JonathanPerry651 Thanks for this! I managed to implement GCS authorization on top of your PR. I opened a draft PR [here](https://github.com/JonathanPerry651/rules_jvm_external/pull/2) just to share the idea.

I will open a PR in upstream `rules_jvm_external` when this PR gets merged

private/rules/coursier.bzl
```diff
@@ -863,9 +874,11 @@ def generate_pin_target(repository_ctx, unpinned_pin_target):
             lock_file_location = "/".join([package_path, file_name])  # e.g. path/to/some.json
 
         if repository_ctx.attr.resolver == "maven":
-            resolver_target = Label("//private/tools/java/com/github/bazelbuild/rules_jvm_external/resolver/maven:MavenMain")
+            main_class = "com.github.bazelbuild.rules_jvm_external.resolver.maven.MavenMain"
+            resolver_lib = Label("@rules_jvm_external//private/tools/java/com/github/bazelbuild/rules_jvm_external/resolver/maven:maven-resolver-lib")
```
shs96c: `Label(//private/tools/java/...)` The `rules_jvm_external` workspace name may not exist.

private/rules/coursier.bzl
```diff
@@ -863,9 +874,11 @@ def generate_pin_target(repository_ctx, unpinned_pin_target):
...
         elif repository_ctx.attr.resolver == "gradle":
-            resolver_target = Label("//private/tools/java/com/github/bazelbuild/rules_jvm_external/resolver/gradle:GradleMain")
+            main_class = "com.github.bazelbuild.rules_jvm_external.resolver.gradle.GradleMain"
+            resolver_lib = Label("@rules_jvm_external//private/tools/java/com/github/bazelbuild/rules_jvm_external/resolver/gradle:gradle-resolver-lib")
```
shs96c: Same here

private/tools/java/com/github/bazelbuild/rules_jvm_external/resolver/cmd/AbstractMain.java
```diff
@@ -179,60 +191,68 @@ private static Set<DependencyInfo> fulfillDependencyInfos(
...
-      return null;
-    }
+  private static MetadataService resolveMetadataService(
+      MetadataService localMetadataService, EventListener listener) {
+    return SpiLoader.load(
+        MetadataService.class,
+        localMetadataService,
+        service -> {
+          listener.onEvent(
+              new PhaseEvent(
```
shs96c: This feels more like a log event

private/tools/java/com/github/bazelbuild/rules_jvm_external/resolver/remote/LocalMetadataService.java
```diff
@@ -0,0 +1,63 @@
+// Copyright 2024 The Bazel Authors. All rights reserved.
```
shs96c: We're in 2026.

resolver/src/main/java/com/github/bazelbuild/rules_jvm_external/resolver/BUILD
```diff
@@ -0,0 +1,58 @@
...
+            "com.google.guava:guava",
+            repository_name = "rules_jvm_external_deps",
+        ),
+    ],
+    exports = [
+        "//private/tools/java/com/github/bazelbuild/rules_jvm_external",
+    ],
+)
+
+java_library(
```
shs96c: Why so many targets? it seems like you can merge everything into a single target?

shs96c (changes requested):

shs96c (approved):
LGTM. Thank you!

PR #1596, opened 2026-06-25
Don't deduplicate additional_coursier_options

Dolfik1:
`additional_coursier_options` is passed straight to coursier as an ordered list of command-line arguments, but when merging installs across modules we were running it through the same dedup logic as repositories and netrc lines. That collapsed repeated tokens and broke coursier's repeatable flags.

For example this install:
```
    maven.install(
        additional_coursier_options = [
            "--enable-modules",
            "--variant", "org.gradle.category=library",
            "--variant", "org.gradle.usage=runtime",
        ],
    )
```
ended up with only a single `--variant`, because the duplicate `--variant` tokens were removed. The repeatable `--variant` flag was added in newer versions of coursier, so this only shows up once you're on a recent enough version.

This change concatenates the options (root first) instead of deduplicating them. The other list attributes still dedup as before. One side effect is that if the same option is declared by both a root and a non-root module it now appears twice, which is fine since these are positional arguments and we can't safely dedup them.

Pulled the concat into a small helper and added unit tests covering the repeated --variant case, ordering, and cross-module duplicates.

### DIFF
--- a/private/extensions/maven.bzl
+++ b/private/extensions/maven.bzl
@@ -578,6 +578,16 @@ def _merge_repo_lists(root_list, non_root_list):
             seen.append(item)
     return merged_list
 
+def concat_coursier_options(root_list, non_root_list):
+    """Concatenate coursier options (root first), never deduplicating.
+
+    Unlike the other repo list attributes, `additional_coursier_options` is an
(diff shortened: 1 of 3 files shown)


shs96c (approved):
LGTM. Thank you!

PR #1597, opened 2026-06-29
Skip package-info.class when indexing jars

shs96c:


cheister (approved):

cheister:
Need to fix the build

PR #1600, opened 2026-06-29
Fix TOML AAR overrides using original artifact copy targets

### LINKED ISSUE #976: Support providing maven coordinates using libs.toml files
It would be great to support passing `libs.toml` files containing maven coordinates into `rules_jvm_external`.

I see a few benefits:
- Would help improve dependency management and let tools like `dependabot` and `renovate` help manage automatically bumping version updates.
- Would reduce the need to clutter our `MODULE.bazel` with giant lists of maven dependencies.

Potential issues:
- Conveying attributes like `testonly`, `neverlink`, and `packaging (aar|jar|...)` might be harder. Technically this is supported by the Toml format already but we would need to document this.

### Example:
```starlark
# MODULE.bazel
maven = use_extension("@rules_jvm_external//:extensions.bzl", "maven")
maven.install(
    # Toml file containing all of the maven coords
    toml = "//:deps.toml",
)
```

```toml
# deps.toml
[versions]
androix-activity-version = "1.8.0"
androidx-experiemental-version = "1.3.0"
...

[libraries]
androidx-activity-compose = { module = "androidx.activity:activity-compose", package = "aar", version.ref = "androix-activity-version" }
androidx-activity-ktx = { module = "androidx.activity:activity-ktx", package = "aar", version.ref = "androix-activity-version" }
...
```

Bencodes:
When an artifact is overridden, the generated `original_` target should keep using the real resolved artifact repository as its backing file source. The copy rule derived the source repo from the synthetic `original_` coordinates, which created references to non-existent repositories for AAR artifacts parsed from `libs.versions.toml`.

Fixes https://github.com/bazel-contrib/rules_jvm_external/issues/976 and https://github.com/bazel-contrib/rules_jvm_external/issues/1421

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -307,6 +307,11 @@ dev_maven.from_toml(
     ],
     libs_versions_toml = "//tests/integration:libs.versions.toml",
 )
+dev_maven.override(
+    name = "from_files",
+    coordinates = "com.google.android.gms:play-services-tasks",
+    target = "@//tests/integration/override_targets:play_services_tasks_override",
+)
 dev_maven.amend_artifact(
     name = "from_files",
     testonly = "true",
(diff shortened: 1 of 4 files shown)


shs96c (approved):
LGTM. Nice catch, and thank you

PR #1601, opened 2026-07-05
Pass COURSIER_CACHE to coursier when fetching unpinned repositories

vadikmironov:
Coursier derives its default cache location from the JVM's `user.home`, which on Linux comes from the passwd database and ignores `$HOME`, while these rules derive the shared cache location from `$HOME` (or `$XDG_CACHE_HOME`). When the two disagree — e.g. `$HOME` remapped to a network-mounted home directory — coursier bypasses the shared cache during a repin, re-downloads artifacts into the passwd home, and the fetch then fails with:

```
Error while trying to parse the path of file in the coursier cache: /home/<user>/.cache/coursier/v1/...
```

**Repro:** on Linux, run `REPIN=1 bazel run @unpinned_maven//:pin` with `HOME` pointing anywhere other than the passwd home dir (`env HOME=/some/other/dir ... bazel run @unpinned_maven//:pin`). Before this change the repin hard-fails as above; after it, coursier uses (and populates) the `$HOME`-derived shared cache.

**Fix:** always pass the Starlark-computed cache location to the spawned coursier process via the `COURSIER_CACHE` environment variable, as the pinned flow already does. The logic is extracted into `get_coursier_environment` (public for testing). If none of the variables a shared cache location can be derived from is set (`HOME`/`XDG_CACHE_HOME`/`COURSIER_CACHE`), the environment is left untouched so coursier keeps using its own default.

Behavior note: a `-Dcoursier.cache=...` passed via `COURSIER_OPTS` is now overridden by the forced environment variable in the unpinned flow; that configuration previously failed the relativization step anyway.

A follow-up PR will fix the same `user.home` vs `$HOME` skew in the Java resolver tooling (`.netrc` lookup, local `.m2`/`.gradle` caches).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/private/rules/coursier.bzl
+++ b/private/rules/coursier.bzl
@@ -989,6 +989,34 @@ def get_coursier_cache_or_default(repository_ctx, use_unsafe_shared_cache):
 
     return default_cache_dir
 
+# Compute the environment for the spawned coursier process, forcing the cache
+# location that get_coursier_cache_or_default computed.
+#
+# Coursier itself would otherwise derive its default cache location from the
+# JVM's `user.home` property, which on Linux comes from the passwd database and
+# ignores $HOME, while this file derives it from $HOME (or $XDG_CACHE_HOME). If
+# the two disagree (e.g. $HOME points at a network mount), coursier would
+# bypass the shared cache and report artifact paths that cannot be relativized
+# against it. Setting COURSIER_CACHE explicitly keeps both sides in agreement.
+# For repositories that do not use the shared cache, this also prevents
+# coursier from writing into home directories.
+# https://github.com/bazelbuild/rules_jvm_external/issues/301
+# https://github.com/coursier/coursier/blob/1cbbf39b88ee88944a8d892789680cdb15be4714/modules/paths/src/main/java/coursier/paths/CoursierPaths.java#L29-L56
+#
+# This method is public for testing.
+def get_coursier_environment(repository_ctx, use_unsafe_shared_cache):
+    if use_unsafe_shared_cache and not _is_windows(repository_ctx):
+        # Every shared cache location on this platform is derived from one of
+        # these variables. If none of them is usable, forcing COURSIER_CACHE
+        # would point coursier at a nonsensical path; leave the environment
+        # alone and let coursier use its own default instead.
+        os_env = repository_ctx.os.environ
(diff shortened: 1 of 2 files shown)


shs96c:
Why not just set the `COURSIER_CACHE` env var in this? That's the kind of problem it's designed to fix....

If that's not the right answer, why not make [this branch](https://github.com/bazel-contrib/rules_jvm_external/blob/master/private/rules/coursier.bzl#L1088-L1099) unconditional?

vadikmironov:
Thanks a lot for the feedback!

Setting `COURSIER_CACHE` manually does work, and this PR keeps it working (an existing value is read first and passed through). For the setup with mounted home and redefined `HOME` env var the first experience of setting up rules_jvm_external is a bit rough though - one runs `bazel run @unpinned_maven//:pin` and gets back a cryptic stack trace saying "Error while trying to parse the path of file in the coursier cache". Discovering that the fix is to set `COURSIER_CACHE` requires digging through the rules_jvm_external source and finding the `user.home` vs `$HOME` mismatch. The rule already knows where it expects the cache to be and this change just tells the spawned coursier process the same thing.

Making that branch unconditional would pass the repo-local directory for unpinned repos too (defeating the shared cache they are supposed to use). If you meant something like `--cache get_coursier_cache_or_default(ctx, _is_unpinned(ctx))` then this is an option and it would be equivalent to this change. I opted for the env var because `--cache` is coursier's highest-precedence setting - users can still pass their own `--cache` via `additional_coursier_options` and it wins over the env var. Forcing `--cache` ourselves would compete with the user setting, but more importantly the coursier invocation would fail for the unpinned case as coursier rejects a duplicated `--cache` option ("Argument --cache: argument ??? already specified").

vadikmironov:
@shs96c , please do let me know if the explanation make sense? Just to re-iterate - this PR is using `COURSIER_CACHE` env var approach also honoring user input when provided. As mentioned, `--cache` option doesn't look to be working as clean.

vadikmironov:
@shs96c , any chance this would make it into 7.1? It's still a pain on mounted NAS home and while I can make it work, it's a bit of a nuisance to remember to set the right env var.

shs96c (approved):
LGTM. Sincere apologies for making you wait this long, and thank you for your patience.

vadikmironov:
@shs96c , no need to apologise at all. Thanks a lot for the review and for getting this in, much appreciated.

PR #1602, opened 2026-07-06
fix: make :outdated work when runfiles trees are disabled

shs96c:
`bazel run @maven//:outdated` fails with `Error: Unable to access jarfile ...` when runfiles trees are not materialised (`--noenable_runfiles`, the default on Windows), because the generated `sh_binary` passes `$(location)` rootpaths that are resolved relative to the working directory. Resolve the arguments through the Bash runfiles library instead, as `:pin` has done since #900. Verified against a bzlmod workspace consuming rules_jvm_external as an external repo under default flags, `--noenable_runfiles`, and `--experimental_sibling_repository_layout`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/private/outdated.sh
+++ b/private/outdated.sh
@@ -1,12 +1,23 @@
 #!/usr/bin/env bash
 
-set -e -o pipefail
+# --- begin runfiles.bash initialization v3 ---
+# Copy-pasted from the Bazel Bash runfiles library v3.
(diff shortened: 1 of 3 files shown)


escardin:
This fixed my issue!

cheister (approved):

PR #1604, opened 2026-07-09
BOM/POM coordinates condition marked as `skipped` must not be turned into http_file downloads

mikhail-0330:
Artifacts with no `file` (e.g. BOM/POM coordinates marked as `skipped` in the lock file because they have no jar) must not be turned into  `http_file` downloads: there is nothing to fetch and constructing the  `.jar` URL for them results in spurious 404s. 
See the equivalent guard in `pinned_coursier_fetch_impl` (`if artifact.get("file"):`).
(https://github.com/bazel-contrib/rules_jvm_external/blob/master/private/rules/coursier.bzl#L1290)


shs96c (approved):
LGTM

PR #1610, opened 2026-07-20
Use a UTF-8 locale for jvm_import jar actions

dzbarsky:
`jvm_import` invokes `AddJarManifestEntry_deploy.jar` directly with `current_java_runtime`, so the generated `java_binary` launcher no longer supplies a UTF-8 locale. On Linux, a Bazel action without a locale uses an ASCII filesystem encoding. When the execroot or output base contains non-ASCII characters, Java cannot open `AddJarManifestEntry_deploy.jar`; Bazel's `//src/test/shell/bazel:bazel_determinism_test` reports `Unable to access jarfile` or `ClassNotFoundException`.

This matches the working fix in bazelbuild/bazel#30147.


shs96c (approved):
Sure.

PR #1611, opened 2026-07-21
Add support for GCS auth in both `resolution` and `jar download` phases

acecilia:
Add support for GCS auth in both `resolution` and `jar download` phases:

* Resolution phase:
  * Gradle resolver: 
    * The resolver now converts `gcs://` to `https://storage.googleapis.com/`
    * Auth token is obtained using the `google-auth-library-oauth2-http` library. `GcpTokenProvider` wraps the calls to the library
  * Maven resolver: 
    * `GcsDownloadService` SPI plugin uses `google-cloud-storage` library, and delegates non-GCS URIs to the default service. `HttpDownloaderTransporterFactory` now accepts `gcs` scheme
    * Auth token is obtained by the `google-cloud-storage` library

* Jar download phase: 
  * `download_pinned_deps.bzl` converts `gcs://` to `https://` inline and uses Bazel's http_file
  * Auth token is obtained via Bazel --credential_helper (delegated to the client to configure). Example of credential helper [here](https://github.com/tweag/rules_gcs/blob/a5080a97b4cc869ccb2d00281591389c25caff63/tools/credential-helper)

### DIFF
--- a/private/tools/java/com/github/bazelbuild/rules_jvm_external/resolver/gradle/AuthMethod.java
+++ b/private/tools/java/com/github/bazelbuild/rules_jvm_external/resolver/gradle/AuthMethod.java
@@ -1,0 +1,22 @@
+// Copyright 2025 The Bazel Authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.github.bazelbuild.rules_jvm_external.resolver.gradle;
+
+/** Authentication method for a {@link Repository}. */
+public enum AuthMethod {
+  NONE,
+  BASIC,
+  BEARER,
+}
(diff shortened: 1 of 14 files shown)


acecilia:
@shs96c would appreciate some feedback here 🙏

private/tools/java/com/github/bazelbuild/rules_jvm_external/resolver/gradle/AuthMethod.java
```diff
@@ -0,0 +1,22 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.github.bazelbuild.rules_jvm_external.resolver.gradle;
+
+/** Authentication method for a {@link Repository}. */
+public enum AuthMethod {
```
shs96c, line 18: This feels like something that should live a level higher. We can move it in a later PR if needed.

shs96c (changes requested):
Apologies for how long it took to get to this review. Thank you for your patience. 

A couple of open questions and comments:

1. We appear to ignore the Maven resolver entirely here. I think it'd be enough to add the service to the maven resolver binary as a runtime dep.
2. I _think_ `createRepository` will fetch a token for any `gcs:// repo`, so a public bucket that would resolve anonymously via HTTPS now throws `RuntimeException` if ADC isn't configured.
3. There are no docs: would be nice to let people know that this works now
4. As written, the built-in registration makes the resolver unusable for anyone with a custom `DownloadService` registered via resolver_extra_dependencies because `SpiLoader` throws on multiple impls, and there's no way to opt out. Please either wire GCS into the default downloader (which I'd recommend) without the META-INF registration, or extend `SpiLoader` to support composition first (which seems like a big lift).

PR #1612, opened 2026-07-23
fix(publish): Apparently we need a specific version of the publish yaml

shs96c:


vinnybod (commented):
Why not take this a step further and use the commit hash? Immutability/Reproducibility ftw :)

shs96c:
The tag _should_ be immutable, and the version number is easier to read and check to see if we need to update it.

vinnybod (approved):

PR #1615, opened 2026-07-24
fix: make Gradle resolver cache writes race-tolerant

thirtyseven:
## Problem

When pinning a large dependency graph with `resolver = "gradle"`, the pin aborts non-deterministically with `java.nio.file.FileAlreadyExistsException` while copying a freshly-downloaded artifact into the local cache. The failing artifact differs from run to run — whichever one loses a race between parallel downloader threads — so it reproduces on large graphs and not on small ones. Serializing the resolver (`RJE_MAX_THREADS=1`) makes it disappear, confirming a pure concurrency race.

```
java.io.UncheckedIOException: java.nio.file.FileAlreadyExistsException:
  .../.gradle/caches/modules-2/files-2.1/io/netty/.../netty-...-sources.jar
    at ...resolver.remote.Downloader.performDownload(Downloader.java)
Caused by: java.nio.file.FileAlreadyExistsException
    at java.base/java.nio.file.Files.copy(Files.java)
```

## Root cause

For the Gradle resolver, `localRepository` is the **shared** Gradle module cache (`~/.gradle/caches/modules-2/files-2.1`), which the Gradle daemon and other RJE worker threads write to concurrently. `Downloader.performDownload` cached downloads with:

```java
Files.copy(pathInRepo, cachedResult, REPLACE_EXISTING);
```

`Files.copy(REPLACE_EXISTING)` is not atomic: `REPLACE_EXISTING` only deletes a target that exists *at check time*, so a target created by another writer in the TOCTOU window makes the underlying create fail with `FileAlreadyExistsException`. A plain in-place copy also lets a concurrent worker observe a **half-written** cache file via the `Files.exists(cachedResult)` check at the top of `performDownload` and then hash a truncated artifact.

## Fix

Cache to a temp file in the destination directory, then atomically move it into place (`ATOMIC_MOVE`, with a non-atomic fallback for filesystems that don't support it). Artifacts here are content-addressed, so a destination another writer already produced is *success* — `FileAlreadyExistsException` on the move is ignored, and readers never see a partial file.

## Test

Adds `DownloaderTest.cachingDownloadToleratesPreexistingDestination`, which pre-creates the cache destination (simulating a concurrent writer) with caching enabled and asserts the download succeeds instead of throwing.

### DIFF
--- a/private/tools/java/com/github/bazelbuild/rules_jvm_external/resolver/remote/Downloader.java
+++ b/private/tools/java/com/github/bazelbuild/rules_jvm_external/resolver/remote/Downloader.java
@@ -14,6 +14,7 @@
 
 package com.github.bazelbuild.rules_jvm_external.resolver.remote;
 
+import static java.nio.file.StandardCopyOption.ATOMIC_MOVE;
 import static java.nio.file.StandardCopyOption.REPLACE_EXISTING;
 
 import com.github.bazelbuild.rules_jvm_external.Coordinates;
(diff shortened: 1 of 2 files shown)


shs96c (approved):
LGTM. Thank you! Must have been a pain to track down

PR #1616, opened 2026-07-31
Return repo_metadata from repository rules

keith:
This is an explicit marker to opt out of the remote repo contents cache,
and satisfies https://github.com/bazelbuild/bazel/pull/30202 for finding
other cases.


shs96c (approved):
LGTM. Thank you for this!

PR #1618, opened 2026-08-11
Encapsulate bzlmod dependency layering

shs96c:
The `bzlmod` layering has grown organically and it isn't always clear what will happen in all circumstances. This PR resolves that issue by clearly documenting the expected behaviour and adding tests around that behaviour. 

To clarify, this is how we expect bazel module layering to work, comparing root and non-root modules:

For each Maven coordinate, more than one module may declare it, and any declaration may force its version. Only one declaration survives to the resolver. It is chosen by these rules, in order:

1. A forced version in the root module always wins.
2. Otherwise, a forced declaration beats any unforced one, whatever the versions.
3. Otherwise, the highest version wins, regardless of which module declared it.

Ties and conflicts:
- Two non-root modules that force different versions is an error and fails before resolution. The root can settle it by forcing the version itself.
- On a tie (equal versions, or the same forced version from more than one module) the first module's declaration is kept; the root counts as first.
- The surviving declaration is kept whole: its exclusions and flags come with it, and nothing is merged in from the declarations that lost.
- A non-root artifact marked `testonly` is dropped.

This means that the root module can always specify dependencies how it wants, but non-root modules can contribute.

BOM-only contributors now trigger the contribution warning, and `duplicate_version_warning` is applied at the layering boundary as `warn`, `error`, or `none`, remaining silent whenever the root version is selected.

### DIFF
--- a/docs/bzlmod.md
+++ b/docs/bzlmod.md
@@ -194,9 +194,10 @@ The current layering rules are:
 | 5 | One non-root module only | Its declaration is retained. A non-root artifact marked `testonly` is filtered out. |
 | 6 | Multiple non-root modules, with no force | The highest version is retained. Equal versions keep the first module's complete declaration, including exclusions and flags. |
 | 7 | Forced root and unforced non-root | The root declaration is retained, regardless of version. |
-| 8 | Multiple non-root modules force different versions | The declarations are silently collapsed to the highest version instead of failing. The winning declaration's flags are retained. A still-higher unforced declaration can therefore replace a lower forced declaration and remove the force. |
+| 8 | Multiple non-root modules force different versions, and the root does not force that coordinate | Layering fails before resolution. The error identifies both contributing modules and versions, and tells the user to declare the wanted version in the root module with `force_version = True`. |
 
-Rule 8 is a current limitation: conflicting non-root forces do not report an error.
+A single forced non-root declaration still participates in highest-version deduplication with other
+non-root declarations. A higher unforced declaration can replace it and remove the force.
 
 "Highest" uses the Maven `ComparableVersion` ordering implemented by
 `private/rules/maven_version.bzl`, not lexical string ordering. Non-root deduplication uses this
@@ -247,6 +248,7 @@ contributing module:
 | Condition | Behaviour |
 |---|---|
 | An unacknowledged non-root module contributes artifacts | Always prints the contribution warning. |
+| Non-root modules force different versions of a coordinate that the root does not force | Fails with the contributing module names and versions, and tells the user how to select a version in the root module. |
 | `known_contributing_modules` excludes an artifact or BOM contributor | Prints an `INFO` message when `RJE_VERBOSE` is set. |
 | A higher unforced non-root version survives alongside the root version, or a differently versioned forced non-root replaces it | Prints the root-version warning when `REPIN` or `RULES_JVM_EXTERNAL_REPIN` is set. |
 | A non-root-only coordinate is added | Prints an `INFO` message when a repin variable and `RJE_VERBOSE` are both set. |
--- a/docs/bzlmod.md
+++ b/docs/bzlmod.md
@@ -178,20 +178,24 @@ another namespace.
 
 The root module and its dependencies have different roles during layering. The root contributes
 the declarations that belong to the current Bazel project. Every other module is a non-root
-contributor. Coordinates are conceptually matched by `group:artifact:packaging:classifier`, meaning that a classified JAR layers independently of its unclassified JAR.
+contributor. Coordinates are conceptually matched by `group:artifact:packaging:classifier`,
+meaning that a classified JAR layers independently of its unclassified JAR.
 
-When performing [duplicate coordinate checks](#diagnostics), the keys declarations by `group:artifact:classifier`,
-but not packaging. Packaging-distinct declarations at different versions can therefore warn or fail
-even though the extension layers them independently. It also continues to check multiple root
-declarations. Ordinary cross-module conflicts with the same layering key no longer reach this check.
+When performing [duplicate coordinate checks](#diagnostics), the declarations are keyed by
+`group:artifact:classifier`, but not packaging. Packaging-distinct declarations at different
+versions can therefore warn or fail even though the extension layers them independently. It also
+continues to check multiple root declarations. Ordinary cross-module conflicts with the same
+layering key no longer reach this check.
 
 ### Version precedence
 
 For conflicts between modules, layering selects one complete declaration for each coordinate. The
 selected declaration supplies its exclusions, `neverlink`, `testonly`, `force_version`, packaging,
-classifier, and other fields. Fields from discarded declarations are not merged into it. Duplicate
-declarations made by the root module normally remain for the repository-level check. If any module
-sets `force_version` on the same coordinate at two different versions, layering will fail with an
+classifier, and other fields. Fields from discarded declarations are not merged into it. Layering
+does not deduplicate within the root module, so repeated root declarations for one coordinate reach
+the existing repository-level duplicate check, which warns or fails according to
+`duplicate_version_warning`. Forcing is the exception: if any module, the root included, sets
+`force_version` on the same coordinate at two different versions, layering will fail with an
 error message.
 
 The surviving declaration is chosen by these rules:
@@ -206,14 +210,15 @@ On ties and conflicts:
 - On a tie (equal versions, or the same forced version from more than one module) the first module's declaration is kept; the root counts as first.
 - A non-root artifact marked `testonly` is dropped.
 
-Be aware that non-default packaging and classifiers remain independent of each other and the plain versioned coordinate. This may lead to some surprises when resolution is complete.
+Be aware that non-default packaging and classifiers remain independent of each other and of the
+plain versioned coordinate. This may lead to some surprises when resolution is complete.
 
 "Highest" uses the Maven `ComparableVersion` ordering implemented by
 `private/rules/maven_version.bzl`, not lexical string ordering.
 
 `version_conflict_policy = "pinned"` changes this interaction. For the Gradle and Maven resolvers,
-root artifacts are marked as `force_version` before layering. The duplicate-force check applies to declared
-forces before this policy is applied. Maven then marks every versioned root declaration.
+root artifacts are marked as `force_version` before layering. The duplicate-force check applies to
+declared forces before this policy is applied. Maven then marks every versioned root declaration.
 Gradle first selects one version for each root `group:artifact`: an unclassified declaration takes
 precedence over classified declarations, and Maven `ComparableVersion` order selects among
 declarations with the same classification status. Every root declaration for that module at the
@@ -267,7 +272,8 @@ forced root, an unforced root with the highest version, an equal unforced tie, a
 forced non-root whose complete declaration survives. Coordinates that the root did not declare
 also have no version-selection diagnostic; the contribution warning covers their origin.
 
-As an example, the version-selection warning when layering selects a version different from the root version looks like:
+As an example, the version-selection warning when layering selects a version different from the
+root version looks like:
 
 ```
 WARNING: For dependency 'com.google.protobuf:protobuf-java' the root @maven repo wants version 3.25.5, but got 4.27.2 from the bazel_worker_java bazel dep. Please update the version in your MODULE.bazel or set `force_version = True`.
--- a/docs/bzlmod.md
+++ b/docs/bzlmod.md
@@ -172,43 +172,96 @@ for matching.
 
 ## Module dependency layering
 
-In order to allow modules to collaborate on required dependencies, the `bzlmod` extension will
-collect the artifacts from all tags with the same `name` attribute together before performing a
-dependency resolution. You'll know this is happening because a message will be printed to inform
-you which modules are contributing to which namespace:
+The extension collects declarations from all tags with the same `name` before resolving them. Each
+name is an independent Maven repository namespace. Declarations in one namespace never affect
+another namespace.
+
+The root module and its dependencies have different roles during layering. The root contributes
+the declarations that belong to the current Bazel project. Every other module is a non-root
+contributor. Coordinates are matched by `group:artifact`, with non-default packaging and classifiers
+included in the key. For example, a classified JAR layers independently of its unclassified JAR.
+
+### Version precedence
+
+The current layering rules are:
+
+| # | Declarations for one coordinate | Layering behaviour |
+|---|---|---|
+| 1 | Root only | The root declaration is retained. |
+| 2 | Root and non-root, with no `force_version` | A lower or equal non-root declaration is discarded. A higher non-root declaration is retained alongside the root declaration and the resolver chooses between them. Coursier and Gradle choose the highest version, while the Maven resolver chooses the root version because it uses the first direct declaration and root declarations are first. `duplicate_version_warning = "error"` fails before resolution when both versions survive. |
+| 3 | Unforced root and forced non-root | The non-root declaration is retained only when its version is strictly higher. A lower or equal declaration is silently discarded, including its force flag. This means an equal non-root declaration does not pin the coordinate against a transitive upgrade. |
+| 4 | Forced root and forced non-root | The root declaration is retained and the non-root declaration is discarded. |
+| 5 | One non-root module only | Its declaration is retained. A non-root artifact marked `testonly` is filtered out. |
+| 6 | Multiple non-root modules, with no force | The highest version is retained. Equal versions keep the first module's complete declaration, including exclusions and flags. |
+| 7 | Forced root and unforced non-root | The root declaration is retained, regardless of version. |
+| 8 | Multiple non-root modules force different versions | The declarations are silently collapsed to the highest version instead of failing. The winning declaration's flags are retained. A still-higher unforced declaration can therefore replace a lower forced declaration and remove the force. |
+
+Rules 3 and 8 are current limitations: a non-root force does not take precedence over an unforced
+root in every case, and conflicting non-root forces do not report an error.
+
+"Highest" uses the Maven `ComparableVersion` ordering implemented by
+`private/rules/maven_version.bzl`, not lexical string ordering. Non-root deduplication uses this
+ordering before declarations are compared with the root.
+
+Rule 2 is resolver-dependent because the extension passes both declarations to the resolver when
+the non-root version is higher. Coursier uses its default highest-version reconciliation. Gradle
+also selects the highest version. The Maven resolver normalises duplicate direct
+`group:artifact` declarations to the first version, which is the root version.
+
+`version_conflict_policy = "pinned"` changes this interaction. For the Gradle and Maven resolvers,
+root artifacts are force-marked before layering, so the root wins even against a higher forced
+non-root declaration. For Coursier, the layering step is unchanged and every surviving direct
+version is later passed as a `--force-version` argument.
+
+The `force_version` flag can be set by an `artifact` tag, an `amend_artifact` tag, or a regular
+artifact read by `from_toml`. Coordinates in `install.artifacts` cannot carry the flag. BOMs use the
+same extension-layer precedence rules as artifacts, but a BOM can only gain `force_version` through
+`amend_artifact`. A TOML BOM is returned before the TOML force field is copied. Resolver requests do
+not propagate a BOM force flag, so it affects extension-layer priority only.
+
+### Contributors and configuration
+
+When the root and other modules contribute artifacts to the same namespace, the extension prints a
+message such as:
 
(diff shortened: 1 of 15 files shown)


docs/bzlmod.md
```diff
@@ -172,44 +172,98 @@ for matching.
...
+included in the key. For example, a classified JAR layers independently of its unclassified JAR.
+
+### Version precedence
+
+The current layering rules are:
+
+| # | Declarations for one coordinate | Layering behaviour |
+|---|---|---|
+| 1 | Root only | The root declaration is retained. |
+| 2 | Root and non-root, with no `force_version` | A lower or equal non-root declaration is discarded. A higher non-root declaration is retained alongside the root declaration and the resolver chooses between them. Coursier and Gradle choose the highest version, while the Maven resolver chooses the root version because it uses the first direct declaration and root declarations are first. `duplicate_version_warning = "error"` fails before resolution when both versions survive. |
```
shs96c: This is wrong. The highest version, no matter where it's from should be chosen.

docs/bzlmod.md
```diff
@@ -172,44 +172,98 @@ for matching.
...
+| 3 | Unforced root and forced non-root | The non-root declaration is retained and the root declaration is discarded, regardless of version. An equal non-root declaration therefore retains its pin against a transitive upgrade. |
+| 4 | Forced root and forced non-root | The root declaration is retained and the non-root declaration is discarded. |
+| 5 | One non-root module only | Its declaration is retained. A non-root artifact marked `testonly` is filtered out. |
+| 6 | Multiple non-root modules, with no force | The highest version is retained. Equal versions keep the first module's complete declaration, including exclusions and flags. |
+| 7 | Forced root and unforced non-root | The root declaration is retained, regardless of version. |
+| 8 | Multiple non-root modules force different versions, and the root does not force that coordinate | Layering fails before resolution. The error identifies both contributing modules and versions, and tells the user to declare the wanted version in the root module with `force_version = True`. |
+
+A single forced non-root declaration still participates in highest-version deduplication with other
```
shs96c: This shouldn't happen. If we have a single forced non-root declaration, that should be the one that wins. We need to correct that in this PR and document in the table above (probably leaving the "multiple non-root modules" item as the last thing in the table)

docs/bzlmod.md
```diff
@@ -172,44 +172,98 @@ for matching.
...
+non-root declarations. A higher unforced declaration can replace it and remove the force.
+
+"Highest" uses the Maven `ComparableVersion` ordering implemented by
+`private/rules/maven_version.bzl`, not lexical string ordering. Non-root deduplication uses this
+ordering before declarations are compared with the root.
+
+Rule 2 is resolver-dependent because the extension passes both declarations to the resolver when
```
shs96c: Rule 2 isn't working the way it should

docs/bzlmod.md
```diff
@@ -172,44 +172,98 @@ for matching.
...
 
 `The maven repository 'multiple_lock_files' has contributions from multiple bzlmod modules, and will be resolved together: ["bzlmod_lock_files", "rules_jvm_external"]`
 
-In the root module, if this is expected and known, you can disable this warning by adding
-the list of modules to the `known_contributing_modules` attribute of the `install` tag. The entry
-to add will be printed for you as part of the warning. Once you set a value for `known_contributing_modules` then only those modules will be allowed to contribute dependencies.
+If those contributions are expected, set `known_contributing_modules` on the root `install` tag.
+The warning includes the value to add. Once this attribute is non-empty, only listed modules may
+contribute artifacts or BOMs to that namespace. A module that contributes only BOMs does not trigger
+the contribution warning.
```
shs96c: A module contributing only BOMs _should_ trigger the contribution warning. Please fix in this PR.

docs/bzlmod.md
```diff
@@ -172,44 +172,98 @@ for matching.
...
-    com.google.guava:guava (31.1-jre, 33.2.1-jre)
-```
+| Condition | Behaviour |
+|---|---|
+| An unacknowledged non-root module contributes artifacts | Always prints the contribution warning. |
+| Non-root modules force different versions of a coordinate that the root does not force | Fails with the contributing module names and versions, and tells the user how to select a version in the root module. |
+| `known_contributing_modules` excludes an artifact or BOM contributor | Prints an `INFO` message when `RJE_VERBOSE` is set. |
+| A higher unforced non-root version survives alongside the root version, or a differently versioned forced non-root replaces it | Prints the root-version warning when `REPIN` or `RULES_JVM_EXTERNAL_REPIN` is set. |
+| A non-root-only coordinate is added | Prints an `INFO` message when a repin variable and `RJE_VERBOSE` are both set. |
+| Multiple versions reach the repository rule | `duplicate_version_warning` controls whether to warn, fail, or do nothing. Its default is `warn`. |
```
shs96c, line 268: We should warn unless the version from the root module has been selected (either via a `force_version` or because it has the highest version)

docs/bzlmod.md
shs96c: 🤖 Updated: layering now retains one complete declaration per cross-module coordinate and selects the highest version regardless of whether it came from the root or a non-root module. Equal-version ties retain the root declaration.

docs/bzlmod.md
shs96c: 🤖 Fixed: rule 2 now replaces the root declaration when a higher non-root version wins instead of passing both declarations to the repository rule. Tests cover the selection and its relocated multiple-version diagnostic.

docs/bzlmod.md
shs96c: 🤖 Updated: a single forced non-root declaration now wins over every unforced declaration, including a higher root version. Conflicting forced non-root modules still fail unless the root forces the coordinate, and the multiple-non-root cases are last in the table.

docs/bzlmod.md
shs96c: 🤖 Fixed: contribution detection now uses the union of non-root artifact and BOM contributors, so a BOM-only module triggers the warning. known_contributing_modules acknowledges contributors from either map.

docs/bzlmod.md
shs96c: 🤖 Updated: duplicate_version_warning now applies at the layering boundary as warn, error, or none when the selected version differs from the root version. It remains silent when the root version is selected by force, by being highest, or on an equal-version tie.

private/lib/layering.bzl
```diff
@@ -0,0 +1,335 @@
...
+def contributing_modules_warning(repo_name, known_contributing_modules, non_root_bazel_dep_to_items):
+    """Returns the warning for contributions from modules not acknowledged by the root."""
+    contributing_module_names = non_root_bazel_dep_to_items.keys()
+    new_contributing_modules = sets.difference(sets.make(contributing_module_names), known_contributing_modules)
+    if sets.length(new_contributing_modules) > 0:
+        return (
+            "The maven repository '%s' has contributions from multiple bzlmod modules, and will be resolved together: %s." % (
+                repo_name,
+                sorted(contributing_module_names),
+            ) + "\nSee https://github.com/bazel-contrib/rules_jvm_external/blob/master/docs/bzlmod.md#module-dependency-layering" +
```
shs96c: Is this URL correct now?

docs/bzlmod.md
```diff
@@ -172,44 +172,113 @@ for matching.
...
-collect the artifacts from all tags with the same `name` attribute together before performing a
-dependency resolution. You'll know this is happening because a message will be printed to inform
-you which modules are contributing to which namespace:
+The extension collects declarations from all tags with the same `name` before resolving them. Each
+name is an independent Maven repository namespace. Declarations in one namespace never affect
+another namespace.
+
+The root module and its dependencies have different roles during layering. The root contributes
+the declarations that belong to the current Bazel project. Every other module is a non-root
+contributor. Coordinates are matched by `group:artifact`, with non-default packaging and classifiers
```
cheister: Would it be more clear to just say that coordinates are matched by `group:artifact:packaging:classifier` ?

docs/bzlmod.md
```diff
@@ -172,44 +172,113 @@ for matching.
...
+included in the key. For example, a classified JAR layers independently of its unclassified JAR.
+
+### Version precedence
+
+For conflicts between modules, layering selects one complete declaration for each coordinate. The
+selected declaration supplies its exclusions, `neverlink`, `testonly`, `force_version`, packaging,
+classifier, and other fields. Fields from discarded declarations are not merged into it. Duplicate
+declarations made by the root module normally remain for the repository-level check. If any module
+sets `force_version` on the same coordinate at two different versions, layering fails first.
```
cheister: Maybe reword to "Layering will fail" or "a <type of error> will be raised if force_version is set on the same coordinate in two layers

docs/bzlmod.md
```diff
@@ -172,44 +172,113 @@ for matching.
...
+
+The surviving declaration is chosen by these rules, in order:
```
cheister: Probably don't need ", in order" at the end of this sentence

docs/bzlmod.md
```diff
@@ -172,44 +172,113 @@ for matching.
...
+2. Otherwise, a forced declaration beats any unforced one, whatever the versions.
+3. Otherwise, the highest version wins, regardless of which module declared it.
+
+On ties and conflicts:
+
+- Two non-root modules that force different versions is an error and fails before resolution. The root can settle it by forcing the version itself.
+- On a tie (equal versions, or the same forced version from more than one module) the first module's declaration is kept; the root counts as first.
+- A non-root artifact marked `testonly` is dropped.
+
+Within one non-root module, a forced declaration beats its unforced duplicates. Forced declarations
```
cheister: Feels like we don't need this paragraph?

docs/bzlmod.md
```diff
@@ -172,44 +172,113 @@ for matching.
...
+for the same coordinate at different versions fail before selection. Matching forced versions retain
+the first complete declaration. Among unforced declarations, the highest version wins and an
+equal-version tie retains the first complete declaration. This check uses the same coordinate key as
+layering, so non-default packaging and classifiers remain independent.
+
+"Highest" uses the Maven `ComparableVersion` ordering implemented by
+`private/rules/maven_version.bzl`, not lexical string ordering.
+
+`version_conflict_policy = "pinned"` changes this interaction. For the Gradle and Maven resolvers,
+root artifacts are force-marked before layering. The duplicate-force check applies to declared
```
cheister: What does "force-marked" mean?

docs/bzlmod.md
```diff
@@ -172,44 +172,113 @@ for matching.
...
+forces before this policy is applied. Maven then marks every versioned root declaration.
+Gradle first selects one version for each root `group:artifact`: an unclassified declaration takes
+precedence over classified declarations, and Maven `ComparableVersion` order selects among
+declarations with the same classification status. Every root declaration for that module at the
+selected version is then marked forced, including classified declarations. The root consequently
+wins because it now forces the coordinate. For Coursier, layering is unchanged and the one
+surviving direct version is later passed as a `--force-version` argument. A higher non-root
+version can therefore displace the root under Coursier and then be pinned.
+
+The `force_version` flag can be set by an `artifact` tag, an `amend_artifact` tag, or a regular
```
cheister, line 230: Can you re-phrase this part?  Sounds like you can only use force_version on artifacts and force_version from BOMs are ignored?

docs/bzlmod.md
```diff
@@ -172,44 +172,113 @@ for matching.
...
-transitive dependency is at an unexpected version you'll have the information you need to diagnose
-the problem.
+### Diagnostics
 
-When dependencies are layered in this way, you may see a warning similar to:
+Layering keeps the following diagnostics so that unexpected versions can be traced to their
+contributing module:
 
-```
-"WARNING: The following maven modules appear in multiple sub-modules with potentially different versions. Consider adding one of these to your root module to ensure consistent versions:
```
cheister, line 201: Is this message no longer output?

docs/bzlmod.md
```diff
@@ -172,44 +172,113 @@ for matching.
...
+even though the extension layers them independently. It also continues to check multiple root
+declarations. Ordinary cross-module conflicts with the same layering key no longer reach this check.
+
+For example, the version-selection warning is:
 
 ```
 WARNING: For dependency 'com.google.protobuf:protobuf-java' the root @maven repo wants version 3.25.5, but got 4.27.2 from the bazel_worker_java bazel dep. Please update the version in your MODULE.bazel or set `force_version = True`.
 ```
 
-You can either update the version in the root module to the highest version or set `force_version = True` in the root module to ensure that version will be the one used in the dependency resolution.
```
cheister, line 211: I think this part is still useful to keep

docs/bzlmod.md
```diff
@@ -172,44 +172,113 @@ for matching.
...
+| A non-root-only coordinate is added | Prints an `INFO` message when a repin variable and `RJE_VERBOSE` are both set. |
+| Multiple versions reach the repository rule | `duplicate_version_warning` controls whether to warn, fail, or do nothing. Its default is `warn`. |
+
+No version-selection diagnostic is emitted when the root version is selected. This includes a
+forced root, an unforced root with the highest version, an equal unforced tie, and an equal-version
+forced non-root whose complete declaration survives. Coordinates that the root did not declare
+also have no version-selection diagnostic; the contribution warning covers their origin.
 
-The resolver will use the highest version artifact from the root and sub-modules. If the root version is not the highest you will see a warning during repinning similar to:
+The repository-level duplicate check keys declarations by `group:artifact` and optional classifier,
```
cheister: Should this information be moved a above where the coordinate comparison is explained?

private/extensions/maven.bzl
```diff
@@ -77,7 +81,7 @@ install = tag_class(
         "additional_netrc_lines": attr.string_list(doc = "Additional lines prepended to the netrc file used by `http_file` (with `maven_install_json` only).", default = []),
         "use_credentials_from_home_netrc_file": attr.bool(doc = "Whether to pass machine login credentials from the ~/.netrc file to coursier.", default = False),
         "duplicate_version_warning": attr.string(
-            doc = """What to do if there are duplicate artifacts
+            doc = """What to do if layering selects a non-root version instead of the root version, or if duplicate artifacts reach the repository rule
```
cheister: I'm not sure I understand what "duplicate artifacts reach the repository rule" means.  Can we say, the same artifact is duplicated across contributing modules"?

private/extensions/maven.bzl
```diff
@@ -701,62 +573,21 @@ def maven_impl(mctx):
         merged_repo.update(non_root_repo)
         merged_repo.update(root_repo)
 
-        # Special handling for artifacts and boms - deduplicate with root priority
-        root_artifacts = apply_root_version_conflict_policy(
-            root_repo.get("artifacts", []),
-            root_repo.get("resolver", _DEFAULT_RESOLVER),
-            root_repo.get("version_conflict_policy", "default"),
+        layered = layer_maven_namespace(
```
cheister: nit: rename layered to layered_artifacts_and_boms

private/lib/layering.bzl
```diff
@@ -0,0 +1,335 @@
...
+    )
+    diagnostics.extend(boms.diagnostics)
+
+    return struct(
+        artifacts = artifacts.artifacts,
+        boms = boms.artifacts,
+        diagnostics = diagnostics,
+    )
+
+def remove_fields(s):
```
cheister: Should this be named remove_empty_fields instead of remove_fields ?

docs/bzlmod.md
shs96c: Done

docs/bzlmod.md
shs96c: Done

docs/bzlmod.md
shs96c: Amended. The warning at the end is valuable.

docs/bzlmod.md
shs96c: Marked with `force_version`. Clarified.

docs/bzlmod.md
shs96c: Removed the last half of the paragraph.

docs/bzlmod.md
shs96c: This message isn't output in `master`. The current message, in `master` and with this PR, looks like:

```
WARNING: For dependency 'com.fasterxml.jackson.core:jackson-databind' the root @maven repo wants version 2.14.3, but got 2.17.2 from the sub2 bazel dep. Please update the version in your MODULE.bazel or set `force_version = True`.
```

That's covered in the table below as "Layering selects a version different from the root version"

docs/bzlmod.md
shs96c: I put this here because it's useful for trying to figure out why a warning about versions has been printed. Lifted up to earlier, and adding a link to the diagnostics section.

private/extensions/maven.bzl
shs96c: If the root module declares the same coordinates more than once, then we pass all of them to the resolver to resolve. This is the warning that the repo rule will print. Reworded.

private/extensions/maven.bzl
shs96c: Done

private/lib/layering.bzl
shs96c: That's probably a better name. Renamed.

docs/bzlmod.md
```diff
@@ -172,37 +172,108 @@ for matching.
...
+as a Maven dependency, or when the project is only used as the root module.
+
+### <a id="diagnostics"></a>Diagnostics
+
+Layering keeps the following diagnostics so that unexpected versions can be traced to their
+contributing module:
+
+| Condition | Behaviour |
+|---|---|
+| An unacknowledged non-root module contributes artifacts or BOMs | Always prints the contribution warning. |
```
cheister: We should add the actual warning messages printed in this table so we don't have to map "the contribution warning" back to what the user actually sees.

Then add something that would be possible fixes for the message

docs/bzlmod.md
shs96c: Amended.

private/lib/layering.bzl
```diff
@@ -0,0 +1,335 @@
...
+                    )
+                    if duplicate_version_warning == "error":
+                        fail(message)
+                    elif duplicate_version_warning == "warn":
+                        duplicate_artifact_warning = duplicate_artifact_warning + "\nWARNING: " + message
+        if keep_root_artifact:
+            filtered_root_artifacts.append(root_artifact)
+
+    # Add any remaining non root artifacts that weren't found in the root artifact list
+    addtional_artifact_message = ""
```
vinnybod, line 168: NIT: `addtional_artifact_message` -> `additional_artifact_message`

private/lib/layering.bzl
```diff
@@ -0,0 +1,335 @@
...
+    coordinate_to_forced_artifact = {}
+    for bazel_dep_name in bazel_dep_to_non_root_artifacts:
+        module_coordinate_to_artifact = {}
+        module_artifacts = bazel_dep_to_non_root_artifacts.get(bazel_dep_name, [])
+        _fail_if_conflicting_forces(bazel_dep_name, module_artifacts)
+        for artifact in module_artifacts:
+            if not getattr(artifact, "testonly", False):
+                artifact_key = to_key(artifact)
+                if _candidate_takes_precedence(module_coordinate_to_artifact.get(artifact_key), artifact):
+                    module_coordinate_to_artifact[artifact_key] = artifact
```
vinnybod, line 87: Here we are failing if there are conflicting `force=True`, but we are filtering the `testonly` artifacts right after. It seems to make more sense to do the filtering first, then check for conflicts since those `testonly` artifacts are being removed anyway.

```suggestion
        module_artifacts = [
            artifact
            for artifact in bazel_dep_to_non_root_artifacts.get(bazel_dep_name, [])
            if not getattr(artifact, "testonly", False)
        ]
        _fail_if_conflicting_forces(bazel_dep_name, module_artifacts)
        for artifact in module_artifacts:
            artifact_key = to_key(artifact)
            if _candidate_takes_precedence(module_coordinate_to_artifact.get(artifact_key), artifact):
                module_coordinate_to_artifact[artifact_key] = artifact
```

private/lib/layering.bzl
```diff
@@ -0,0 +1,335 @@
...
+                                artifact_key,
+                                previous_bazel_dep_name,
+                                previous_artifact.version,
+                                bazel_dep_name,
+                                artifact.version,
+                            ) +
+                            "Add an `artifact` tag to the root module at the version you want and set `force_version = True`.",
+                        )
+                else:
+                    coordinate_to_forced_artifact[artifact_key] = (bazel_dep_name, artifact)
```
vinnybod, line 106: Correct me if I'm wrong, but we should be outputting diagnostics for these `force_version = True` instances?

cheister (approved):

PR #1619, opened 2026-08-11
Restructure rules_jvm_external documentation

shs96c:
Restructures and rewrites the user documentation around one canonical Bzlmod quickstart and topic pages, with an expanded README landing page and modernised examples for the Bazel Central Registry. It also makes the generated API references reproducible and complete, and adds migration and operational guidance. The docs now document Gradle's dependency string notation for coordinates, no longer claim that the Coursier resolver cannot resolve BOMs (stale since 3289580), no longer call the Gradle resolver experimental, and cover `bazel run @maven//:direct_deps`.

`kt_jvm_export` documentation is generated directly from `//:kt_defs.bzl`, and the `docs/kt_defs_doc.bzl` mirror is removed. This folds in the module-dependency-layering rewrite from #1618 and tracks that PR's clarified rules, so the layering page describes the behaviour #1618 introduces rather than the current release; because #1618 also rewrites `docs/bzlmod.md`, whichever lands second must rebase.


docs/declaring-artifacts.md
```diff
@@ -0,0 +1,242 @@
...
+
+The `artifacts` attribute accepts coordinate strings only. An inline artifact
+map with an `exclusions` field fails under Bzlmod with an error such as:
+
+```text
+error converting value for attribute artifacts: expected value of type 'string' ... got None (NoneType)
+```
+
+Use a separate `maven.artifact` tag instead. This pattern was required by
+[grpc-java#10576](https://github.com/grpc/grpc-java/issues/10576). Remove the
```
shs96c: We don't need this link to grpc-java

docs/kt_defs_doc.bzl
```diff
@@ -0,0 +1,84 @@
+# This documentation-only mirror exists because rules_kotlin 2.2.2 does not
```
shs96c: Is this really not exposed by anything? `bazel query --noincompatible_disallow_empty_glob 'kind(bzl_library, @rules_kotlin//...)' ` suggests that there are some bzl_library rules that might be of use.

docs/lock-files.md
```diff
@@ -0,0 +1,158 @@
...
+```shell
+REPIN=1 bazel run @foo//:pin
+REPIN=1 bazel run @bar//:pin
+```
+
+Within one generated repository name, all contributing tags must agree on a
+single lock file. See [Module dependency layering](module-dependency-layering.md)
+for contributions from non-root Bazel modules.
+
+## Find outdated dependencies
```
shs96c, line 141: It's also possible to list all first-order dependencies (with their versions) using `bazel run @maven//:direct_deps` This will include the version numbers for all artifacts. This is useful for a number of reasons, one being when a coordinate has the version supplied from a BOM.

docs/mirrors-offline.md
```diff
@@ -0,0 +1,111 @@
...
+reuse the same pinned artifact. `--nofetch` makes an uncached repository fetch
+fail instead of accessing the network.
+
+Fetching `@maven//...` warms the Maven artifacts only. A disconnected build must
+also prepare Bazel modules, toolchains, and other external repositories. Follow
+Bazel's
+[offline build guidance](https://bazel.build/external/faq#how-do-i-prepare-and-run-an-offline-build)
+for the complete workspace.
+
+## Rewrite artifact URLs to a mirror
```
shs96c: This is the common case, and should be put first

docs/module-dependency-layering.md
```diff
@@ -0,0 +1,93 @@
...
+included in the key. For example, a classified JAR layers independently of its unclassified JAR.
+
+## Version precedence
+
+The current layering rules are:
+
+| # | Declarations for one coordinate | Layering behaviour |
+|---|---|---|
+| 1 | Root only | The root declaration is retained. |
+| 2 | Root and non-root, with no `force_version` | A lower or equal non-root declaration is discarded. A higher non-root declaration is retained alongside the root declaration and the resolver chooses between them. Coursier and Gradle choose the highest version, while the Maven resolver chooses the root version because it uses the first direct declaration and root declarations are first. `duplicate_version_warning = "error"` fails before resolution when both versions survive. |
```
shs96c: This feels weird. It should really be the same behaviour as rule 6: only one version should be presented to the resolver. We should update #1618 to fix that

docs/module-dependency-layering.md
```diff
@@ -0,0 +1,93 @@
...
+| 3 | Unforced root and forced non-root | The non-root declaration is retained and the root declaration is discarded, regardless of version. An equal non-root declaration therefore retains its pin against a transitive upgrade. |
+| 4 | Forced root and forced non-root | The root declaration is retained and the non-root declaration is discarded. |
+| 5 | One non-root module only | Its declaration is retained. A non-root artifact marked `testonly` is filtered out. |
+| 6 | Multiple non-root modules, with no force | The highest version is retained. Equal versions keep the first module's complete declaration, including exclusions and flags. |
+| 7 | Forced root and unforced non-root | The root declaration is retained, regardless of version. |
+| 8 | Multiple non-root modules force different versions, and the root does not force that coordinate | Layering fails before resolution. The error identifies both contributing modules and versions, and tells the user to declare the wanted version in the root module with `force_version = True`. |
+
+A single forced non-root declaration still participates in highest-version deduplication with other
```
shs96c: A single forced non-root declaration should take precedence over all other non-root declarations. We need to fix that.

docs/module-dependency-layering.md
```diff
@@ -0,0 +1,93 @@
...
+[`private/rules/maven_version.bzl`](../private/rules/maven_version.bzl), not lexical string ordering. Non-root deduplication uses this
+ordering before declarations are compared with the root.
+
+Rule 2 is resolver-dependent because the extension passes both declarations to the resolver when
+the non-root version is higher. Coursier uses its default highest-version reconciliation. Gradle
+also selects the highest version. The Maven resolver normalises duplicate direct
+`group:artifact` declarations to the first version, which is the root version.
+
+`version_conflict_policy = "pinned"` changes this interaction. For the Gradle and Maven resolvers,
+root artifacts are force-marked before layering, so the root wins even against a higher forced
```
shs96c: s/force-marked/have the `force_version` attribute set/

docs/module-dependency-layering.md
```diff
@@ -0,0 +1,93 @@
...
+## Contributors and configuration
+
+When the root and other modules contribute artifacts to the same namespace, the extension prints a
+message such as:
+
+`The maven repository 'multiple_lock_files' has contributions from multiple bzlmod modules, and will be resolved together: ["bzlmod_lock_files", "rules_jvm_external"]`
+
+If those contributions are expected, set `known_contributing_modules` on the root `install` tag.
+The warning includes the value to add. Once this attribute is non-empty, only listed modules may
+contribute artifacts or BOMs to that namespace. A module that contributes only BOMs does not trigger
```
shs96c: This last sentence is a bug, and we should fix that in a follow up PR.

docs/resolution.md
```diff
@@ -0,0 +1,242 @@
...
+| `RJE_ASSUME_PRESENT` | Skip remote existence checks and assume dependencies are present. |
+| `RJE_UNSAFE_CACHE` | Maven uses the shared `$HOME/.m2/repository` by default. Set to `0` or `false` for an isolated cache. |
+
+The shared local Maven repository can be used during resolution, but local paths
+are not written to the lock file unless `m2local` appears in `repositories`.
+The resolver also reads credentials from `$HOME/.netrc`.
+
+## Gradle
+
+The Gradle resolver is experimental. It requires a lock file, which may
```
shs96c: The gradle resolver is no longer experimental. We should make sure that's no longer said in the docs.

docs/declaring-artifacts.md
shs96c: 🤖 Removed in 1c9abae.

docs/lock-files.md
shs96c: 🤖 Documented in 1c9abae: docs/lock-files.md now has a "List direct dependencies" section covering `bazel run @maven//:direct_deps`, including the BOM-supplied-version case.

docs/mirrors-offline.md
shs96c: 🤖 Moved first in 1c9abae.

docs/module-dependency-layering.md
shs96c: 🤖 Reworded in 1c9abae.

docs/resolution.md
shs96c: 🤖 Fixed in 1c9abae. docs/resolution.md and README.md were the only two places saying this; a sweep found no others.

docs/kt_defs_doc.bzl
shs96c: 🤖 Tried it. At our pinned rules_kotlin 2.2.2 the direct target fails analysis: kt_defs.bzl transitively loads src/main/starlark/core/plugin/{common,providers}.bzl, which are only covered by a bzl_library that is private to rules_kotlin (and the source files are not exported, so a local shim bzl_library fails visibility too). rules_kotlin 2.4.0 makes that plugin bzl_library public, and under a temporary 2.4.0 override the direct Stardoc target builds with explicit deps on //:implementation, @rules_java//java/private:proto_support, @rules_kotlin//kotlin:kotlin, @rules_kotlin//kotlin/internal:internal, and @rules_kotlin//src/main/starlark/core/plugin:plugin; the output is identical to the mirror's apart from a corrected load path. Since rules_kotlin is a regular bazel_dep, the 2.2.2 -> 2.4.0 bump propagates to consumers, so it should not ride this docs PR. 1c9abae records the verified recipe in the mirror's TODO; happy to do the bump plus mirror removal as a small follow-up PR.

docs/module-dependency-layering.md
shs96c: 🤖 Agreed this is a behaviour change for #1618 rather than this docs PR. Tracked; this page documents the current behaviour and will be updated when the #1618 change lands.

docs/module-dependency-layering.md
shs96c: 🤖 Agreed, same as above: tracked as a #1618 behaviour change, and this page will be updated to match.

docs/module-dependency-layering.md
shs96c: 🤖 Agreed, tracked for a follow-up PR. The sentence stays for now since it describes current behaviour.

PR #1620, opened 2026-08-14
Remove -noverify

timothyklim:
-noverify was removed in JDK 27 (deprecated since JDK 13) and now causes the launcher to fail with an unrecognized option error. It was only needed for proguarded coursier JARs from very old releases: https://github.com/coursier/coursier/pull/1245


shs96c (approved):
LGTM. Thank you!

PR #1625, opened 2026-08-20
Handle overrides for fileless artifacts

aNOOBisTheGod:
`maven.override` crashes while generating a pinned repository when the overridden artifact has no file, for example OkHttp's POM-only `com.squareup.okhttp3:okhttp-bom:5.1.0` ([source](https://github.com/square/okhttp), [POM](https://repo1.maven.org/maven2/com/squareup/okhttp3/okhttp-bom/5.1.0/okhttp-bom-5.1.0.pom), [no JAR](https://repo1.maven.org/maven2/com/squareup/okhttp3/okhttp-bom/5.1.0/okhttp-bom-5.1.0.jar)).

For fileless artifacts, only generate the override alias.


shs96c (approved):
LGTM. Thank you!

PR #1626, opened 2026-08-21
feat: expose kt_defs.bzl via public bzl_library for stardoc

mikhail-0330:
Currently, `kt_defs.bzl` is not exported and is not included in the public `bzl_library`, so stardoc throws an error “missing bzl_library”.

The changes for Kotlin are similar to the existing public target `implementation`, the comment for which already states that it is public precisely so that users can document their code (java_export) using stardoc.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -7,6 +7,7 @@ exports_files([
     "extensions.bzl",
     "specs.bzl",
     "maven_install.json",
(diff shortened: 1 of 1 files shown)


shs96c (approved):
Thank you! LGTM

PR #1627, opened 2026-09-02
feat: allow resolving coordinates as Android consumer using the `gradle` resolver

shs96c:
Kotlin Multiplatform modules need Android consumer selection, while JVM-only dependencies must stay usable in the same graph. This allows `rules_jvm_external` to be used for pure Android projects.

- Add Gradle-only `resolve_for = "android"` plumbing and lockfile invalidation.
- Configure Android variant selection and JVM fallback through the resolver plugin attribute schema.
- Cover Android AAR selection and JVM fallback.

Generated with Codex


cheister:
Does this imply that the coursier resolver does not work for android?

private/tools/java/com/github/bazelbuild/rules_jvm_external/resolver/gradle/plugin/GradleDependencyModelBuilder.java
```diff
@@ -73,6 +74,11 @@
  */
 public class GradleDependencyModelBuilder implements ToolingModelBuilder {
 
+  private static final Attribute<TargetJvmEnvironment> JVM_ENVIRONMENT_ATTRIBUTE =
+      TargetJvmEnvironment.TARGET_JVM_ENVIRONMENT_ATTRIBUTE;
+  private static final Attribute<String> KOTLIN_PLATFORM_ATTRIBUTE =
+      Attribute.of("org.jetbrains.kotlin.platform.type", String.class);
```
cheister, line 80: Is this the only variant we need to resolve for android?  Or should we be setting more?  e.g. 
```
org.gradle.category: library
  org.gradle.jvm.environment: standard-jvm
  org.gradle.libraryelements: jar
  org.gradle.usage: java-runtime
```

PR #1628, opened 2026-09-15
Use rules_pkg in the java-export example

fmeum:
The example loaded `pkg_tar` from `@bazel_tools//tools/build_defs/pkg:pkg.bzl`, which has been deprecated in favor of rules_pkg for years and is slated for removal from `@bazel_tools`.

This switches the example to `@rules_pkg//pkg:tar.bzl`.


shs96c (approved):
LGTM

